PR #25, opened 2015-08-07
Refactor build process

robes:
This pull request merges in changes that will add a build script that will:
- get development dependencies from `npm`
- get runtime dependencies from `bower`
- static code analysis (lint-like analysis by `jshint`)
- cats all files together into `chaise.js`
- compiles and minifies all source files into `chaise.min.js`
- able to now write tests in `mocha+chai` specs
- generates api documentation in markdown


mikedarcy:
Minified js code is harder to debug and troubleshoot because of all of the symbol replacement.  While this code is being actively developed, I would recommend the default build target NOT minify (or concatenate) by default, and there be a explicit "release" target (or somesuch) that minifies for public or release deployments.

robes:
@mikedarcy I agree we should tread lightly on using the minified and maybe even the concatenated code in most deployments for the time being until things are more stabilized. I do think its important to start including it in the build process and testing against them. I like the suggestion of having a *release* target specifically for producing those artifacts.

PR #26, opened 2015-08-16
Catalog Selector Window

codybraun:
Hey guys- 

I've added a catalog selecting window with a preview of the top few tables in each schema to Chaise. It could use some polishing, and I'm just hardcoding a list of catalog ids for now (I'll have to look more at ermrest to see if there's an easy way to get that list of all catalog ids from the registry. That's something else I'm happy to work on). I'm still learning my way around Angular, so let me know if I've screwed anything up horribly. 

If you have no catalog id integers hard coded in here or if you pass a catalog id in the URL, you'll just never see the selector and it might actually be really ugly if you're using catalogs with more than a couple schemas. Anyway, this is maybe not ready to be pulled into the repo but just wanted to check in about it. Let me know what improvements I could make to this or any further functionality that would be useful

catalog_selector


robes:
Thanks for the pull request!

This could potentially be of general use, if you are willing to make some revisions based on feedback. I would like to have another person on our team look at it and make some suggestions. It might help to have a live version she can look at. (You can communicate that privately if you prefer.)

One thing that I imagine we would want to do differently would be to avoid the modal dialog. So far we've done things with slide outs rather than dialogs and we want to keep the design elements consistent. Secondly, we'd want to make sure its mobile friendly in a reactive sense.

We also generally expect our users to be relatively non-technical. So I'm not sure that previewing tables would be informative to them. More likely they'd probably want a list of (catalog id, catalog description) for each catalog, where catalog description is a freeform text description.

One minor comment, you have a few temporary files that made it into your commit. E.g., filename.ext~ or #filename.ext#. (Seeing that, I just added a couple lines to our .gitignore)

Again thanks for the pull request and look forward to more discussion about it.

codybraun:
Hey guys-

I've CCed Kyle Chard, as he had some suggestions for the catalog selector.

I'm totally happy to make whatever revisions you suggest- I've just
implemented the catalog selector as a slide out bar, and I do think it
looks much cleaner and more in line with the rest of the design. I also
removed the table previews and included a short text descriptor (which is
just pulled from the meta data in the registry for each catalog).

I included the catalog selectors in a "settings" bar, which could
potentially contain other view preferences that a user could change, but
I'm not married to that design and it's pretty easy to move that button
somewhere else.

I do actually have a live version up, but I need to make sure I don't have
any sensitive databases sitting on it before I make it public, so here are
some screenshots for now and I'll pass along a link soon. Thanks for the
help and let me know if ya'll have any suggestions for this




​

hongsudt:
@robes We have this pull request waiting. Is this still relevant?

robes:
No

On 6/17/16, 5:26 PM, "Hongsuda" <notifications@github.com<mailto:notifications@github.com>> wrote:


@robes<https://github.com/robes> We have this pull request waiting. Is this still relevant?

hongsudt:
I am closing this issue since it is no longer relevant.

Issue #28, opened 2015-08-17
Cached application code problem


robes:
Whenever we redeploy Chaise and want to use it right away we have to use the browsers 'clear cache' feature otherwise we will just be using the browser's currently cached version of the .js files it loaded previously.

First, we need to identify the best strategy to avoid this problem. For instance, once basic approach out there is to always deploy your new code to a new path, e.g., path/to/v#/myscript.js where v# is some version number. Then we need to fix the deployment procedure for Chaise. This should be scripted so that with a single command we can (re)deploy Chaise. No manual steps should be involved.

karlcz:
Yes, and having per-file versioning can be more friendly to caches by
avoiding the version number change for content that doesn't change
between releases.  There are probably a lot of 3rd-party files
including icons and other bulky assets we change infrequently compared
to others...

howdyjessie:
After Googling around, per-file versioning seems to be the standard way of solving this issue, in which we append a hash to the names of files that have changed since last release to force the browser to redownload. Since we're already using Gulp in the ermrest.js repo, can we use it here as well (and with #31 )?

robes:
we aren’t using gulp. does that have a feature for this?

PR #37, opened 2015-08-20
Per-file versioning rules in Makefile for JS source

howdyjessie:
Appends a hash to the names of files that have changed in app.html (see #28)


robes:
This is looking really cool, great progress and all!

But a few things that need changing. For one, it's broken right now,

```sh
$ make app.html
make: *** No rule to make target `app.html.in', needed by `app.html'.  Stop.
```

I could tell by eyeballing it that the `app.html.in` was missing. I'll see if I can make some other comments in the commits next to the code.

.gitignore
```diff
@@ -13,3 +13,5 @@ jsdoc
 deployment-config.json
 *~
 #*
+*.in
```
robes: that seems wrong. we do want the *.in files. We should only ignore system generate files, temp files, build products, etc. I guess what I'm saying is ignore anything that is _not_ a source file or related.

Makefile
```diff
@@ -17,19 +17,20 @@ BOWER=bower_components
 
 # JavaScript source and test specs
 JS=scripts
-SOURCE=$(JS)/app.js \
-	   $(JS)/controller/ermrestDetailController.js \
+SOURCE=$(JS)/variables.js \
```
robes: could you fix the alignment or the source files. tab to the same spot.

Makefile
```diff
@@ -153,3 +154,14 @@ usage:
 	@echo "    clean     - cleans the dist dir"
 	@echo "    distclean - cleans the dist dir and the dependencies"
 
+
+# Rules to attach checksums to JavaScript source in the header
```
robes: great job figuring out these rules and the sed pattern!

can you move this block above the help/usage section. this is just a style preference, that I like to have the usage at the bottom or the top not somewhere in-between. thanks

Makefile
```diff
@@ -153,3 +154,14 @@ usage:
...
+app.html: app.html.in .make-script-block
+	sed -e '/%SCRIPTS%/ {' -e 'r .make-script-block' -e 'd' -e '}' -i app.html
```
robes: I don't agree with doing in-place edit (i.e., `-i` option) for this use.

We want a clear distinction between the _source file_ (`app.html.in`) and the _output file_ (`app.html`), and the source file should not be modified by the make procedure. Also the output file should be in the `dist` subdir.

robes:
Your recently modified __`app.html`__ should actually be renamed __`app.html.in`__ in the repo. You should:
```sh
$ git mv app.html app.html.in
```

howdyjessie:
Should .make-script-block be added to the hidden file targets section (like LINT and TEST)?

robes:
no, it probably belongs as a dependency of ‘build’.

.gitignore
```diff
@@ -13,3 +13,4 @@ jsdoc
 deployment-config.json
 *~
 #*
+.make-script-block
```
robes: good idea. let's make this
```
.make-*
```

Makefile
```diff
@@ -136,6 +137,16 @@ $(TEST): $(SOURCE) $(SPECS) $(BIN)
 testem:
 	$(BIN)/testem
 
+# Rules to attach checksums to JavaScript source in the header
+app.html: app.html.in .make-script-block
+	sed -e '/%SCRIPTS%/ {' -e 'r .make-script-block' -e 'd' -e '}' app.html.in > $(DIST)/app.html
+.make-script-block: $(SOURCE)
```
robes: i think its a good practice to have a blank line between rules in the makefile

Makefile
```diff
@@ -136,6 +137,16 @@ $(TEST): $(SOURCE) $(SPECS) $(BIN)
 testem:
 	$(BIN)/testem
 
+# Rules to attach checksums to JavaScript source in the header
+app.html: app.html.in .make-script-block
+	sed -e '/%SCRIPTS%/ {' -e 'r .make-script-block' -e 'd' -e '}' app.html.in > $(DIST)/app.html
```
robes: sorry. i know i asked you to put the app.html in DIST but for this PR can you ignore what i said and just make it in the base directory.

we'll need to make more changes in order to use the dist dir, so lets wait and define that as a separate task.

robes:
aside from my inline comments, one more thing. have you tested this to see that things (a) still work, and (b) that is actually fixes the versioning/caching problem?

howdyjessie:
I've tested to see that it does still work but haven't looked at the caching problem yet.

Makefile
```diff
@@ -136,6 +137,17 @@ $(TEST): $(SOURCE) $(SPECS) $(BIN)
...
 	$(BIN)/testem
 
+# Rules to attach checksums to JavaScript source in the header
+app.html: app.html.in .make-script-block
+	sed -e '/%SCRIPTS%/ {' -e 'r .make-script-block' -e 'd' -e '}' app.html.in > app.html
+
+.make-script-block: $(SOURCE)
+	> .make-script-block
+	for file in $(SOURCE); do \
+		checksum=$$(md5sum $$file | cut -c -32) ; \
```
robes: may i suggest changing this to

```sh
checksum=$$(md5sum $$file | awk '{ print $$1 }'); \
```

and bonus points if you can make this work on a Mac with `md5 -q` instead of `md5sum`

howdyjessie:
I just checked to see if it fixes the caching problem and it appears to work. I modified one of the JS files and Chrome returned a 200 response for the modified while returning a 304 Not Modified for the unchanged files.

Makefile
howdyjessie: D'oh! I'd tried to use awk earlier and thought it didn't work because I forgot to add the extra $ in $$1

howdyjessie: I throw down my gauntlet at the bonus points!

howdyjessie: Do we want md5 over md5sum because md5 comes with Mac OS X?

Makefile
```diff
@@ -136,6 +137,17 @@ $(TEST): $(SOURCE) $(SPECS) $(BIN)
...
 	$(BIN)/testem
 
+# Rules to attach checksums to JavaScript source in the header
+app.html: app.html.in .make-script-block
+	sed -e '/%SCRIPTS%/ {' -e 'r .make-script-block' -e 'd' -e '}' app.html.in > app.html
+
+.make-script-block: $(SOURCE)
+	> .make-script-block
+	for file in $(SOURCE); do \
+		checksum=$$(md5 -q $$file) ; \
```
robes: But that would only work on OS X (and other Unix variants) but not Linux (which doesn't have md5sum). You'll want to conditionally determine what is available and then use that conditional. This test seems to work:

```Makefile
# Detemine MD5 utility
ifeq ($(shell which md5),)
    MD5 = md5sum
else 
    MD5 = md5 -q
endif

md5:
    echo $(MD5)
```

That `md5` target is just there for demonstration, but not to be included in the real makefile. Then the command above will change to:

```Makefile
checksum=$$($(MD5) $$file | awk '{ print $$1 }') ; \
```

And you'll need that `awk` command in there which will work with the output of either format of the MD5 utility, which was my real motive in suggesting the change before.

Makefile
```diff
@@ -136,6 +137,24 @@ $(TEST): $(SOURCE) $(SPECS) $(BIN)
...
+	sed -e '/%SCRIPTS%/ {' -e 'r .make-script-block' -e 'd' -e '}' app.html.in > app.html
+
+.make-script-block: $(SOURCE)
+	> .make-script-block
+	for file in $(SOURCE); do \
+		checksum=$$($(MD5) $$file | awk '{ print $$1 }') ; \
+		echo "<script src='$$file?v=$$checksum'></script>" >> .make-script-block ; \
+	done
+
+# Rule to determine MD5 utility
```
robes: could you move this toward the top where we are setting up the other variables? I'd suggest putting it just after we define "BIN". thanks

howdyjessie: Did you mean I should move the MD5 conditional to the top? (It wouldn't make sense to put the app.html rules to the top right?)

robes: right, just the MD5 conditional (starting at line 151)

PR #65, opened 2015-09-26
Chaise detail

bennettl:
Added "record" folder into chaise UI and updated entity links to point it record


robes:
The list of files to be merged includes `app.html` however in the latest master we have dropped app.html and replaced it with `app.html.in`. Maybe this is just getting picked up accidentally.

Also, we don't want your changes to `chaise-config.js` because those are local to your setup.

PR #143, opened 2015-10-21
Chaise#135 Use ermrestLocation for Chaise Record Base URL.

shinyichen:


robes:
checked it out locally and it works for me

PR #165, opened 2015-12-24
Testing specs for starter tests a, b

mashuaihh:
#### Progress
Succeeded in testing and getting same results of starter test c, d on Dec.22. 
But today it is not working. Still working on it.
Might need a little help from developers who implemented front end features so that I can learn more about how the project works, and debugged my tests accordingly.

Merry Xmas


test/e2e/search/protractor.conf.js
```diff
@@ -14,6 +14,6 @@ exports.config = {
   },
   // If ng-app attribute is in a descendant of <body>, tell Protractor where ng-app is
   rootElement: '#main-content',
-  // PTOR_BASE_URL should be http://dev.facebase.org for now.
-  baseUrl: process.env.CHAISE_BASE_URL + '/data/search'
+  // CHAISE_BASE_URL should be http://dev.misd.isi.edu/chaise for now.
+  baseUrl: 'http://dev.misd.isi.edu/chaise' + '/search'
```
robes: Let's not use the hard coded base URL. But the old form was a problem to. I'd suggest making it simply: 
```
baseUrl: process.env.CHAISE_BASE_URL
```

mashuaihh: Ok, I changed it to 
```
baseUrl: process.env.CHAISE_BASE_URL + '/search'
```
I remember Jessie told me this is the preferred way, because later '/record' and '/login' may be concatenated to the CHAISE_BASE_URL. 

Do you think it's alright?

robes:
It looks fine but my env isn't working so I can't test it. Going to reassign to Jessie.

PR #169, opened 2016-01-09
set up CHAISE_BASE_URL in .travis.yml [ci skip]

mashuaihh:
The reason the test keeps failing is because the CHAISE_BASE_URL has not been set up in .travis.yml.
The env variable is missing when Chaise is tested using Travis CI.

I believe it will be working with this commit.

Thanks


mashuaihh:
Just synced with upstream

PR #184, opened 2016-01-23
Angular ui grid in detail page

mikedarcy:
Replacement table/grid for detail page.  Can be enabled by setting the following value in chaiseConfig,js:

detailUiGridEnabled: true

Functionality should be equivalent to existing code with the exception of embedded links.  Links in cells can be supported by ui-grid but I need some clarification on the logic behind the existing algorithm because it seems to add them based on some heuristic which is not clear to me (and often seems to do the wrong thing).  Putting link support back in should be pretty easy once we figure out they should be handled.


robes:
There are a few issues that would have to be resolved before this could be merged (IMO).

1. I'm seeing a problem where the table doesn't load when you first expand the label for it. If I keep expanding / collapsing the table it eventually loads.

2. There are options on the column drop down that don't work, or at least they don't work for the data types in those column. Specifically those are the various aggregates (like count, min, max, etc). Those should be hidden, for those columns that they don't work on.

3. Export as PDF does not work.

4. Transpose is badly broken. When you search on the column heading, the behavior is invalid.

5. Most importantly, although there are no "merge conflicts" when you do merge it with master, the record page does not work. I did not debug it deeply but the behavior is that you get the raw template (displaying the angle brackets all over), which indicates the page has failed somewhere.

mikedarcy:
Most of these issues are correctable, though some might be out of our hands, like PDF export or other built-in features "not-working".  If you could point me to a reference deployment showing the behavior you found, that would help me while I try to recreate the problems.

The reason the column calculation stuff doesn't work properly is that the column metadata doesn't have types assigned yet. Some extra coding needs to be done to map from ermrest-returned type to the type expected by ui-grid.  Should not be a big issue to correct for the normal mode. 

It is going to be an issue in transpose mode because you cannot assign a type other than string to each column, which is why column-based search doesn't make sense in transpose.  Also, you cannot toggle the filter box on a per-column, only globally.  It should be possible to disable all filter boxes on columns while in transpose mode, and re-enable in normal mode.

robes:
Right. I understand that some of these issue may be out of our hands. Still, I'm not sure we can deploy it for users if there are broken features. Seems like that will cause confusion. IMHO

To test it, I put it on our shared test server in my user dir.

Search page to start from:
https://dev.misd.isi.edu/~schuler/chaise-ui-grid/chaise/search/#3/legacy:target_gui?layout=list&page=1

An example record page:
https://dev.misd.isi.edu/~schuler/chaise-ui-grid/chaise/record/#3/legacy:target_gui/id=5&site=iHuman

BTW, I didn't run into my first cited issue above. Maybe that's some weird transient issue.

karlcz:
Looking at those examples, I'd say the transpose feature simply does
not work from a UX point of view... the entire HTML/CSS styling needs
to transpose for the user to understand what is happening:

1. Column headers down the left should be styled as headers, not data
2. Column headers should not be in the scrollable region
3. Column controls should also run vertically
4. Odd/even "row" shading should run vertically
5. Column visibility controls (in the hamburger menu) need to show columns and not record IDs

mikedarcy:
I'm looking into some of the issues now.  For now it can just stay in a branch until it is a bit more stable.

mikedarcy:
@karlcz These are not really things that I can easily change.  I made the transpose function work within the limits of the control, it is pretty much a hack.  The control does not support the idea of transposed data like this, and modifying to do the things you want (which are reasonable) is probably a lot of work to modify the control itself. I don't think its practical to support transpose within this control.

carlkesselman:
Should we revisit the need for transpose at all?  It seems to me that the main use was that if we had a table with a lot of columns, this would be a way to get it usable.  Now with the heuristic for displaying this inline, it would seem the value of this is diminished.

Carl

----------------------------------------------------------
Dr. Carl Kesselman
Dean’s Professor, Epstein Department of Industrial and Systems Engineering
Fellow, Information Sciences Institute
Viterbi School of Engineering

Professor, Preventive Medicine
Keck School of Medicine

University of Southern California
4676 Admiralty Way, Suite 1001, Marina del Rey, CA  90292-6695
Phone:  +1 (310) 448-9338
Email:  carl@isi.edu<mailto:carl@isi.edu>
Web: http://www.isi.edu/~carl








On Jan 26, 2016, at 2:11 PM, karlcz <notifications@github.com<mailto:notifications@github.com>> wrote:

Looking at those examples, I'd say the transpose feature simply does
not work from a UX point of view... the entire HTML/CSS styling needs
to transpose for the user to understand what is happening:

1. Column headers down the left should be styled as headers, not data
2. Column headers should not be in the scrollable region
3. Column controls should also run vertically
4. Odd/even "row" shading should run vertically
5. Column visibility controls (in the hamburger menu) need to show columns and not record IDs

carlkesselman:
I should add that I don’t think our current implementation of transpose does any of this stuff…..

Carl

----------------------------------------------------------
Dr. Carl Kesselman
Dean’s Professor, Epstein Department of Industrial and Systems Engineering
Fellow, Information Sciences Institute
Viterbi School of Engineering

Professor, Preventive Medicine
Keck School of Medicine

University of Southern California
4676 Admiralty Way, Suite 1001, Marina del Rey, CA  90292-6695
Phone:  +1 (310) 448-9338
Email:  carl@isi.edu<mailto:carl@isi.edu>
Web: http://www.isi.edu/~carl








On Jan 26, 2016, at 2:11 PM, karlcz <notifications@github.com<mailto:notifications@github.com>> wrote:

Looking at those examples, I'd say the transpose feature simply does
not work from a UX point of view... the entire HTML/CSS styling needs
to transpose for the user to understand what is happening:

1. Column headers down the left should be styled as headers, not data
2. Column headers should not be in the scrollable region
3. Column controls should also run vertically
4. Odd/even "row" shading should run vertically
5. Column visibility controls (in the hamburger menu) need to show columns and not record IDs

karlcz:
The "new heuristic" is to enable the transposed view automatically which presumes that there is a tranposed view implemented and available.

As I understand Mike, the ui grid simply lacks the underpinnings to support this. In my opinion, transposed views are essential UX in many bioinformatics scenarios, and so this ui grid widget is insufficient and unworkable.

karlcz:
@carlkesselman Sounds like I should file bugs against the current one too then...

carlkesselman:
Yes.  It is a bit rough in my opinion.

So it seems to me that we have two things that we are saying are of use:

1) I think that it is important to streamline finding things in the table, hence search, sort, and selecting columns are quite useful.
2) Karl thinks that transposing rows and columns is useful.  I’m not sure I understand the driver for this.  Can you give me an example?

There is also the case where the nested table is an extension to the entity, (1xN), in which case, we probably shouldn’t present this as a table at all but rather as a potentially grouped set of attibributes.

Carl

----------------------------------------------------------
Dr. Carl Kesselman
Dean’s Professor, Epstein Department of Industrial and Systems Engineering
Fellow, Information Sciences Institute
Viterbi School of Engineering

Professor, Preventive Medicine
Keck School of Medicine

University of Southern California
4676 Admiralty Way, Suite 1001, Marina del Rey, CA  90292-6695
Phone:  +1 (310) 448-9338
Email:  carl@isi.edu<mailto:carl@isi.edu>
Web: http://www.isi.edu/~carl








On Jan 26, 2016, at 3:08 PM, karlcz <notifications@github.com<mailto:notifications@github.com>> wrote:


@carlkesselman<https://github.com/carlkesselman> Sounds like I should file bugs against the current one too then...

carlkesselman:
Can you give me an example of where transposed views would be essential?

Carl

----------------------------------------------------------
Dr. Carl Kesselman
Dean’s Professor, Epstein Department of Industrial and Systems Engineering
Fellow, Information Sciences Institute
Viterbi School of Engineering

Professor, Preventive Medicine
Keck School of Medicine

University of Southern California
4676 Admiralty Way, Suite 1001, Marina del Rey, CA  90292-6695
Phone:  +1 (310) 448-9338
Email:  carl@isi.edu<mailto:carl@isi.edu>
Web: http://www.isi.edu/~carl








On Jan 26, 2016, at 3:04 PM, karlcz <notifications@github.com<mailto:notifications@github.com>> wrote:

The "new heuristic" is to enable the transposed view automatically which presumes that there is a tranposed view implemented and available.

As I understand Mike, the ui grid simply lacks the underpinnings to support this. In my opinion, transposed views are essential UX in many bioinformatics scenarios, and so this ui grid widget is insufficient and unworkable.

karlcz:
@carlkesselman I'd say that the majority of the time we ask biologists
to share data or show us spreadsheets they have developed in
collaborations, they always create transposed tables. What we would
consider the attribute labels are running down the left side and each
column in the array of data represent one record.  Often, but not
always, the record IDs are also highlighted in header/label styling
across the top.

robes:
I am okay with dropping the transpose feature.

karlcz:
@robes I don't understand how to reconcile that with #185

mikedarcy:
Re: PDF export

This can be made to work.  It doesn't work currently because I did not include some required dependencies.  However, I am currently questioning the utility of the export feature in general due to the fact that different projects may have different security concerns, and might not want to allow export of data because of DUA reasons.  

Anyway I am disabling this feature for now, but if we want to move forward with this control we can always get this feature working properly and then allow it to be enabled/disabled on a per-deployment basis (e.g., via chaise-config).

robes:
@karlcz right, #185 gets dropped by this. we have to do something else for that scenario. in the short term it would mean we can't support that. we're all flailing a bit here. really need to step back and make a plan for what to support.

mikedarcy:
I think I've addressed most of Rob's initial issues, with the exception of transpose. 

In transpose mode I've simply disabled the column filter and context menus, since it is unlikely those will ever work properly without modifications to the control code.

Let's re-evaluate and try to come to a consensus on whether this is worth investing more time into.

mikedarcy:
Merging this into master.  Will be disabled by default.  Can enable by setting the following in chaise-config.js:

recordUiGridEnabled: true

PR #187, opened 2016-01-26
Restructure E2E test files

mashuaihh:
Restructure E2E test files and more tests.


mashuaihh:
### Set up e2e file structure, added tests to check record page.

This might be a bug (or to be implemented features?) :
When testing on https://dev.misd.isi.edu/chaise/search/, clicking on some "edit filters" in sidebar attributes selector, ex. clicking **"Wnt1-Cre;Tgfbr2fl/fl;Alk5fl/+"** in **"MOUSE GENOTYPE"**, will result in a window alert message, which will fail the test.

I tried doing the same thing on https://facebase.org/data/search/, everything works well there, so it might be just an issue on the dev site.

robes:
@svoinea we should look into Shuai's comment above. Maybe a bug in chaise/search/?

svoinea:
Added the "AllowEncodedSlashes On" directive in the ssl.conf file.
This seems to have fixed the problem.
This directive is also present in the wsgi_ermrest.conf and wsgi_hatrac.conf files, but somehow, as a known issue, it is not always undertaken from there.

PR #191, opened 2016-01-28
Use url pattern for main entity in record page. Don't link null values.

shinyichen:


shinyichen:
Test links:
https://dev.misd.isi.edu/~jchen/chaise/record/#1/legacy:dataset/id=4894 
https://dev.misd.isi.edu/~jchen/chaise/record/#1/legacy:dataset/id=4833 (for null values)

PR #195, opened 2016-02-02
Data Browser Tour

RFSH:
This pull request adds a tour to data browser page.


chaise-config-sample.js
```diff
@@ -7,5 +7,14 @@ var chaiseConfig = {
     showBadgeCounts: false,
     recordUiGridEnabled: false,
     recordUiGridExportCSVEnabled: true,
-    recordUiGridExportPDFEnabled: true
+    recordUiGridExportPDFEnabled: true,
+    tour: {
```
howdyjessie: Not sure if this is the right place to add tour info; we should avoid putting specific, custom things in here. What's this object for and can it be moved elsewhere?

scripts/tour.js
```diff
@@ -0,0 +1,333 @@
...
+      $(".field-enable.sidebar-toggle:eq(1)").click();
+    }
+  }, {
+    element: '#resultstable .panel:eq(0)',
+    title: "Dataset Details",
+    content: "By clicking on title of each dataset, you can see its details.",
+    placement: "right",
+    onPrev: function(tour) {
+      searchChosenAttr.click();
+      $(".search-box input:eq(2)").val(searchInputValue).trigger('input').keyup();
```
howdyjessie: Is it possible to use tour-specific CSS selectors to get elements? Class names and IDs can change and break the tour.

scripts/tour.js
```diff
@@ -0,0 +1,333 @@
...
+  var chosenValue = false; // for choosing value step.
+  var searchChosenAttr = false; // for searching attribute step.
+  var searchInputAttr = false; // for searching attribute step.
+  var searchInputValue = false; // for searching value step.
+
+  extraAttr.siblings("div").find('input:checked').click(); // removes extra attr from list of attrs
+
+  var tourSteps = [{
+    element: "#navcontainer .sidebar-nav",
+    title: "Choose Attriutes",
```
howdyjessie: Tiny spelling error :)

chaise-config-sample.js
robes: Actually, the tour isn't custom. It's going to be a general site tour feature for chaise.

scripts/tour.js
RFSH: I don't think so. I have to specify elements that I want to show in each step.

chaise-config-sample.js
howdyjessie: Ah okay; does the tour currently start every time Chaise Search loads? How do people feel about a flag to enable/disable the site tour?

RFSH: No, Users have to click on a button to start the tour.

scripts/tour.js
robes: Aref, the suggestion here was to add new class names that have nothing to do with the styling or functionality of the UI application. Just things to use as the selectors for use with the site tour instead of referencing things that are used by the app code (and could therefore be changed in the future).

robes:
@crisaless I don't see any arguments against merging this. Could @RFSH please update the PR so that there are no merge conflicts. (Sorry, our delay in merging this has meant that new code in master makes this PR conflict.) It is probably easy and minor conflicts.

scripts/tour.js
RFSH: I tried my best not to change other pages and just use the existing class names to avoid conflict. But I can add tour specific selectors to these elements if you want.

RFSH:
I updated the PR and fixed the conflicts.

PR #205, opened 2016-02-16
Skipped some tests and fixed bugs

mashuaihh:
### Skipped some tests for now.
I will implement more tests when it is decided what to be tested next.

Sometimes error like this (https://saucelabs.com/beta/tests/f4e938face2f432998a744b0bf22c1f6/watch) occurred. It happened twice when I was using Chaise on my browser (Chrome) and met it several times when doing automation tests.

screen shot 2016-02-16 at 3 00 55 pm

### DIFF
--- a/test/e2e/chaise.page.js
+++ b/test/e2e/chaise.page.js
@@ -135,7 +135,8 @@ function resultContent() {
 function recordPage() {
     this.entityTitle = $('#entity-title');
     this.findEntityKeyByName = function (entityName) {
-        return element(by.cssContainingText('.entity-key.ng-binding', entityName));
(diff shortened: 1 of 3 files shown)


karlcz:
Looking at the server logs, this 500 error seems to be one of two known sources of 500 error that can come out of our web server platform (Apache HTTPD, mod_wsgi, and web.py) independent of our own custom web service logic. It is a transient error.

The workaround is for the AJAX client to treat these like a 503 Service Unavailable and retry with some kind of limited back-off logic for cases where retry never succeeds. We should always do this in Chaise or any of our other web clients.

PR #234, opened 2016-03-10
Small changes to make test work for now

mashuaihh:
## Small changes

Small changes in test/chaise.page.js (a page object).
Modified E2E tests to make them pass on new Chaise features (introduced since my last Pull Request).

I might spend some time making the tests become more general so that new features can be tested against more easily.

Thanks!


mashuaihh:
The CI build fails and I think it might be because of some possible misconfiguration of SauceLabs user account (I think Chaise is using Jessie's SauceLabs account so I cannot check that). I will check with Jessie to see how to fix it.

PR #314, opened 2016-04-27
Make e2e tests facebase-independent

mashuaihh:
Still one spec file to finish


hongsudt:
@robes should I assign this to @chiragsanghvi ?

robes:
@hongsudt sure I think that makes sense to assign to Chirag.

Issue #325, opened 2016-04-29
Referenced tables shows "No rows found"


howdyjessie:
The test case says: "Showing all referenced tables unless table has 0 rows," but sometimes I'll expand a table and it'll say "No rows found."

For example, when looking at [this record page](https://staging.gpcrconsortium.org/chaise/record/#1/gpcr_browser:construct_gui/title=HUMM-1049) on GPCR staging, I expanded the Protein table and saw "No rows found."

shinyichen:
Please see pull request #326

howdyjessie:
Tested out #326 and it worked. The PR has been merged.

PR #326, opened 2016-04-29
Empty table

shinyichen:
This is a quick fix for issue #325. Indicate (0) in the reference table header if table has no rows.


howdyjessie:
Tested out @shinyichen's branch and can confirm #325 is now fixed. Moving forward with merge.

PR #329, opened 2016-05-03
Ermrestjs refactor

jrchudy:
Refactor work has finished on this branch as well as tying in the core of a unit testing framework.


howdyjessie:
@jrchudy Does `npm-debug.log` have to be merged into `master`?

viewer/annotations/annotations.controller.js
```diff
@@ -21,7 +21,7 @@
         vm.createAnnotation = createAnnotation;
         vm.cancelNewAnnotation = cancelNewAnnotation;
 
-        vm.editedAnnotation = null; // Track which annotation is being edited right now
+        resetEditedValues();
```
howdyjessie: @jrchudy Did you mean to assign resetEditedValues to an instance of the controller here?

howdyjessie:
@jrchudy It looks like most things work except for:

1. Creating, editing, and deleting annotations. I'm still getting `Uncaught Type Error: Cannot read property '0' of undefined` in `osd.annotorious_ui.js`, despite your temp stub. If the only reason this error appears is because we haven't changed Annotorious yet then we can circle back to this issue after we've changed Annotorious.

2. When I create a comment and then click the pencil icon to edit it, the comment doesn't turn into an input field.

jrchudy:
npm-debug.log can be removed. I intended resetEditedValues() to be a private method for the controller.

The `property 0 of undefined` issue is associated with trying to redraw the annotation based on the coords. I couldn't figure out how to fix that. When we fix annotorious that issue should go away. The locations that I modify the object for annotorious are all `var stub` and marked with a `// TODO` comment so we should be able to fix it easily.

I will look into the comment box not changing to an input field.

PR #374, opened 2016-05-19
Data entry error handling

jrchudy:
Pull Request to merge data entry error handling into master. To test this functionality, you need to force an error to occur that is not related to data validation. What I did was remove the defaults fro the post request so that a 409 conflict would be returned.


common/errors.js
```diff
@@ -0,0 +1,18 @@
+(function() {
+    'use strict';
+
+    angular.module('chaise.errors', ['ui.bootstrap'])
+
+    // Factory for each error type
+    .factory('ErrorService', ['$injector', function ErrorService($injector) {
```
howdyjessie: `$injector` probably not needed anymore so we're not using `$uibModal` anymore.

howdyjessie:
I tested by removing defaults and passing in extra columns to force a 409. This PR works according the behavior described in the [wiki's entry on error handling](https://github.com/informatics-isi-edu/chaise/wiki/Towards-common-error-handling): 
> display error, upon clicking ok, lead them to the data entry form with existing state and validation errors. (MVP)

@jrchudy There's only one small bit, which is removing the now-unnecessary $injector. See code comment above.

jrchudy:
I removed the $injector library. Also found out putting `$` escapes comments for commits because my commit last night was `removed $injector` not just `removed`.

Issue #352, opened 2016-05-13
Hide columns in data-entry app


karlcz:
The data-entry app should honor the `tag:misd.isi.edu,2015:hidden` annotation on columns as well as the draft `tag:isrd.isi.edu,2016:ignore` annotation which we are designing with the next MVP in mind...

1. The basic hidden one should apply to both the data-entry form and any resulting record display.
2. The new ignore annotation should allow us more precision to omit the field from the data-entry form while including it in the resulting record displays.

When we hide/ignore a column for data-entry, we should specify it in the `defaults` query parameter to tell the server that we wish it to fill in the column for us.

howdyjessie:
I've created PR #378 for this issue. Description of changes and test info is on the PR.

howdyjessie:
The PR has been reviewed and merged into `master`.

PR #378, opened 2016-05-21
Hide columns in data entry app

howdyjessie:
This pull request fulfills #352. It adds a feature that hides a column and its input controls if the column has either a `tag:misd.isi.edu,2015:hidden` or `tag:isrd.isi.edu,2016:ignore` annotation. If a column has the `2016:ignore` annotation and the annotation is an empty array, `null`, or has the `entry` context, then the column is hidden.

Columns that are hidden/ignored are also added to the `defaults` parameter when submitting form data so that the sever will fill in those columns for us.

##### To Test:
This feature is live on [synapse-dev](https://synapse-dev.isrd.isi.edu/~jessie/chaise/data-entry/#2/Zebrafish:Subject). The Zebrafish:Subject table has both `tag:misd.isi.edu,2015:hidden` and `tag:isrd.isi.edu,2016:ignore` annotations on the `ID` and `Sub-sequence` columns. The code can be pulled from [`de-hide-columns#352`](https://github.com/informatics-isi-edu/chaise/tree/de-hide-columns%23352).

### DIFF
--- a/common/interceptors.js
+++ b/common/interceptors.js
@@ -7,7 +7,6 @@
 
         var responseInterceptor = {
             response: function(response) {
-                console.log("Success", response);
                 return response;
             },
             responseError: function(error) {
--- a/data-entry/form.controller.js
+++ b/data-entry/form.controller.js
@@ -30,6 +30,7 @@
         vm.isAutoGen = isAutoGen;
         vm.isForeignKey = isForeignKey;
         vm.matchType = matchType;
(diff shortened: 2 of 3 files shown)


karlcz:
I can confirm that this works the way I expect on my use case. Should assign within UI team to review code and merge though...

howdyjessie:
Okay I'm assigning to @robes for code review.

PR #390, opened 2016-05-24
Fixed record app login

shinyichen:


karlcz:
I verified that this works for me on synapse-dev.

karlcz:
@shinyichen I think you can either merge this yourself or ask @svoinea to review the code change first...?

PR #399, opened 2016-05-27
Record recordset updates

shinyichen:
* Error handling for recordset
* Common navigation bar
* Recordset login/logout
* URL encoding/decoding fixes for record and recordset
* Put URI encoding to common util file.


karlcz:
I'm testing this with the record-recordset branch you pushed for me in chaise, not the fork which is the source of this pull-request...  the specific tests I could think of seem to work now.

PR #403, opened 2016-05-27
Record recordset

shinyichen:
* Error handling for recordset
* Common navigation bar
* Recordset login/logout
* URL encoding/decoding fixes for record and recordset
* Put URI encoding to common util file.


robes:
This looks good. I'll merge it. But I am curious why you chose to make the Errors a conventional javascript module rather than an angular model (like you did with the UriUtils).

shinyichen:
The errors I followed the convention in ermrestjs filters and utilities. When I did the UriUtils I was following what Josh is doing with his common js files. Should I do it the angular way then?

robes:
I think so. I would do it the angular way in this case because it is a module of Chaise which is an angular app. In ermrestjs, we keep it plain javascript because we expect the api to be (possible to be) used in other environments, not just angular.

Issue #362, opened 2016-05-16
Record app doesn't handle unauthorized response consistently


karlcz:
1. Visit a record page where login is automatically started due to ermrest 401 unauthorized response
2. Use logout link at top
3. Use extra logout links from Globus page to logout of everything
4. Close extra Globus tab that has opened
5. Navigate "back" in your browser to the record page

The page displays  `Not Found. We're sorry, we ran into an internal server error when querying for entity references`.  The server log shows that the application actually made requests like:

- GET /ermrest/authn/session  resulting in 404 Not Found
- multiple GET /ermrest/catalog/1/aggregate/.../row_count:=cnt(*)  resulting in 401 Unauthorized

robes:
I just followed the above steps that Karl specified above and I cannot see that the issue has been fixed. I would assume that the fix would be if you went back in step 5 above that you would be redirected to login. But it still display the `Not Found...` message.

I used this link for testing.
https://synapse-dev.isrd.isi.edu/chaise/record/#1/Zebrafish:Subject/Issuer=Dsy&Sub-Sequence=A&Birth%20Date=2015-12-31

robes:
This worked for me on dev.isrd.

PR #408, opened 2016-05-31
Fixed record app session handling #362

shinyichen:
Record app was not handling login if session not found.


robes:
where would i go to test this? I tried a couple links I could think of and it didn't work

https://dev.isrd.isi.edu/chaise/recordset/#4/rbk:image/
https://dev.rebuildingakidney.org/chaise/recordset/#1/rbk:image/

PR #427, opened 2016-06-07
Navbar directive

howdyjessie:
A second/third attempt at the `navbar` directive:
- It accepts 3 new attributes: `server` (required), `brand-image`, `brand-text`.
- Login and logout functions are still housed inside the directive.
- User's `full_name` and `email` remains supported alongside `display_name`.
- It no longer forces a login if session not found.
- It no longer requires `scripts/ermrest.js` or a value provider called `context`.

##### To Test #####
Here is the navbar directive being used on the [data entry app](https://dev.isrd.isi.edu/~jessie/chaise/data-entry/#1/legacy:dataset) and the [recordset app](https://dev.isrd.isi.edu/~jessie/chaise/recordset/#1/legacy:dataset). As far as I can tell, these are the only two apps using the directive at this time.
The code can be pulled down from the `navbar-dir` branch.


robes:
I'm merging anyway, but I think the above comment would be an improvement. IMO

howdyjessie:
@robes It's been renamed.

PR #437, opened 2016-06-08
Test setup

chiragsanghvi:
@robes 

- Implements import utility to create catalog, schema and tables and add entities for the same.
- Makes E2E tests Facebase independent.


robes:
We would like to review this work early next week, according to @hongsudt . We will wait until then to decide on the PR.

Quick observations and comments:
1. we decided we do not want to include the `node_modules` code and should pull this in based on the appropriate `package.json` configuration.
2. we should not include the FaceBase schema and definitely not the FaceBase database which is basically serialized in json in this request.

PR #438, opened 2016-06-08
Error handling for Dataentry

jrchudy:
Error handling for data entry and some common components as well.


jrchudy:
This is for reviewing the code changes as it isn't necessarily easy to test the error cases that are handled.

robes:
It looks okay. Its hard to internalize it. I wish there wen't so many try/catch blocks. Ideally we could consolidate some of them. But I can't offer any specific suggestions. May not be possible.

jrchudy:
I couldn't think of a way to consolidate them but I could look into it a bit more.

karlcz:
1. Some of the try/catch seem like they should be replaced with a simpler conditional code like:
```
  var annokey = 'tag:misd.isi.edu,2015:vocabulary';
  if (key in ftable.annotations) {
     ... do stuff with ftable.annotations[key]...
  }

instead of:

  try {
     var vocabAnnotation = ftable.annotations.get("tag:misd.isi.edu,2015:vocabulary");
  } catch (error) {
     // no vocab annotation, do nothing
  }
```
2. Can we have a type hierarchy?  Rather than having all this stuff to catch and re-throw a more specific exception, why don't we just have ermrestjs throw the more specific exceptions as subtypes of the more generic exceptions it throws now?  E.g. it certainly knows if it is the annotations subobject that raise NotFound and could raise AnnotationNotFound instead...
3. Rather than many error callbacks that have empty body or just a log statement, shouldn't there just be one reusable callback function that gets passed by name in these callsites?
4. Similarly, if there are repeated try/catch patterns that just do logging or something else really generic, those could be abstracted in a function that takes a thunk/callback argument and executes it in that common try/catch block.  All the other callsites can then reuse this common wrapper instead of duplicating the same boilerplate code.
5. Having stubbed out catch blocks or callbacks that do nothing useful and have a comment like "nothing for now" are hard to read and maintain.  Later on, do we know this was supposed to eat the exception and treat it as success, or is it an outline for code that still needs to be written before this will work correctly...?

robes:
@karlcz 
1. is pythonic not javascript. `ftable.annotations` is a class. We could add a `.in(...)` method to it to get this functionality. Seems like we should
2. I think @shinyichen did that and then took it out? Do we still have that? If not we should re-add it so that the exceptions are specific and yes in a type hierarchy and then we can collapse some of this. But you still need to the try/catch around each annotation.get (if you don't first do a annotations.in) because we can't break out of the code for an expected error like that.
3. and 4. i didn't understand hope Josh did.
5. Definitely agree.

karlcz:
Hmm, this works in firefox web console:

- 'foo' in {'a': 1}  --> false
- 'foo' in {'a': 1, 'foo': 2} --> true

robes:
@karlcz Sorry, yeah, you're obviously right. It occurred to me later that I was thinking about something else. We aren't able to use the native arrays for the annotations collections objects because javascript doesn't let us overload the square bracket operators, so we have those custom collections-like classes. With that said, we have to add a `.in(...)` function to do this still.

jrchudy:
1. As @robes noted, `.in` is not available, but it could be added to ermrestJS and use that instead of using `try/catch` logic for simple cases that should fail gracefully and don't need to be caught.

2. I agree that there should be some sort of type hierarchy for exceptions. ErmrestJS should be able to see the request and if it’s for an annotation, it should return an AnnotationNotFound exception. And we should do this for common exceptions across apps, notably the places that I looked for `exception instaceof Errors.NotFoundError` for table/schema/catalog.

3. The catch statements that do nothing could all have a function called that is something like a `genericExceptionCaught()` with a logging function defined. But some cases, notably with an annotation not being found, that won't do anything. It's unnecessary to log for that but it does need to be caught.

4. I realized I did this with each of the `schema/table/catalog/annotation NotFound()` functions. @howdyjessie said the same thing and I wanted to leave them separate for now because we might want to have some separation of logic for these later. (maybe if the input table/schema/catalog is invalid, we redirect to a page with available schemas/tables/catalogs for the user to select from and that logic would be different for each. I don't want it to be assumed those 3 cases should be handled the same way)

5. On [this](https://github.com/informatics-isi-edu/chaise/wiki/Towards-common-error-handling) wiki page, we outline the different cases for error handling. I agree it is hard to read and maintaining it in the future will be us wondering why it's even there.

jrchudy:
1. @robes, would you like this functionality in ermrestJS? I can go and make those changes and test it before this PR is merged.

2. I began to address this but left those sections as commented `TODOs`. We would have to implement this in ermrestJS first. Given that we have the generic exception code in ermrestJS, I think the hierarchy should be defined there. Furthermore, ermrestJS could then return the correct sub-type for the thrown exception.

3/4. I have cut down the number of `try/catch` blocks in `dataEntry.app.js` and added a generic case that is meant to handle if an exception happens. This generic block has a TODO in it that is waiting for an exception hierarchy to be defined.

5. I'm hoping this can be addressed by adding the `.in (...)` functionality for point 1 and the hierarchy of exceptions mentioned in point 2.

robes:
@jrchudy You could still merge this PR and continue working on the changes that should be pushed into ermrestjs and the updates to Chaise that will follow that. I rather not hold up branches of code for long. I leave it up to you and @howdyjessie to do the merge though. Code looks fine to me but I'm not in position to run it somewhere and test it.

jrchudy:
@howdyjessie, do you want to take a look at the changes and see if there is anything else that I may have missed?

howdyjessie:
@jrchudy Looks fine to me!

Issue #402, opened 2016-05-27
Add edit link from record pages


karlcz:
Add a sensible edit button/link from record page to the new record-editing mode of the data-entry app.

howdyjessie:
I made PR #439 for review and assigned @karlcz, @hongsudt, @robes for review.

howdyjessie:
Updated PR #439 again. Please have a look when you can.

hongsudt:
@karlcz can you please review this?

karlcz:
@hongsudt I commented on the PR itself and think I am waiting for further revisions to review and test.

howdyjessie:
@karlcz @robes @hongsudt PR #439 is ready for review again.

hongsudt:
@howdyjessie PR #439 reported a conflict...

karlcz:
@howdyjessie I commented on the PR again.

karlcz:
I confirmed this is working in master and am closing this issue.

PR #439, opened 2016-06-08
Add edit link to record page (#402)

howdyjessie:
**Changes**
- The link redirects user to the recordedit (data entry) app.
- The redirect url is determined via `"../recordedit"` + `window.location.hash`.
- The Edit link only appears if:
 1. The catalog ACL for `content_write_user` has an `*` or matches one of the user's session attribute IDs; AND
 2. Either the table or schema doesn't have an applicable `ignore` annotation; AND
 3. `chaiseConfig.editRecord` is `true`

**To Test**

1. Visit a Record page in my user dir, such as [this record from Yang Chai](https://dev.isrd.isi.edu/~jessie/chaise/record/#1/legacy:dataset/id=5776). Next to the Permalink link, click the "Edit Record" link. You should be redirected to the recordedit (data entry) app where the input fields are pre-filled with this record's values.
2. Visit a Record page that has an `ignore` annotation on its table, like [this dummy record in the `testtoo` table](https://dev.isrd.isi.edu/~jessie/chaise/record/#4/rbk:testtoo/id=1). Next to the Permlink link, there should **not** be any "Edit Record" link.

The code can be pulled down from branch `record-edit-btn`. 

~~Normally I wouldn't have worked on a separate branch for this relatively small enhancement, but I thought it would take more work than was actually necessary when I started.~~ Well now that we have the ACLs, this is no longer the case!


karlcz:
Shouldn't the redirect be using a relative URL or otherwise figuring out the entire path?  This hard-coded absolute path means you cannot have several Chaise deployments and keep them separate from each other.

howdyjessie:
@karlcz That's true, I didn't think about the possibility of multiple Chaise deployments. The redirect url is now created via `".../data-entry"` + `window.location.hash`, based on the assumption that a Chaise app lives on same level as other apps (i.e. `/chaise/data-entry|recordset|record ...`).

I also switched to this relative url in the recordedit/data-entry app, which redirects to record or recordset upon form submission.

karlcz:
I tested this and looked at the code and it makes sense. However, I think @robes mentioned wanting to do some ACL check to decide whether to display the edit button.  Do we want that before we merge the PR?

robes:
Right @karlcz, this was underspecified... We should check whether the user is in the `content_write_user` ACL before displaying the Edit button. I think it would add confusion for deployments that are public read-only sites, even to display the Edit button.

@howdyjessie This would be a good opportunity to make a change in ermrestjs because I don't think it exposes the `meta` resource. Can you talk with @shinyichen about what you need to do and send her a PR with the changes?

howdyjessie:
Okay I've updated this PR. Now the edit link only appears if the catalog ACL for `content_write_user` has an `*` or matches one of the user's session attribute IDs.

karlcz:
I think we decided that this feature should obey the `ignore` annotation and hide the edit button if the table or schema is set to ignore in contexts of `edit` and `entry`.

howdyjessie:
From today's standup: In addition to the catalog ACL check, the record app must additionally check whether the current table has an `ignore` annotation with an `entry` context with an `edit` sub-context. If so, don't show the edit link. Otherwise, show the link.

howdyjessie:
I've updated the top of the thread to reflect the additional check for an `ignore` annotation at the table level in terms of changes and test info.

karlcz:
I verified that this honors the `tag:isrd.isi.edu,2016:ignore` annotation on the table, but it does not seem to observe the same annotation at the schema level.

According to our [specification](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-ignore) this annotation should also apply to whole schemas, i.e. every table in the schema. I think this should be further defined to follow the same override characteristics we've said for other annotation tags. The logic should be something like:

1. Default ignore list is `[]` i.e. nothing ignored
2. Schema-level annotation (if any) replaces the ignore list
3. Table-level annotation (if any) replaces the ignore list

Thus a schema-level annotation could be `["entry"]` meaning that most tables in the schema will have no data-entry enabled. But, if one table has its own ignore list `["create"]` then that table would expose an edit record button.  Does that make sense?  I can update the annotation spec if this sounds good...?

howdyjessie:
@karlcz Agh I missed the "or schema" part of your previous comment! My bad.

But yes this hierarachy makes sense to me; it's like CSS, where rules with greater specificity take precedence over those with lower specificity. 

So for this issue, when checking whether to display the Edit link, is it the best/quickest way to do that by:
1. Checking the catalog ACL for write access, and if the authenticated user is allowed write access;
2. Then checking the table for a relevant `ignore` annotation, and, if I don't find one;
3. Then checking the schema for a relevant `ignore`?

howdyjessie:
I've pushed the code that also checks for schema for any ignore annotations and updated the top of the thread to reflect this addition.

robes:
I'm not sure about the status of this PR. I tried the first link "this record from Yang Chai" and I get the following error after clicking on the edit link from the record page.
recordediterror

howdyjessie:
@robes Sorry about the red ugliness. I had uploaded code from another branch into my user dir when you were testing. The testing info remains the same. Can you try again?

hongsudt:
@karlcz Can you please review this up to the synapse use case?

karlcz:
I tried to test the `record-edit-btn` branch on my machine and get an error `app.js:1290:17  SyntaxError: expected expression, got ')'`.  This is using master branch of everything else.

howdyjessie:
@karlcz It turns out I botched a merge conflict; can you test this branch again?

karlcz:
It doesn't crash anymore but I am not seeing an edit button even though I think the ACL allows it and the chaise config enables it.

karlcz:
Thanks to @howdyjessie who clarified the config option is `editRecord` not `recordEdit` and that I had a stray `2016:ignore` annotation suppressing the edit link, I can now see the edit record link and see that it is disabled if I set the `editRecord` config to `false`.

However, the actual edit page is not showing any existing values when it loads. Every field is blank. At the same time, it doesn't show the `+` button so I think it knows it is in the edit mode rather than new record mode...?

karlcz:
By the way, this other issue may be a bug in the recordedit app since the URL looks OK to me (with a predicate restricting it to the existing record ID).  So, it might be right to merge this PR and chase that other bug separately...

howdyjessie:
@karlcz When going to the [recordedit page that the edit links directs you to](https://synapse-dev.isrd.isi.edu/chaise/recordedit/#1/Zebrafish:Subject/ID=ZfDsy20151231A), the console says a couple ermrestjs files are missing (`errors/networkerrors.js`, `errors/validationerrors.js `), which is making a number of error messages clog up the console. I checked that those two files are in ermrestjs' `master`, so could you re-pull it and see if that changes anything? (I'd do it myself but I see unstaged changes in the folder and I don't want to mess up anything.)

Issue #433, opened 2016-06-08
Check validity when duplicating last record


howdyjessie:
When working in multi-row entry, the user can click "Add Record" to duplicate the most recently added row. If that row has an invalid value, however, that value is not copied into the duplicated row.

hongsudt:
@karlcz @robes what do you think about complete duplication even when the cell values are invalid?  To me, this will make it easier for the users to understand the app behavior.

robes:
I do not understand the description.

howdyjessie:
@robes I made [a short video](http://recordit.co/BLipUTbPgN) to demonstrate. The duplicated row doesn't have the value from `C_smallint` while it does have the value from `C_int`.

howdyjessie:
From today's standup: Only allow user to duplicate a form column if that form column has no invalid values. If that form column has 1+ invalid value, disable the "Add Record" button and/or display an alert.

hongsudt:
I would prefer to get an error box telling me what to do than having the button disabled.

howdyjessie:
There are now 2 buttons for adding a record to the form.
1. "Add Record" just adds a blank row/column.
2. "Duplicate Last Record" copies the last row/column if the last row/column has no invalid values. Otherwise, the app will display an error instead.

These two buttons should disambiguate what a singular "Add Record" would do.

You can see the feature live [here](https://dev.isrd.isi.edu/~jessie/chaise/data-entry/#1/public:form_test). This has already been push into `master`. [This commit](https://github.com/informatics-isi-edu/chaise/commit/06294b50946290e12e9ffd7a3e34effae8148f18) is the relevant diff to look at for code review.

karlcz:
I don't understand why we would go from 1 to 2 buttons.  Where did this requirement come from?

howdyjessie:
@karlcz I added the 2nd button of my own accord. This is not a strict requirement and is certainly up for debate.

karlcz:
I don't like the extra buttons.  I think it makes it cluttered and confusing for a naive user.  Perhaps, if these could be reduced to idiomatic icons w/ a tooltip and better button placement, it might be OK.

Right now, the buttons compete for attention with the `Submit Data` button.  If they were instead down next to the form it might make sense:
- A `Copy` button (usually some kind of binary figure with arrow and small plus sign?) closely aligned with the record that will be copied and on the right, i.e. to the right corner while the `Remove` button is centered over the record.
[...]

PR #452, opened 2016-06-17
Modify Copy and Remove button behaviors

howdyjessie:
This PR brings two updates:
- #433: The copy button duplicates only the most recently added row/column in the data entry/recordedit app. But before it attempts the duplication, the app first checks that the most recently added row/column doesn't have any invalid values entered by the user.
- The delete button that's used to delete an individual row/column from the data entry form is now displayed via a new Chaise-wide `deleteLink` directive. This directive displays a UI control that performs a deletion routine that you pass in. If the `confirmDelete` property in `chaise-config.js` is `true`, the directive will display a confirmation modal before running your deletion routine. More details about usage can be found at the top of [`delete-link.js`](https://github.com/informatics-isi-edu/chaise/blob/recedit-copy-row/common/delete-link.js#L1-L22).

**To Test**
These changes can be seen in my [user dir](https://dev.isrd.isi.edu/~jessie/chaise/recordedit/#1/legacy:dataset). The code can be pulled down from this branch.

Tagging @shinyichen and @jrchudy so they know about this new directive.


hongsudt:
@howdyjessie I need some training to get me started with how to review the pull request..

howdyjessie:
@hongsudt No problem, it's not difficult to get started! I can swing by your office when you've got some spare time.

howdyjessie:
@hongsudt I fixed the issue that caused the test URL to break. (It was because I forgot my user dir's files reflected the files needed for a separate PR but not this one.) I've updated the [test URL](https://dev.isrd.isi.edu/~jessie/chaise/recordedit/#1/legacy:dataset) at the top of this thread to reflect data entry's name change to recordedit. Can you try again?

PR #459, opened 2016-06-21
changing the includes function to contains

jrchudy:
Per a conversation I had with @robes, I changed the `.includes(...)` function to `.contains(...)`. [Here](https://en.wikipedia.org/wiki/Element_(mathematics)#Notation_and_terminology) is the noteworthy information.

"Logician George Boolos strongly urged that "contains" be used for membership only and "includes" for the subset relation only."


jrchudy:
Assigning to @robes for review and merge.

howdyjessie:
Interesting subtlety here! I didn't know about "contains" vs "includes".

Issue #442, opened 2016-06-08
Support 2015 Default annotation if the schema and table are missing from the url


hongsudt:

jrchudy:
https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2015-default , attaching the associated wiki page detailing the use case for this annotation.

jrchudy:
@robes, for data-entry how do we want to handle this? @hongsudt, @howdyjessie, and I discussed it after the division meeting. 

If a user types in a url that is incomplete (`...\data-entry\` or `...\data-entry\#1` or so on), do we want to infer the table they want to go to by looking at the `Default` annotation for catalog, schema, and table? Or should we be sending the user back to the landing page where they can then click on a valid data-entry table?

One other though that came to mind was giving them a selector for the data that is missing. Say they navigate to `...\data-entry\`, should they get a catalog selector? Or do we want to show them a list of the available `...\<catalog>\<schema>:<table>` combinations that the user is allowed to edit (submit to) This goes for any amount of missing info. So if they navigate to `...\data-entry\#1`, should they get a schema selector? Or do we want to show them a list of available `\schema:table` combinations that they can navigate to. And just to exhaust the considered cases. Should they get a table selector if they navigate to `....\data-entry\#1\Zebrafish` or see a list again of the ones they are allowed to edit?

This is thinking about ACL permissions at the same time.

robes:
I would favor one of two approaches here:
1. if they don't specify the `<schema>:<table>` then use the one specified in the Default annotation, or
2. just give an error page.

I would not do more than that, like giving a schema and table selector or redirecting back to a landing page. To me it just seems like this is more of an error to ever get to this page without it being a proper link.

I would be fine with this app just giving an error message to the user. But if we think we need to support Default annotation then I suppose we can do that.

jrchudy:
I agree with the case of data-entry that we don't want to infer a table the user wants to input data for. I would assume a user is only trying to go to data-entry to input data for a specific catalog, schema, and table. For data-entry showing error messages should be enough to notify the user that they are trying to navigate to somewhere they didn't expect and it wouldn't be useful getting defaulted to a page they didn't even know about. (redirect might just confuse them more)

Added record and recordset tags to this because it also applies to them.

hongsudt:
For the error message, upon clicking ok, it should direct me to the landing
page.  This error handling approach should be in the common module.

karlcz:
[...]

PR #467, opened 2016-06-21
Default behavior for recordedit

jrchudy:
When a user navigates to recordedit without a catalog, schema, or table defined they are given an error dialog and then redirected to the landing page. This is meant to address issue #442 as the default behavior when catalog, schema, or table is not present in the URL.

This can be tested on my [synapse dev machine](https://synapse-dev.isrd.isi.edu/~jchudy/chaise/data-entry/#200/Zebraish:Subject). That URL points to catalog `200` which does not exist. The same type of dialog can be seen for the case of an improper schema or table name as well or if all of that information is missing, [for instance](https://synapse-dev.isrd.isi.edu/~jchudy/chaise/data-entry/).

Assigning to @robes and @howdyjessie to review the code changes and merge the pull request.


howdyjessie:
@jrchudy If I'm logged in, the error dialog appears and clicking OK takes me back to the landing page. But if I'm logged out, the error dialog doesn't show up, and then I see an error about `UriUtils` in the console:

screen shot 2016-06-22 at 12 38 43 pm

jrchudy:
This is a side effect of the refactoring where we have been re-doing the logic for unauthorized. I believe we talked about dealing with 401 by having ermrestJS try to log the user in, but I could be wrong.

jrchudy:
I updated the code to include the Chaise catchAll function that also handles the login flow if the user is not logged in and is required to. They are redirected to the login page where they can choose which login provider they want.

This resolves the issue that Jessie noted when looking at the code. This can be tested [here](https://synapse-dev.isrd.isi.edu/~jchudy/chaise/recordedit/#2/Zebrafish:Subject) by first making sure you are logged out before navigating to this page.

Makefile
```diff
@@ -220,6 +220,7 @@ RE_SHARED_JS_DEPS=$(JS)/vendor/jquery-latest.min.js \
 	$(COMMON)/utils.js \
 	$(COMMON)/validators.js \
 	$(COMMON)/navbar.js \
+	$(COMMON)/templates/errorDialog.controller.js \
```
robes: seems odd to see a controller.js in a 'templates' subdirectory?

recordedit/recordEdit.app.js
```diff
@@ -201,23 +196,24 @@
                 // } else if (exception instanceof Errors.SchemaNotFoundError) {
```
robes: probably don't need to preserve the commented code

robes:
This all looks fine to me and I tried both of the links (s/data-entry/recordedit). I made a couple minor comments in the code. This looks fine to merge to me.

common/errors.js
```diff
@@ -4,7 +4,42 @@
...
+            }
+
+            var modalInstance = $uibModal.open({
+                templateUrl: '../common/templates/errorDialog.html',
+                controller: 'ErrorDialogController',
+                controllerAs: 'ctrl',
+                backdrop: 'static',
+                keyboard: false,
+                resolve: {
+                    params: function() {
```
howdyjessie: Is it necessary to return `params` via this function instead of just
```
resolve: {
    params: params
}
```
?

common/templates/errorDialog.controller.js
```diff
@@ -0,0 +1,15 @@
+(function() {
+    'use strict';
+
+    angular.module('chaise.errors')
+
+    .controller('ErrorDialogController', ['$uibModalInstance', 'params', function ConfirmDeleteController($uibModalInstance, params) {
```
howdyjessie: For consistency, it might be better to change `ConfirmDeleteController` to `ErrorDialogController`.

common/errors.js
```diff
@@ -4,7 +4,42 @@
...
+                        return params;
+                    }
+                }
+            });
+
+            modalInstance.result.then(function () {
+                $window.location = chaiseConfig.dataBrowser ? chaiseConfig.dataBrowser : $window.location.origin;
```
howdyjessie: If we're truly redirecting (are we? cc @robes?) and not just going to another page, then it should be `$window.location.replace(url)`. [window.location vs window.location.replace()](http://stackoverflow.com/questions/1865837/whats-the-difference-between-window-location-and-window-location-replace)

jrchudy:
Assigned back to @howdyjessie to make sure changes look good. I did a small refactor for viewer by removing `confirmDelete.controller.js` and instead using `modalDialog.controller.js` in the common folder. Also moved `confirmDelete.html` template to the `common/templates` folder. 

Refactored changes for [recordedit](https://synapse-dev.isrd.isi.edu/~jchudy/chaise/recordedit/#2/Zebrafish:Subject). Try changing the url parameters or removing them completely. As well as being logged out when trying to navigate to the page.

Refactored changes for [viewer](https://dev.rebuildingakidney.org/~jchudy/chaise/viewer/#1/rbk:image/id=7). Make a comment and try to delete it. As well as Make  new annotation and try to delete that with and without a comment. All of this should still work.

howdyjessie:
Everything looks good to me; I'm merging this PR.

Issue #466, opened 2016-06-21
Common library function for parsing URL fragments


robes:
One of the things that should be refactored in Chaise and we can start on this with the record2 app, is to factor out the URL (fragment) parsing into the common library.

I also have a proposal on the table to write a parser in ermrestjs, something I’ve considered doing for a long time, so that we have one parser for URL fragments. See this and comments are welcomed.

https://github.com/informatics-isi-edu/ermrestjs/issues/55

As for Chaise, I still think it should have a common function for parsing URL fragments, even if its just a wrapper around this (to be developed) ermrestjs data path parser. For instance, there will be chaise specific options to be parsed. We should also keep in mind that each app may have some app specific options of its own, so a common library function need to accommodate that.

robes:
This should also help to address issue #465.

RFSH:
I noticed that we are using `window.location.origin` in the code that is used for URL parsing. 

Since [it is not supported by IE11- and has some issues with IE11](https://developer.mozilla.org/en-US/docs/Web/API/Window/location#Browser_compatibility), we should make sure that it is defined:

```javascript
//source: http://tosbourn.com/a-fix-for-window-location-origin-in-internet-explorer/
if (!window.location.origin) {
  window.location.origin = window.location.protocol + "//" + window.location.hostname + (window.location.port ? ':' + window.location.port: '');
}
```

PR #485, opened 2016-06-30
Url parser

jrchudy:
The code in this pull request satisfies the requirements for issue #466. A common URL parsing function was created and the logic refactored out of `viewer`, `recordset`, and `recordedit`. 

Assigning the Pull Request to @robes, @howdyjessie, and @shinyichen for review. Each app can be tested on my dev user directory code. @shinyichen, could you look at [recordset](https://dev.isrd.isi.edu/~jchudy/chaise/recordset/#1/legacy:dataset/id::gt::100&id::lt::13000@sort(summary)) and make sure functionality is still as expected. I made sure all the context values were still valid. And could you make sure there wasn't anything I missed that needs to be considered when parsing `recordset` URLs.

@howdyjessie, both [viewer](https://dev.rebuildingakidney.org/~jchudy/chaise/viewer/#1/rbk:image/id=7) and [recordedit](https://synapse-dev.isrd.isi.edu/~jchudy/chaise/recordedit/#2/Zebrafish:Subject) still behave as we expect. Could you also make sure I didn't miss anything with the parse function?


common/utils.js
```diff
@@ -5,19 +5,105 @@
...
+        * @desc
+        * converts a string to an URI encoded string
+        */
+        function fixedEncodeURIComponent(str) {
+            return encodeURIComponent(str).replace(/[!'()*]/g, function (c) {
+                return '%' + c.charCodeAt(0).toString(16).toUpperCase();
+            })
+        }
+
+        function parseURLFragment(context) {
```
robes: I'm okay with this interface. But if I were designing this as an api, I would have made it:
```
function parseURLFragment(location, context) 
```
And not have it dependent on $window directly. I would allow `context` to be `null` and if the caller didn't pass that parameter create a new `context` object. And then the function would always return `context` (whether a new instance or the passed in instance). Caller could use it like:
```
var context = UriUtils.parseURLFragment($window.location);
```
or
```
/* I already have a mycontext */
UriUtils.parseURLFragment($window.location, mycontext);
```

recordedit/recordEdit.app.js
```diff
@@ -21,47 +21,16 @@
...
-            var params = parts[1].split(':');
-            if (params.length > 1) {
-                context.schemaName = decodeURIComponent(params[0]);
-                context.tableName = decodeURIComponent(params[1]);
-            } else {
-                context.tableName = decodeURIComponent(params[0]);
-            }
-        }
+        // Parse the URL
+        context = utils.parseURLFragment(context);
```
robes: `context = ...` is not necessary (unless I'm mistaken) and therefore confusing.

viewer/viewer.app.js
```diff
@@ -17,41 +17,26 @@
...
-            context.serviceURL = chaiseConfig.ermrestLocation;
-        }
-
-        var hash = window.location.hash;
-
-        if (hash === undefined || hash == '' || hash.length == 1) {
-            return;
-        }
+        // Parse the URL
+        context = utils.parseURLFragment(context);
```
robes: `context = ...` not needed again

robes:
Overall, I think this is looking good. Made a couple few minor comments inline.

jrchudy:
I updated the code with Rob's suggestions and verified it was still working as I expected from before.

PR #489, opened 2016-07-05
E2e test

chiragsanghvi:
Apart from that I needed to create a new branch as the older one had an issue with Travis. Following is the description from the previous pull request.

@robes  

- Separated config for server, cookie, etc. 
  - excluded cookie from the config. Fetched from the environment variable
- display log : to prevent TravisCI from terminating
- document best practices for how to use describe, where to put things, etc. 
  - dependency, how to install, saucelab variables  [Protractor Configuration](https://github.com/informatics-isi-edu/chaise/blob/e2e_test/doc/protractor.md)
  - chaise readme: how to test
  - [test configuration](https://github.com/informatics-isi-edu/chaise/blob/e2e_test/test/e2e/README.md)
- independent npm package to [import data](https://www.npmjs.com/package/ermrest-data-utils) and its [ReadMe](https://github.com/chiragsanghvi/ErmrestDataUtils/blob/master/README.md) 
- separate out the data dependent vs data independent 
- some utility function to make the test per deployment easier (config, etc.)


robes:
Maybe I missed this earlier? But at this point the tests are failing and it conflicts with master. (Sorry if I missed this)

jrchudy:
@chiragsanghvi, could you update the links? Besides that, the tests look good. I'm not sure what @robes's intended structure for this was, but I like the way that you separated the specs (based on app and then based on data dependency). @RFSH, could you glance over these changes and see if anything sticks out to you that should be fixed?

I'm guessing these tests are passing although the Travis CI build is failing. Could you check on that

chiragsanghvi:
@jrchudy I will update the links once I am done changing the documentation. Its outdated for now. 
Need to look into the failing testcases, its a resource issue which the browser can't fetch. Will fix it and let you know.

RFSH:
Everything looks good to me and the documentation cover most of the things that a new developer needs to know on how to add new e2e tests. However, There are two things that are worth mentioning:
- I think we should keep all the e2e test suits inside a parent `suits`(or any other name) folder; instead of adding them directly inside `e2e` folder.
- In test cases `chaise.page.js` has been used, but the documentation doesn't clearly mention how we can expand and reuse it.

chiragsanghvi:
I have made the folder changes and now the tests are passing. I will be updating the wiki for e2e test.

hongsudt:
@chiragsanghvi let's merge this request..

Issue #492, opened 2016-07-07
navbar: use chaise config for brand image and brand text


hongsudt:
This is currently being done in search app, but not the new navbar. 

The new navbar doesn't look into the chaise config and if we need to change the branding on all apps, each app needs to define them separately.

PR #495, opened 2016-07-07
navbar: Use chaise config for brand image, brand text, and brand url

RFSH:
Looking at [chaise-config.js](https://github.com/informatics-isi-edu/chaise/blob/master/chaise-config-sample.js), there are three variables that are used for the navbar:
- `headTitle`
- `navbarBrand`
- `navbarBrandImage`

This PR adds these variables to the common navbar (issue: #492).

@robes @jrchudy @howdyjessie Can you please review this PR?


jrchudy:
The changes look good to me and non-intrusive. If the config options aren't set, the UI will still show properly.

Issue #487, opened 2016-07-02
viewer: add # of matches on top of search result


hongsudt:
This should help with usability e.g. if n is large, users need to scroll down.

howdyjessie:
This issue has been addressed in PR #496, which is now under review.

Issue #483, opened 2016-06-29
viewer: reduce some white space in the chaise annotation


hongsudt:
The description is not often used during the annotation. Is there a way to shrink the white space, so more annotations get displayed on the screen.

howdyjessie:
This issue has been addressed in PR #496, which is now under review.

Issue #481, opened 2016-06-29
viewer: an on/off switch to show/hide visible annotations


hongsudt:

howdyjessie:
This issue has been addressed in PR #496, which is now under review.

Issue #480, opened 2016-06-29
viewer: update annotations in open-seadragon to be consistent with search result


hongsudt:
The open-seadragon can be in any zoom level, @meihuisu Is there a way to automatically zoom out so that all the search results will be displayed in the open-seadragon?

howdyjessie:
This issue has been addressed in PR #496, which is now under review.

PR #496, opened 2016-07-07
Sync annotation visibility from Chaise to OSD

howdyjessie:
This PR brings some new features to the Image Viewer app; it addresses issues #480, #481, #483, and #487.

- #480: 
 - The annotations visible in OSD are now consistent with search results on the left panel. If an annotation appears/disappears on the left panel, it also appears/disappears in OSD, _but not the other way around_.
 - Searching through annotations in the left panel is no longer instantaneous. Clicking the new search (🔍 ) button or pressing Enter while the search input is focused executes the search.
 - Clicking the Reset button clears the search input, and all annotations become visible in both the left panel and OSD.
- #481: A new Show/Hide toggle button has been added above OSD. Clicking this button toggles the visibility of annotations currently displayed in the left panel.
- #483/style tweaks: Each annotation "card" is now a little more compact. Particularly, if an annotation has no description, the description area's margins are now collapsed.
- #487: A number of annotation matches is now displayed above the search results. It's basically a count of currently visible annotations in the left panel.

**To Test**
Visit [this page](https://dev.rebuildingakidney.org/~jessie/chaise/viewer/#1/rbk:image/id=1) on dev rbk to test these new features. This branch can also be pulled down to your own user directory for testing.

### DIFF
--- a/viewer/annotations/annotations.controller.js
+++ b/viewer/annotations/annotations.controller.js
@@ -16,6 +16,7 @@
         vm.resetSearch = resetSearch;
         vm.filterAnnotations = filterAnnotations;
         vm.sortSectionsFirst = sortSectionsFirst;
+        vm.setVisibility = setVisibility;
 
         vm.createMode = false;
         vm.newAnnotation = {config:{color: vm.defaultColor}};
@@ -93,7 +94,11 @@
         });
 
         function filterAnnotations(annotation) {
+            var typeIsVisible = vm.filterByType[annotation.type]; // boolean that answers "are annotations of a certain type visible?"
             if (!vm.filter) {
+                if (typeIsVisible) {
+                    annotation.config.visible = true;
+                }
                 return true;
             }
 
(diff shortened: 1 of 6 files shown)


viewer/annotations/annotations.controller.js
```diff
@@ -116,12 +121,19 @@
                     var numCommentProps = commentProps.length;
                     for (var p = 0; p < numCommentProps; p++) {
                         if (commentProps[p] && commentProps[p].toLowerCase().indexOf(vm.query) > -1) {
+                            if (typeIsVisible) {
+                                annotation.config.visible = true;
+                                AnnotationsService.syncVisibility();
+                                vm.getNumVisibleAnnotations();
```
jrchudy: I see this piece of code in 3 places. Maybe pull out the function to a private one in this controller.

jrchudy:
I have two comments about your code:

- When you go to the viewer page, the show/hide button says `show`. When you click it, it changes to `hide` and nothing in OSD changes. When you click it again, it flips to `show`, and the annotations in OSD are hidden. Clicking `show` this time, shows the annotations. It seems like the button should appear `hide` when the page loads.

- My other comment is the code comment. Basically you have a block of code in an `if` statement duplicated 3 times. Might make sense to create a private function to the controller for that chunk of code.

howdyjessie:
@jrchudy Okay I fixed the two issues you mentioned. Can you test again?

Issue #88, opened 2015-10-02
Record view (aka Details page) claims catalog does not exist, though it does


robes:
Not sure why but on Safari the details / record page doesn't load the data while testing on Firefox it does. I'll attach a screen shot showing the url, the page display, and the error console.

Not sure if this could be related to #87 .

![chaise-record-404](https://cloud.githubusercontent.com/assets/4202842/10251427/7927b6c8-68e5-11e5-9c3c-ac767aa0ba79.jpg)

bennettl:
Works for me now in Safari
![screen shot 2015-10-02 at 10 56 28 pm](https://cloud.githubusercontent.com/assets/1672224/10261688/de5d70be-6958-11e5-9923-cd0a70964ee4.png)

PR #506, opened 2016-07-18
Ermrestjs revisions

jrchudy:
The corresponding changes to the revised APIs in ermrestJS. This uses the `resolve` function and `Reference` object with server side (ermrestJS) parsing. Also an example of promise chaining so we don't have multiple error callbacks (and only need 1 catch). 

The related PR in ermrestJS is [#88](https://github.com/informatics-isi-edu/ermrestjs/pull/88). This can be tested [here](https://dev.isrd.isi.edu/~jchudy/chaise/record-two/#1/legacy:dataset/id=269).


jrchudy:
This should be up to date completely with the changes from @robes's suggestions in [PR 88](https://github.com/informatics-isi-edu/ermrestjs/pull/88).

jrchudy:
The changes in this PR coincide with the changes made to ermrestJS to include Reference, Page, and Tuple objects. Those changes are captured by [PR 100](https://github.com/informatics-isi-edu/ermrestjs/pull/100) in ermrestJS. The code can be tested [here](https://dev.isrd.isi.edu/~jchudy/chaise/record-two/#1/legacy:dataset/id=269) in my local dev environment.

PR #507, opened 2016-07-19
Add app name param to .getServer() calls

howdyjessie:
This is related to informatics-isi-edu/ermrestjs#69.

For the record2, recordset, recordedit, and viewer apps, each of these apps now pass in a second argument, the app's name, in its respective `ermrestFactory.getServer(...)` call so that ermrestjs can append the app names to the relevant async calls.

**Question**: I currently store an app's name inside its `context` provider (`$rootScope.context` for record2), and then pass in `context.appName` into the `.getServer` call. Is this sufficient? Should I create a mapping in `common` so that app names aren't as closely coupled with each app's code? Another way?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,8 @@ HTML=search/index.html \
 	 recordset/index.html \
 	 matrix/index.html \
 	 viewer/index.html \
-	 recordedit/index.html
+	 recordedit/index.html \
+	 record-two/index.html
 
 # ERMrestjs Deps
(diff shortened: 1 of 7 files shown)


jrchudy:
I think that having it on the context object per app is sufficient. Each app is only going to use it's own name and not need to know about the others.

howdyjessie:
Talked to @hongsudt, who also believes this is sufficient for now. We can come back this topic as needed.

Issue #465, opened 2016-06-21
Support disjunctive filters


howdyjessie:
Currently the recordset app supports conjunctions but not disjunctions. The recordedit/data-entry app needs the disjunctive filters in order to display a recently submitted set of records. #461 has more related info.

howdyjessie:
@hongsudt This issue needs to be completed before #461, which in turns should be completed soon [according to](https://github.com/informatics-isi-edu/chaise/issues/461#issuecomment-227520749) @karlcz.

karlcz:
The ermrest disjunctive operator is `;` so I suggest we use the same syntax here to avoid confusion, i.e. `.../recordset/Schema:Table/Col=val1;Col=val2` with the same decoding logic as elsewhere:

1. Split the filter path on `/` to get a list of expressions which are to be conjoined
2. Split each expression on `;` to get a list of predicates which are to be disjoined
3. Split each predicate by looking for the `=` operator (or the `::` _opname_ `::` pattern for other supported operators)
4. URL decode the left-hand side column name and the right-hand side value

Of course encoding the permalink follows these steps in reverse, URL encoding each column name or value before gluing predicates and expressions back together.

hongsudt:
@howdyjessie can you take on this task or does it have to be @shinyichen ?

howdyjessie:
@hongsudt No I can do it too. :)

hongsudt:
We decided at the meeting today that @howdyjessie will work on this.

PR #509, opened 2016-07-19
Parse complex filters in URL (#465)

shinyichen:
This update added to the common URL parsing to handle complex filter parameters.

### DIFF
--- a/viewer/viewer.app.js
+++ b/viewer/viewer.app.js
@@ -31,10 +31,11 @@
         // should we allow for improper URLs here?
         // what if there are 2 filters and the id filter is the second one.
         // Is that improper or should it be parsed and ignore the other filter?
-        if (context.filter.type === "BinaryPredicate" &&
-            context.filter.operator === "=" &&
-            context.filter.column.toLowerCase() === "id") {
-            context.imageID = context.filter.value;
+        var filter = context.filter;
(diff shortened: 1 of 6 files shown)


shinyichen:
I made changes to recordEdit and viewer apps, so including @howdyjessie @jrchudy here.

jrchudy:
Changes look good to me. I used a lot of this code in the parse function I wrote in ermrestJS and I'm sure there's some refactoring that can be done that we can look at. @howdyjessie do the changes for viewer look okay to you? I'm not sure if we need to be so precise, but they work.

viewer/viewer.app.js
```diff
@@ -31,11 +31,11 @@
         // should we allow for improper URLs here?
         // what if there are 2 filters and the id filter is the second one.
         // Is that improper or should it be parsed and ignore the other filter?
-        angular.forEach(context.filters, function(filter) {
-            if(filter.name.toLowerCase() === "id") {
-                context.imageID = filter.value;
-            }
-        });
+        if (context.filter.type === "BinaryPredicate" &&
```
howdyjessie: Tiny nit here but I see `context.filter` accessed multiple times and it's slightly better to just store that in a variable for quicker access.

howdyjessie:
Looks good to me too :) Merge at will.

Issue #461, opened 2016-06-21
RecordEdit should show inserted records after submit


karlcz:
The ERMrest entity POST operation responds with the actual records that were created, including server-generated keys when the client is using the `defaults` query parameter.

After submit, the data-entry app should go to a recordset page filtered by the new row keys.  Currently, it seems to go to an unfiltered page which is confusing, particularly once the new rows won't show on the first page of results.

karlcz:
@hongsudt I think this improvement should be done soon as it affects usability for pilot projects.  I.e. I think it is a higher priority than the other new apps in the pipeline.

hongsudt:
@jrchudy @howdyjessie one of you should take a look at this.

karlcz:
BTW this feature will depend on the recordset feature to accept filters in the URL.  It exists but is a little buggy as described in #464

howdyjessie:
>After submit, the data-entry app should go to a recordset page filtered by the new row keys. Currently, it seems to go to an unfiltered page which is confusing, particularly once the new rows won't show on the first page of results.

@robes, @jrchudy, and I have discussed this bit before but we landed on simply redirecting to the table, unfiltered, because recordset doesn't support `::leq::` and `::geq::` yet. (Are `::leq::` and `::geq::` even the right operators to use in this situation?)

karlcz:
No, exact equality is the right operator.  It should be a disjunction of several keys, not a range.  (You cannot assume that server-generated keys are in contiguous and non-overlapping orders for multi-row insert...)

howdyjessie:
It turns out recordset doesn't support disjunctive filters yet, so this issue depends on #465.

hongsudt:
@shinyichen is this ready to be reviewed?

hongsudt:
@karlcz I think this has been fixed, so I assigned you to review.

karlcz:
I did a test on synapse-dev and doing a multi-insert still sends me to recordset without any filters so this is *not* done.

howdyjessie:
Reassigned to @shinyichen.

howdyjessie:
@karlcz I just tested and merged @shinyichen's PR #521, which should redirect properly now. Can you test on synapse-dev?

karlcz:
works for me...

PR #521, opened 2016-08-01
RecordEdit redirect to recordset with disjunctive filters #461

shinyichen:
Support disjunctive filters in RecordEdit redirect #461


howdyjessie:
I've tested the redirect and it works.

PR #526, opened 2016-08-02
Add markdown-container to record app

RFSH:
In order to customize the css for markdown, I added the `markdown-container` class to app.css. 

Record app will attach this class to the elements that are `markdown` typed and therefore they have the correct styles. Take a look at [this example](https://dev.isrd.isi.edu/~ashafaei/chaise/record/#2939/resource:markdown/id=123).

For now, the only css rules are, showing bullets in ordered and unordered lists. New rules can be added.

### DIFF
--- a/record/assets/views/record.html
+++ b/record/assets/views/record.html
@@ -38,7 +38,7 @@
                                 {{value | sanitizeValue }}
                         </a>
                         <div ng-class="{'fixed-font' : entity.sequences.indexOf(key) > -1, '' : entity.sequences.indexOf(key) == -1}">
-                            <span ng-if="!entity[key + '_link'] || value === null" ng-bind-html="value | sanitizeValue" ng-class="entity.colDataTypes[key]=='markdown'?'markdown-container':''"></span>
+                            <span ng-if="!entity[key + '_link'] || value === null" ng-bind-html="value | sanitizeValue" ng-class="'markdown-container': entity.colDataTypes[key]=='markdown'"></span>
                         </div>
                     </td>
                 </tr>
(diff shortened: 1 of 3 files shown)


hongsudt:
This looks nice..

hongsudt:
@chiragsanghvi @howdyjessie can you review this pull request and do the merge?

record/assets/views/record.html
```diff
@@ -38,7 +38,7 @@
                                 {{value | sanitizeValue }}
                         </a>
                         <div ng-class="{'fixed-font' : entity.sequences.indexOf(key) > -1, '' : entity.sequences.indexOf(key) == -1}">
-                            <span ng-if="!entity[key + '_link'] || value === null" ng-bind-html="value | sanitizeValue"></span>
+                            <span ng-if="!entity[key + '_link'] || value === null" ng-bind-html="value | sanitizeValue" ng-class="entity.colDataTypes[key]=='markdown'?'markdown-container':''"></span>
```
howdyjessie: This is more a style question/preference, but why not use: 
`ng-class="'markdown-container': entity.colDataTypes[key]=='markdown'"`? I find it easier to read and you don't need the "else" branch this way.

RFSH:
Thanks to @chiragsanghvi, we found a bug in record app. Basically when display name annotation was defined on the column, the record app couldn't show the appropriate tooltip and column type.

 The bug is resolved now. [This](https://dev.isrd.isi.edu/~ashafaei/chaise/record/#2985/error_schema:valid_table_name/id=4) is the latest code with the sample data that Chirag has created. Both the tooltip and markdown styles of `Column Name` are showing.

I also changed the ng-class statement based on @howdyjessie suggestion.

hongsudt:
Thank you all for speedy bug fixes and reviews..

PR #527, opened 2016-08-03
Table directive

jrchudy:
Assigning to @howdyjessie and @chiragsanghvi. The table directive has been built with basic functionality. The sort functions are there but not working yet as I haven't been able to test it with ideal data or fetching data from the related reference objects.

There is a section in `record.app.js` that is commented out, if you want to see what the tables look like, uncomment that code and re-upload it. This is a hacked set of data that is identical to the table above. 

There was some cleaning up of code as well. I learned more about directives and revisited `record.js` where I defined the `tableDisplay` directive. I removed the `link` function and `transclude` property. My understanding of `transclude` was wrong before.

If you have any questions, don't be afraid to ask. I will check periodically tonight and throughout the week to see if either of you have questions.


jrchudy:
Most recent commits include the end-to-end test case for the table directive on the record page (the related tables). You can see the latest [here](https://dev.isrd.isi.edu/~jchudy/chaise/record-two/#1/legacy:dataset/id=1568).

jrchudy:
All the tests pass when I run them locally. Not sure why they keep failing in travis, when I look they are tests completely unrelated to anything that I changed. I didn't change any of the data files or other specs.

chiragsanghvi:
Its because of some json parsing error. You can see the whole test flow [here](https://saucelabs.com/beta/tests/0a4907781ed04b3d866a582f3ecbef41/watch)

sauce labs test details

jrchudy:
I added an expectation to my test case to check for the order of the foreign keys. Made changes to `record.app.js` to make sure data was input into the data array in the proper order because of asynchonous callbacks. There won't be anymore changes.

@chiragsanghvi and @howdyjessie can you look at the table directive code and the test cases one last time.

common/table.js
```diff
@@ -0,0 +1,129 @@
...
+                tableName:          null,
+                tableDisplayName:   null,
+                columns:            [],
+                filter:             null,
+                sortedBy:           null,
+                sortOrder:          null,
+                rowset:             null,
+                keycols:            [],
+                count:              0
+            };
```
howdyjessie: General programming question: I've always wondered, when is it good to list out things and space them out such that all the values in an object are aligned, such as in `scope.tableModel`? Is it just a matter of preference?

common/templates/table.html
```diff
@@ -0,0 +1,20 @@
+<div class="outer-table">
+    <table class="table table-striped table-bordered table-hover">
+        <!-- sortable column header -->
+        <thead>
+            <tr>
+                <th nowrap ng-repeat="col in tableModel.columns">
+                    <span class="table-column-displayname" ng-click="sortby(col.name)">{{col.displayname}}</span>
```
howdyjessie: Should this be `ng-click=sortBy`? (I'm assuming it's referring to the `scope.sortBy` in your directive.

howdyjessie:
It looks fine to me. 👍

common/table.js
jrchudy: This purely a matter of preference. This is infinitely more readable, in my mind, than having a single space after each `:`.

common/templates/table.html
jrchudy: ah yes, but none of it works right now, so it doesn't matter really..

Issue #498, opened 2016-07-09
Recordedit shows existing null date as epoch zero (1969-12-31) instead of blank


karlcz:
When using recordedit to edit records that have optional date columns, an existing null date field is displayed as  "0 seconds from unix epoch" date value, e.g. `1969-12-31` instead of showing a blank value.

If the user doesn't notice this and submits a revision, the null will get replaced with that bogus value instead of remaining null as before.

- Description updated 2016-08-02 after re-testing chaise/master

karlcz:
I'm not sure if I made a mistake observing this, or if a fix was done already.  I now see that I can delete the date field and hit submit and end up with a NULL value in ERMrest as desired.

However, on reopening the record with recordedit, it again *displays* the "0 seconds from unix epoch" date value `1969-12-31` instead of showing a blank/null value.  This will then submit bad data the next time if I don't manually clear it again.

Should I open a separate issue for this or just revise this issue description?

howdyjessie:
@karlcz I'm not aware intentional efforts to fix this bug, but seeing as there's no commit activity associated with this issue, a revision of the description should suffice.

hongsudt:
@chiragsanghvi can you take a look at this?

karlcz:
I confirmed that the null date is now preserved in the entry field when opening an existing record to edit it.

Issue #502, opened 2016-07-13
When adding a row/column to the form for boolean fields, the most recently added row/column should be duplicated and appended to the form. 


chiragsanghvi:
@howdyjessie @jrchudy 

This doesn't works for boolean field if you select an empty value in the first column and then duplicate it.

For example, if you select an **empty** value for the boolean field `is luxurious` for a record and then duplicate it, you will find **false** set for the `is luxurious` field for the second row.

![data entry app](https://cloud.githubusercontent.com/assets/1658152/16819972/a942a468-4903-11e6-81a7-1e57637024d6.png)

karlcz:
I confirmed that the new row copies the previous record's boolean state properly now.

Issue #503, opened 2016-07-14
Recordedit sets Boolean fields as "" instead of null.


chiragsanghvi:
@jrchudy @robes @howdyjessie 

**Scenario**: 

1. Open  this [RecordAdd](https://dev.isrd.isi.edu/chaise/recordedit/#1337/product:accommodation) page.
2. Open browser console and run following script to enable edit
```javascript
$('.modal').remove(), $('.modal-backdrop').remove(), $('body').removeClass('modal-open');
```
2. Open the boolean dropdown which is not mandatory
3. Select the empty("") option
4. Fill all other required fields
5. Press Submit Data.

When you select the empty value for a boolean field explicitly from the dropdown, you end up getting Ermrest API error. If you don't select anything from the dropdown it works fine.

The JSON request is
```javascript
[
  {
    "title": "some title",
    "website": "https://www.google.com",
    "category": "Hotel",
    "rating": 2.5,
    "summary": "Some summary",
    "description": "Some description",
    "no_of_rooms": 119,
    "opened_on": "2016-07-13T07:00:00.000Z",
    "luxurious": ""    // this is not valid as it should send null 
  }
]
```

You get following error for above request
```
400 Bad Request
The request is malformed. Bad JSON array input. ERROR:  invalid input syntax for type boolean: ""
```

karlcz:
See also #498 as I think there is a more general problem with nulls getting turned into some other empty/zero value

hongsudt:
@chiragsanghvi can you look into this as well?

chiragsanghvi:
Yes sure

karlcz:
I confirmed that the boolean case works as expected now, sending a null to ermrest when the blank boolean value is chosen.

PR #532, opened 2016-08-05
Recordedit issues

chiragsanghvi:
This **PR** fixes issues #498 and #503 . You can check it [here](https://dev.isrd.isi.edu/~chirag/chaise/recordedit/#3420/product:accommodation). `opened_on` is a **date** field and `luxurious` is a **boolean** field, which can be found at the end of the page.


hongsudt:
@karlcz can you please review. Thanks.

recordedit/form.controller.js
```diff
@@ -104,6 +104,29 @@
             $log.info(response);
         }
 
+
+        /*
```
karlcz: Shouldn't this treatment be nearly universal for all column types?  It is mapping a blank user-entry to NULL which is the normal way of encoding "nothing provided" in the database.

If we don't treat all columns this way, I think it should be specific exceptions that **bypass** rather than having many cases fall through cracks

karlcz:
I tested and it works but I made code comments... we should handle this null mapping for all cases, not keep revisiting it one column type at a time.

recordedit/form.controller.js
chiragsanghvi: @karlcz  For `text` and `longtext` fields should we send null in case when the user empties the input ?

karlcz: I think so.  I don't see why we would normally want to store explicit empty strings in the DB.

chiragsanghvi: @karlcz  Ok, I will make the changes, but will keep separates switch cases, so that in future if we want to transform values based on some other criteria , we can just add one more `if` statement to that `switch case`. Does that works ?

karlcz: Just make sure the default for new/unexpected column types is to treat
empty input as NULL.  We don't want this to break every time we add
something like longtext, markdown, etc. in a project.

chiragsanghvi:
Made changes suggested by @karlcz . Also fixed issue #502

Issue #510, opened 2016-07-21
Update recordset to use the common ErrorService


shinyichen:
Currently recordset has its own error alert UI, which is inconsistent with the rest of the apps.

PR #546, opened 2016-08-23
Refactor Recordset app to use Reference API #523

shinyichen:
Addresses issue #523 (refactor recordset to use Reference API) and #510 (update recordset to use common AlertsService)
This version depends on ermrestjs changes in pull request 166 (https://github.com/informatics-isi-edu/ermrestjs/pull/166) with the addition of Location and other changes to support recordset app, which is address in ermrestjs issue 141 (https://github.com/informatics-isi-edu/ermrestjs/issues/141)


common/utils.js
```diff
@@ -60,16 +60,86 @@
...
-                context.sort = hash.match(/@sort\((.*)\)/)[1];
-                hash = hash.split("@sort(")[0];
+                hash = hash.split("@sort(")[0]; // remove modifiers from uri
+            } else if (hash.indexOf("@before(") !== -1) {
+                hash = hash.split("@before(")[0]; // remove modifiers from uri
+            } else if (hash.indexOf("@after(") !== -1) {
+                hash = hash.split("@after(")[0]; // remove modifiers from uri
+            } else if (hash.indexOf("?limit=") !== -1) {
+                hash = hash.split("?limit=")[0]; // remove modifiers from uri
             }
```
howdyjessie: The repetition of `if hash.indexOf(some_modifier) !== -1 ... hash = hash.split(some_modifier)[0]` seems like a good candidate for a currying function here, especially if there will be additional modifiers.

recordset/recordset.html
```diff
@@ -2,101 +2,101 @@
...
-                            200</a>
-                        </li>
-                    </ul>
-                </span>
-            </div>
+        <!-- loading spinner -->
+        <div ng-show="pageInfo.loading" id="spinner">
+            <img src="spinner.gif" style="display: block; margin: 0 auto; width: 32px; height: 32px; opacity: 0.8;"/>
+            <div style="margin-top: 15px;">Loading...</div>
+        </div>
```
howdyjessie: @hongsudt This div appears in almost all of our Chaise apps. We should refactor this styling out onto the main stylesheet, and perhaps this div into its own little spinner directive. Spinners aren't exactly high priority but maybe this could be a small issue.

recordset/recordset.html
```diff
@@ -2,101 +2,101 @@
...
+        <!-- pagination -->
+        <div style="margin-bottom: 10px;">
+            <span class="dropdown" style="margin-left: 5px">
+                <button class="btn btn-sm btn-default dropdown-toggle" type="button" data-toggle="dropdown">Items per page
+                    <span class="caret"></span></button>
+                <ul class="dropdown-menu">
+                    <li><a href ng-click="pageLimit(10)">
+                        <span ng-show="pageInfo.pageLimit === 10" class="glyphicon glyphicon-ok"></span>
+                        <span ng-show="pageInfo.pageLimit !== 10" class="glyphicon glyphicon-invisible"></span>
+                        10</a>
```
howdyjessie: Seems like these spans can be simplified to something that avoids manipulating the DOM via [`ng-class`](https://docs.angularjs.org/api/ng/directive/ngClass). Something like: `<span class="glyphicon" ng-class="{'glphyicon-ok': pageInfo.pageLimit === 10, 'glyphicon-invisible': pageInfo.pageLimit !== 10}"></span>`

recordset/recordset.js
```diff
@@ -89,193 +82,194 @@ angular.module('recordset', ['ermrestjs', 'chaise.navbar', 'chaise.utils', 'chai
...
-                recordsetModel.rowset = rowset;
-                pageInfo.recordStart -= pageInfo.pageLimit;
-                pageInfo.recordEnd = pageInfo.recordStart + rowset.length() -1;
+            $rootScope.reference = previous; // TODO contextualize for recordset
+            $log.info("Reference:", $rootScope.reference);
 
-                pageInfo.loading = false;
+            recordsetModel.columns = $rootScope.reference.columns.map(function (column, index, array) {
+                return {"name": column.name, "displayname": column.displayname};
+            });
```
howdyjessie: Why not just `recordsetModel.columns = $rootScope.reference.columns`? Is it something like you just only need the name and displayname properties of each column?

shinyichen:
I fixed jessie's first comment. I had troubles making it work with ng-class, so I didn't put that in. And finally, the columns was written before Reference API. I changed it on the branch I am currently working on.

recordset/recordset.html
hongsudt: @jessie can you create an issue for it?

recordset/recordset.js
```diff
@@ -284,142 +278,81 @@ angular.module('recordset', ['ermrestjs', 'chaise.navbar', 'chaise.utils', 'chai
...
-                    if (error instanceof ERMrest.UnauthorizedError) {
-                        // session has expired, login
-                        Session.login($window.location.href);
-                    }
-                });
-            }, function (error) {
-                pageInfo.loading = false;
-                pageInfo.previousButtonDisabled = true;
-                pageInfo.nextButtonDisabled = true;
+    var ermrestUri = UriUtils.chaiseURItoErmrestURI($window.location);
```
jrchudy: In my branch, `chaise-to-ermrest-uri`, I made changes to this function so that it throws errors when catalog or table/schema are not specified. You probably want to wrap this in a `try/catch`.

recordset/recordset.html
howdyjessie: @hongsudt Here it is, for you to assign as needed: https://github.com/informatics-isi-edu/chaise/issues/550.

Issue #519, opened 2016-07-29
null value should be displayed appropriately


hongsudt:
By default, I think it should be an empty string.

karlcz:
See also 2015:display `show_nulls` instructions.

hongsudt:
@howdyjessie this should be handled as part of the pretty print functionality.

hongsudt:
This issue is dependent on informatics-isi-edu/ermrestjs#152. 

The app then should check the Tuple.values. If the column value is null, then that column should not be printed, otherwise, print whatever is in that value. The ermrestjs will deal with setting the right null display for the context.

howdyjessie:
This issue is currently blocked/paused due to discrepancies in how we specify contexts for annotations (i.e. `"record"` and `"detailed"` both refer to entity-level apps).

Currently the `show_nulls` property in the `2015:display` annotation uses the `detailed` context to determine a column's `formatvalue` when that column has a `null` value. In Record 2, however, the app uses `Reference.contextualize.record`, which sets the app's context to `"record"`, not `"detailed"`. Because of this, ermrestjs is unable to return the correct tuple value for null columns. 

@RFSH and I consulted with @robes about this, and we've decided the first step to resolving this discrepancy is by editing the [annotations doc](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md) to codify the contexts we'd like to use consistently. @RFSH will make a PR with changes, and we can discuss what terms to use in this PR thread.

RFSH:
I created the [pull request](https://github.com/informatics-isi-edu/ermrest/pull/95)

howdyjessie:
This has been addressed in PR #547.

PR #547, opened 2016-08-23
STILL IN PROGRESS: record2: Do not display null columns

howdyjessie:
This PR addresses #519.

For each column's pretty-printed value in Tuple.value, if a column's value is `null`, then that column should not be displayed.


test/e2e/specs/record2/helpers.js
```diff
@@ -1,29 +1,28 @@
 var chaisePage = require('../../utils/chaise.page.js');
 
 exports.testPresentation = function (tableParams) {
-	it("should have '" + tableParams.title +  "' as title", function() {
+	xit("should have '" + tableParams.title +  "' as title", function() {
```
jrchudy: Should this `xit` and all subsequent ones be renamed back to `it`?

record-two/record.app.js
```diff
@@ -31,6 +31,7 @@
             $rootScope.recordDisplayname = tuple.displayname;
             $rootScope.recordValues = tuple.values;
             $rootScope.columns = $rootScope.reference.columns;
+            console.log(tuple.values);
```
jrchudy: Do we need this? If so, use `$log.info` instead.

jrchudy:
No specific concerns besides the ones you already addressed.

Issue #537, opened 2016-08-13
common module for dealing with default catalog and table


hongsudt:
We should have a Chaise common module, defined as an Angular service, that converts the _Chaise URL_ into the appropriate _ERMrest URL_.

It should determine whether to use a default catalog/table or throw an error. Here are possible urls: 
* `chaise/app-name` (e.g. no catalog is specified) : throws an error if 
  * default catalog is not specified in the chaise config
  * default table is not specified in the chaise config
* `chaise/app-name/#<catalog#>` (e.g., no `[schema:]table` is specified: throws an error if 
  * default table is not specified in the chaise config

All apps should handle these errors the same way. However, if it is a valid url (no error thrown), the behavior will be dependent on the app e.g. for record 2, it should pop up a message and direct the app to the recordset or search page.  

Note that `ERMrest.resolve(url)` will do the parsing, validation, and error checking for the ERMrest URL. We should keep these two functions separate and let the application call one after the other. Let's assume this hypothetical function is called `c2eURL(chaiseURL)`.

```javascript
// ...maybe in the run block of an application
try {
    var ermrestURL = c2eURL(window.location);
    ERMrest.resolve(ermrestURL).then( ...
    );
}
catch (e) {
    // ...check error, could be error thrown from the c2eURL function
}
```

jrchudy:
PR #548 merged. Closing this issue.

PR #548, opened 2016-08-23
Chaise to ermrest uri

jrchudy:
The chaise to ermrest uri function now tries to handle defaults for when either the `catalogId` or `schema:table` is missing from the url. If neither are present and the defaults are not set, errors are thrown that are caught by each chaise app. Record2 catches these errors and notifies the user with a dialog popup. If the user selects `ok`, they are returned to the landing page or based on the `chaiseConfig.dataBrowser` configuration.

In record2, if there is no filter attached to the request, it redirects the user to the search page with the same `catalog/schema:table`. If those are omitted, then it tries to redirect to search with the default catalog and table used.

[Link](https://dev.isrd.isi.edu/~jchudy/chaise/record-two/) with no 'hash' included (no catalog id or any more defining information). I have catalog `#1` set as default in chaise config and `legacy:dataset` as the default table. This should redirect to search with those defaults. Here's a url that works: `https://dev.isrd.isi.edu/~jchudy/chaise/record-two/#1/legacy:dataset/id=269`.


jrchudy:
This resolves issue #537 for record2.

jrchudy:
After talking with @hongsudt, there is a popup that notifies the user that a filter is missing and that they will be redirected to search if they select ok.

common/errors.js
```diff
@@ -6,12 +6,12 @@
     // Factory for each error type
     .factory('ErrorService', ['$log', '$window', '$uibModal', 'UriUtils', 'AlertsService', 'Session', function ErrorService($log, $window, $uibModal, UriUtils, AlertsService, Session) {
 
-        function errorPopup(exception) {
-            $log.info(exception);
+        function errorPopup(message, errorCode, pageName, redirectLink) {
```
howdyjessie: Might be a good idea to have a default `redirectLink`. This would break in other recordedit app, for example, because recordedit doesn't pass in a redirect url when using `errorPopup`.

jrchudy: I made the changes to record-edit as well. I also factored it out in `recordedit.app.js`, but not here.

PR #559, opened 2016-08-25
Recordset use common table directive. #540

shinyichen:
Updated table directive to work on both recordset and record-two apps


common/templates/table.html
```diff
@@ -3,18 +3,18 @@
...
+                    <span ng-show="vm.sortby!==col.name" class="fa fa-fw fa-sort" ng-click="sortby({colname: col.name})" ng-style="{'cursor': 'pointer'}"></span>
                 </th>
             </tr>
         </thead>
 
         <!-- rows -->
-        <tr class="table-row" ng-repeat="row in tableModel.rowset track by $index">
-            <td ng-repeat="col in tableModel.columns">{{row.values[$index]}}</td>
+        <tr ng-repeat="row in vm.rowValues track by $index" ng-click="gotoRowLink({index : $index})" ng-style="{'cursor': 'pointer'}">
+            <td ng-repeat="val in row track by $index">{{val}}</td>
```
jrchudy: In removing the class attribute from the <tr>, the tests can no longer find this element. That's why they are failing.

Issue #558, opened 2016-08-24
Display columns with HTML values as HTML


howdyjessie:
Each value in a tuple returned from ermrestjs should have a `isHTML` boolean flag. If `isHTML` is true, display that value as HTML content.

This is dependent on informatics-isi-edu/ermrestjs#163.

howdyjessie:
PR #560 has been merged.

PR #560, opened 2016-08-25
record2: Display columns with HTML values as HTML

howdyjessie:
This PR addresses #558.

This PR checks `Tuple.isHTML` to determine how to display a column's value. If `isHTML` is `true`, display as HTML content. Otherwise, display escaped content.

I believe this PR is covered by existing tests, particularly the test case `"should validate the values of each column"`, which checks when a Markdown column is expected to display its HTML tags.


shinyichen:
Looks fine.

Issue #553, opened 2016-08-24
Permanent link for record2


hongsudt:
See comments in #476

hongsudt:
Please take a look at what @shinyichen did for recordset.

howdyjessie:
This issues is currently under review with PR #568.

jrchudy:
Merged PR #568, closing this issue.

PR #568, opened 2016-08-31
Record2: Add permalink

howdyjessie:
This PR addresses #553 by adding a "Permalink" link to Record 2, much like it was in the old record app.


jrchudy:
No issues from my end. I will merge your PR and then bring it into my branch before we merge the other buttons.

PR #569, opened 2016-09-02
+/- shown next to related table name

jrchudy:
By default related tables are expanded and show a +/- depending on the state. This can be tested [here](https://dev.isrd.isi.edu/~jchudy/chaise/record-two/#1/legacy:dataset/id=269).

This resolves issue #555.


robes:
commits look fine to me. but, looks like a couple unrelated tests are failing...

Issue #562, opened 2016-08-26
HTML Formatting in recordTable directive


shinyichen:
Use HTML formatting if row value is html.

PR #577, opened 2016-09-09
Display HTML in Record Table

shinyichen:
* Addresses issue #562: Display HTML in record table.
* Use compact context in recordset
* Initial e2e testing setup for Recordset.

### DIFF
--- a/record-two/record.app.js
+++ b/record-two/record.app.js
@@ -25,7 +25,7 @@
         };
     }])
 
-    .run(['ERMrest', 'UriUtils', 'ErrorService', 'pageInfo', '$log', '$rootScope', '$window', '$sce', function runApp(ERMrest, UriUtils, ErrorService, pageInfo, $log, $rootScope, $window, $sce) {
+    .run(['ERMrest', 'UriUtils', 'DataUtils', 'ErrorService', 'pageInfo', '$log', '$rootScope', '$window', '$sce', function runApp(ERMrest, UriUtils, DataUtils, ErrorService, pageInfo, $log, $rootScope, $window, $sce) {
         var context = {};
         $rootScope.pageInfo = pageInfo;
         UriUtils.setOrigin();
@@ -81,16 +81,7 @@
                                     sortOrder: null,    // asc (default) or desc
                                     rowValues: []       // array of rows values
                                 };
-                                model.rowValues = page.tuples.map(function(tuple, index, array) {
-                                    var row = [];
-                                    tuple.values.forEach(function(value, index) {
-                                        row.push({
-                                            isHTML: tuple.isHTML[index],
-                                            value: (tuple.isHTML[index]? $sce.trustAsHtml(value) : value)
-                                        });
-                                    });
-                                    return row;
-                                });
+                                model.rowValues = DataUtils.getRowValuesFromPage(page);
                                 $rootScope.tableModels[i] = model;
                             });
                         })(i);
--- a/record-two/record.app.js
+++ b/record-two/record.app.js
@@ -25,7 +25,7 @@
         };
     }])
 
-    .run(['ERMrest', 'UriUtils', 'ErrorService', 'pageInfo', '$log', '$rootScope', '$window', function runApp(ERMrest, UriUtils, ErrorService, pageInfo, $log, $rootScope, $window) {
+    .run(['ERMrest', 'UriUtils', 'ErrorService', 'pageInfo', '$log', '$rootScope', '$window', '$sce', function runApp(ERMrest, UriUtils, ErrorService, pageInfo, $log, $rootScope, $window, $sce) {
         var context = {};
         $rootScope.pageInfo = pageInfo;
         UriUtils.setOrigin();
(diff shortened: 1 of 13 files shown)


recordset/recordset.js
```diff
@@ -112,8 +113,15 @@
...
-                    return tuple.values;
+                recordsetModel.rowValues = page.tuples.map(function(tuple, index, array) {
+                    var row = [];
+                    tuple.values.forEach(function(value, index) {
+                        row.push({
+                            isHTML: tuple.isHTML[index],
+                            value: (tuple.isHTML[index]? $sce.trustAsHtml(value) : value)
+                        });
+                    });
+                    return row;
```
jrchudy: I see this block of code duplicated 3 different times in the recordset controller. It might be cleaner to factor it out into a function of the controller.

recordset/recordset.js
```diff
@@ -317,17 +339,24 @@
...
             recordsetModel.rowValues = page.tuples.map(function(tuple, index, array) {
-                return tuple.values;
+                var row = [];
+                tuple.values.forEach(function(value, index) {
+                    row.push({
+                        isHTML: tuple.isHTML[index],
+                        value: (tuple.isHTML[index]? $sce.trustAsHtml(value) : value)
+                    });
+                });
+                return row;
```
jrchudy: This is the same block of code that is duplicated in the controller, maybe there's a way that you could factor it out and re-use it in both places. A common utility might make sense because this is duplicated in record-two also.

record-two/record.app.js
```diff
@@ -81,8 +81,15 @@
...
-                                    return tuple.values;
+                                model.rowValues = page.tuples.map(function(tuple, index, array) {
+                                    var row = [];
+                                    tuple.values.forEach(function(value, index) {
+                                        row.push({
+                                            isHTML: tuple.isHTML[index],
+                                            value: (tuple.isHTML[index]? $sce.trustAsHtml(value) : value)
+                                        });
+                                    });
+                                    return row;
```
jrchudy: Comment to point out duplicated code in record-two that is also in recordset.

howdyjessie:
@shinyichen I don't have anything to add to @jrchudy's comments, except that I think the comment you wrote above `window.onhashchange` to be very helpful and lovely.

recordset/recordset.js
chiragsanghvi: Why are we combing the `value` and `isHTML` into an object ? If thats the case then `tuple.values` should return an array of objects by default that contains the `value` and `isHTML` instead of a plain string array. Or else we should be handling this in the angular HTML template itself.

shinyichen:
Updated as suggested by @jrchudy 
Good idea, @chiragsanghvi.

record-two/record.app.js
```diff
@@ -25,7 +25,7 @@
         };
     }])
 
-    .run(['ERMrest', 'UriUtils', 'ErrorService', 'pageInfo', '$log', '$rootScope', '$window', function runApp(ERMrest, UriUtils, ErrorService, pageInfo, $log, $rootScope, $window) {
+    .run(['ERMrest', 'UriUtils', 'DataUtils', 'ErrorService', 'pageInfo', '$log', '$rootScope', '$window', '$sce', function runApp(ERMrest, UriUtils, DataUtils, ErrorService, pageInfo, $log, $rootScope, $window, $sce) {
```
jrchudy: One last nitpick to remove `$sce` from being injected.

recordset/recordset.js
```diff
@@ -62,8 +63,8 @@
     }])
 
     // Register the recordset controller
-    .controller('recordsetController', ['$scope', '$rootScope', 'context', 'pageInfo', '$window', 'recordsetModel', 'UriUtils', 'Session', '$log', 'ErrorService',
-        function($scope, $rootScope, context, pageInfo, $window, recordsetModel, UriUtils, Session, $log, ErrorService) {
+    .controller('recordsetController', ['$scope', '$rootScope', 'context', 'pageInfo', '$window', 'recordsetModel', 'UriUtils', 'DataUtils', 'Session', '$log', 'ErrorService', '$sce',
+        function($scope, $rootScope, context, pageInfo, $window, recordsetModel, UriUtils, DataUtils, Session, $log, ErrorService, $sce) {
```
jrchudy: same comment as above, remove `$sce`.

recordset/recordset.js
```diff
@@ -274,8 +269,8 @@
     }])
 
     // Register work to be performed after loading all modules
-    .run(['$window', 'pageInfo', 'context', 'recordsetModel', 'ERMrest', '$rootScope', 'Session', 'UriUtils', '$log', 'ErrorService', 'AlertsService',
-        function($window, pageInfo, context, recordsetModel, ERMrest, $rootScope, Session, UriUtils, $log, ErrorService, AlertsService) {
+    .run(['$window', 'pageInfo', 'context', 'recordsetModel', 'ERMrest', '$rootScope', 'Session', 'UriUtils', 'DataUtils', '$log', 'ErrorService', 'AlertsService', '$sce',
+        function($window, pageInfo, context, recordsetModel, ERMrest, $rootScope, Session, UriUtils, DataUtils, $log, ErrorService, AlertsService, $sce) {
```
jrchudy: remove `$sce` here as well.

Issue #578, opened 2016-09-09
Use Reference.display.page_size if it is defined


hongsudt:
The 2016 table display annotation allows a default page_size to be set per context. This value is exposed through the display object. If this is defined, it should be used by the recordset. If it is not defined, recordset can use its own default. 

Please also change the default page size from 10 to 25.

jrchudy:
Merged PR #589, closing this issue.

PR #589, opened 2016-09-12
Use annotated page size for recordset tables

shinyichen:
PR for issue #578 
Use `tag:isrd.isi.edu,2016:table-display` `page_size` annotation as the default page size. If it's not specified, set default page size to 25.

### DIFF
--- a/recordset/recordset.html
+++ b/recordset/recordset.html
@@ -32,6 +32,17 @@
                 <button class="btn btn-sm btn-default dropdown-toggle" type="button" data-toggle="dropdown">Items per page
                     <span class="caret"></span></button>
                 <ul class="dropdown-menu">
+                    <li ng-if="pageInfo.pageLimit !== 10 &&
+                        pageInfo.pageLimit !== 10 &&
+                        pageInfo.pageLimit !== 25 &&
+                        pageInfo.pageLimit !== 50 &&
+                        pageInfo.pageLimit !== 75 &&
+                        pageInfo.pageLimit !== 100 &&
+                        pageInfo.pageLimit !== 200"><a href ng-click="pageLimit(pageInfo.pageLimit)">
+                        <span ng-show="true" class="glyphicon glyphicon-ok"></span>
+                        <span ng-show="false" class="glyphicon glyphicon-invisible"></span>
+                        {{ pageInfo.pageLimit }} (Custom)</a>
+                    </li>
                     <li><a href ng-click="pageLimit(10)">
                         <span ng-show="pageInfo.pageLimit === 10" class="glyphicon glyphicon-ok"></span>
                         <span ng-show="pageInfo.pageLimit !== 10" class="glyphicon glyphicon-invisible"></span>
--- a/recordset/recordset.html
+++ b/recordset/recordset.html
@@ -33,13 +33,11 @@
                     <span class="caret"></span></button>
                 <ul class="dropdown-menu">
                     <li ng-if="pageLimits.indexOf(pageInfo.pageLimit) === -1"><a id="custom-page-size" href ng-click="pageLimit(pageInfo.pageLimit)">
-                        <span ng-show="true" class="glyphicon glyphicon-ok"></span>
-                        <span ng-show="false" class="glyphicon glyphicon-invisible"></span>
+                        <span class="glyphicon glyphicon-ok"></span>
                         {{ pageInfo.pageLimit }} (Custom)</a>
                     </li>
                     <li ng-repeat="limit in pageLimits"><a href ng-click="pageLimit(limit)">
-                        <span ng-show="pageInfo.pageLimit === limit" class="glyphicon glyphicon-ok"></span>
-                        <span ng-show="pageInfo.pageLimit !== limit" class="glyphicon glyphicon-invisible"></span>
+                        <span class="glyphicon" ng-class="pageInfo.pageLimit === limit ? glyphicon-ok : glyphicon-invisible"></span>
                         {{ limit }}</a>
                     </li>
                 </ul>
--- a/recordset/recordset.html
+++ b/recordset/recordset.html
@@ -32,46 +32,15 @@
                 <button class="btn btn-sm btn-default dropdown-toggle" type="button" data-toggle="dropdown">Items per page
                     <span class="caret"></span></button>
                 <ul class="dropdown-menu">
-                    <li ng-if="pageInfo.pageLimit !== 10 &&
-                        pageInfo.pageLimit !== 10 &&
-                        pageInfo.pageLimit !== 25 &&
-                        pageInfo.pageLimit !== 50 &&
-                        pageInfo.pageLimit !== 75 &&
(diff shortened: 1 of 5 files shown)


recordset/recordset.html
```diff
@@ -32,6 +32,17 @@ <h2 id="page-title">Records for {{ vm.tableDisplayName }}</h2>
                 <button class="btn btn-sm btn-default dropdown-toggle" type="button" data-toggle="dropdown">Items per page
                     <span class="caret"></span></button>
                 <ul class="dropdown-menu">
+                    <li ng-if="pageInfo.pageLimit !== 10 &&
+                        pageInfo.pageLimit !== 10 &&
```
jrchudy: You duplicated `... !== 10`.

recordset/recordset.html
```diff
@@ -32,6 +32,17 @@ <h2 id="page-title">Records for {{ vm.tableDisplayName }}</h2>
...
+                        pageInfo.pageLimit !== 25 &&
+                        pageInfo.pageLimit !== 50 &&
+                        pageInfo.pageLimit !== 75 &&
+                        pageInfo.pageLimit !== 100 &&
+                        pageInfo.pageLimit !== 200"><a id="custom-page-size" href ng-click="pageLimit(pageInfo.pageLimit)">
```
jrchudy: Is this supposed to check if the user defined a custom page size? Maybe this check should be moved to a function even though it is only being used once.

jrchudy:
Looking at the HTML for pagination, you are repeating yourself a decent amount. Maybe we should have a default page sizes array: `[ 10, 25, 50, ...]` and then you can iterate over that array to create all of the links in the dropdown. Then you would only have to write the following:
```
<li **ng-repeat="limit in pageLimits"**><a href ng-click="pageLimit(**limit**)">
    <span ng-show="pageInfo.pageLimit === 10" class="glyphicon glyphicon-ok"></span>
    <span ng-show="pageInfo.pageLimit !== 10" class="glyphicon glyphicon-invisible"></span>
    **limit**</a>
</li>
```

We could then use that array as part of the ng-if statement for the first `li` element for a custom page size.

recordset/recordset.html
```diff
@@ -32,6 +32,17 @@ <h2 id="page-title">Records for {{ vm.tableDisplayName }}</h2>
...
+                        <span ng-show="true" class="glyphicon glyphicon-ok"></span>
+                        <span ng-show="false" class="glyphicon glyphicon-invisible"></span>
```
jrchudy: Each place that you define 2 elements that work as an if/else block, it looks like the only difference is a class that you are applying. You could shrink these elements down to one element that looks like this:
```
<span class="glyphicon" ng-class="{'glyphicon-ok': true, 'glyphicon-invisible': false></span>
```

recordset/recordset.html
```diff
@@ -32,6 +32,17 @@ <h2 id="page-title">Records for {{ vm.tableDisplayName }}</h2>
...
+                        {{ pageInfo.pageLimit }} (Custom)</a>
+                    </li>
                     <li><a href ng-click="pageLimit(10)">
                         <span ng-show="pageInfo.pageLimit === 10" class="glyphicon glyphicon-ok"></span>
                         <span ng-show="pageInfo.pageLimit !== 10" class="glyphicon glyphicon-invisible"></span>
```
jrchudy: <span class="glyphicon" ng-class="pageInfo.pageLimit === 10 ? 'glyphicon-ok' : 'glyphicon-invisible'></span>

recordset/recordset.html
```diff
@@ -32,35 +32,15 @@ <h2 id="page-title">Records for {{ vm.tableDisplayName }}</h2>
                 <button class="btn btn-sm btn-default dropdown-toggle" type="button" data-toggle="dropdown">Items per page
                     <span class="caret"></span></button>
                 <ul class="dropdown-menu">
-                    <li><a href ng-click="pageLimit(10)">
-                        <span ng-show="pageInfo.pageLimit === 10" class="glyphicon glyphicon-ok"></span>
-                        <span ng-show="pageInfo.pageLimit !== 10" class="glyphicon glyphicon-invisible"></span>
-                        10</a>
+                    <li ng-if="pageLimits.indexOf(pageInfo.pageLimit) === -1"><a id="custom-page-size" href ng-click="pageLimit(pageInfo.pageLimit)">
+                        <span ng-show="true" class="glyphicon glyphicon-ok"></span>
+                        <span ng-show="false" class="glyphicon glyphicon-invisible"></span>
```
jrchudy: I'm not sure that this second case will ever be reached. I don't think it's necessary.

recordset/recordset.html
```diff
@@ -32,35 +32,15 @@ <h2 id="page-title">Records for {{ vm.tableDisplayName }}</h2>
                 <button class="btn btn-sm btn-default dropdown-toggle" type="button" data-toggle="dropdown">Items per page
                     <span class="caret"></span></button>
                 <ul class="dropdown-menu">
-                    <li><a href ng-click="pageLimit(10)">
-                        <span ng-show="pageInfo.pageLimit === 10" class="glyphicon glyphicon-ok"></span>
-                        <span ng-show="pageInfo.pageLimit !== 10" class="glyphicon glyphicon-invisible"></span>
-                        10</a>
+                    <li ng-if="pageLimits.indexOf(pageInfo.pageLimit) === -1"><a id="custom-page-size" href ng-click="pageLimit(pageInfo.pageLimit)">
+                        <span ng-show="true" class="glyphicon glyphicon-ok"></span>
```
jrchudy: You don't need `ng-show` here because this element should only be drawn if the ng-if 1 line above passes.

recordset/recordset.html
```diff
@@ -32,35 +32,15 @@ <h2 id="page-title">Records for {{ vm.tableDisplayName }}</h2>
...
-                        <span ng-show="pageInfo.pageLimit !== 100" class="glyphicon glyphicon-invisible"></span>
-                        100</a>
-                    </li>
-                    <li><a href ng-click="pageLimit(200)">
-                        <span ng-show="pageInfo.pageLimit === 200" class="glyphicon glyphicon-ok"></span>
-                        <span ng-show="pageInfo.pageLimit !== 200" class="glyphicon glyphicon-invisible"></span>
-                        200</a>
+                    <li ng-repeat="limit in pageLimits"><a href ng-click="pageLimit(limit)">
+                        <span ng-show="pageInfo.pageLimit === limit" class="glyphicon glyphicon-ok"></span>
+                        <span ng-show="pageInfo.pageLimit !== limit" class="glyphicon glyphicon-invisible"></span>
```
jrchudy: We can reduce the amount of elements that are drawn by using the following conditional class syntax. It should half the # of elements that are drawn for the dropdown menu.

```
<span class="glyphicon" ng-class="pageInfo.pageLimit === limit ? glyphicon-ok : glyphicon-invisible"</span>
```

jrchudy:
Merging this PR as @shinyichen has addressed my comments and it passes the build now.

PR #592, opened 2016-09-13
Fixed html.

shinyichen:


jrchudy:
You could have just made this change directly on master @shinyichen. I will go ahead and merge it, no need to wait for the build to run.

Issue #584, opened 2016-09-09
Use headTitle and customCSS properties in newer apps


howdyjessie:
`chaise-config.js` has a `customCSS` property that's currently honored in the Search and Record 1 apps, but not the rest. The custom CSS file should be used in the other apps as well.

There's also a `headTitle` property that's used in Search, Record 1, Viewer, and RecordEdit, but not the others, so `headTitle` should be added in RecordSet and Record 2.

This issue should fix #516 and is somewhat related to #543 (dropdown support in navbar) because we need all apps to include the custom CSS in order to achieve a consistent look for the navbar across all apps.

howdyjessie:
Currently blocked as described in PR #599. :(

howdyjessie:
#599 has been merged!

PR #599, opened 2016-09-15
Honor customCSS and headTitle Chaise config properties

howdyjessie:
This PR addresses #584. It adds a small service to `chaise.utils` called `headInjector`, which modifies an app's `<head>` if the `customCSS` or `headTitle` properties exist. This is just so that we don't repeat the same bit of code across apps (and apps that didn't have this use those `chaiseConfig` properties, now have it).

**About failing tests (help needed! cc @chiragsanghvi)** 
I've spent a lot of time on these tests, to the point where I feel I should turn my efforts toward my next tasks, which is why I'm opening this PR.
All the tests currently pass on my local machine, but fail with different test cases each time I restart my tests on Travis. The only commonality I see is that the tests are for the Search app. Otherwise, sometimes the failing cases are related to timed-out waits, invisible elements, this weird 304 error,
screen shot 2016-09-15 at 1 47 51 pm

or this weird error about being unable to find catalog 1 after the last spec has passed, which is what the latest Travis fails with:
screen shot 2016-09-15 at 3 18 56 pm

If anybody knows that this behavior isn't just the nature of our current testing system, please let me know how I can fix these failures. :(


howdyjessie:
**Update**: This PR is ready for review now because the tests seem to pass consistently.

The failing test situation described above was resolved by adding a configuration file (`viewer.dev.json`) for Viewer app tests. Thanks to @chiragsanghvi for the help! (However, I'm still unsure how a missing configuration file for Viewer app tests could've caused failures in Search app tests.)

Issue #597, opened 2016-09-15
Use Reference.display.page_size


jrchudy:
Record2 should be using the page_size when fetching data for the related entities.

jrchudy:
PR #600 was merged. Closing this issue.

PR #600, opened 2016-09-15
Record2 pagesize

jrchudy:
Record2 now shows data in related tables based on the page size for `table-display` annotation.

This resolves issue #597.


record-two/record.app.js
```diff
@@ -70,10 +70,12 @@
 
                     for (var i = 0; i < $rootScope.relatedReferences.length; i++) {
                         $rootScope.relatedReferences[i] = $rootScope.relatedReferences[i].contextualize.compactBrief;
-                        // We want to limit the number of values shown by default
-                        // Maybe have a chaise config option
+
+                        if ($rootScope.relatedReferences[i].display.defaultPageSize) {
+                            pageInfo.pageLimit = $rootScope.relatedReferences[i].display.defaultPageSize;
+                        }
```
hongsudt: if the default is not specified, pageLimit should be set to 25, which is the recordset app default.

PR #675, opened 2016-10-18
Fix placement of tour popover when sidebar is on the left

RFSH:
You can check it out [here](https://dev.isrd.isi.edu/~ashafaei/chaise/search).


howdyjessie:
Looks good to me, merging now.

PR #680, opened 2016-10-19
Recordset Search

shinyichen:
Requires ermrestjst update: https://github.com/informatics-isi-edu/ermrestjs/pull/244

Add search to recordset and test cases.

### DIFF
--- a/recordset/recordset.html
+++ b/recordset/recordset.html
@@ -26,15 +26,16 @@
                 <!-- search box -->
                 <div class="input-group">
                     <input type="text"
+                           id="search-input"
                            ng-model="vm.search"
                            class="form-control"
                            enter-key="search(vm.search)"
                            placeholder="search {{vm.tableDisplayName}}">
                     <span class="input-group-btn">
-                        <button class="btn btn-primary" ng-click="search(vm.search)" role="button">
+                        <button id="search-submit" class="btn btn-primary" ng-click="search(vm.search)" role="button">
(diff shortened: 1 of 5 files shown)


recordset/recordset.html
```diff
@@ -21,22 +21,43 @@ <h2 id="page-title">{{ vm.tableDisplayName }}</h2>
...
-            </span>
+        <div class="row" style="margin-bottom: 10px;">
+            <div class="col-lg-4">
+                <!-- search box -->
+                <div class="input-group">
+                    <input type="text"
+                           id="search-input"
+                           ng-model="vm.search"
+                           class="form-control"
+                           enter-key="search(vm.search)"
```
howdyjessie: More a personal opinion but I would think that naming your directive `onEnter` would be more intuitive than `enterKey`. That way it can read like "on Enter run this `search` function".

RFSH:
@shinyichen @howdyjessie It's not an important issue but shouldn't we trim user input, check for this value for not being empty and then send it to `reference.search`?

howdyjessie:
@RFSH @shinyichen The string trimming is a good idea; I'd think most people would expect `       search term      `  to equal `search term`. But for the empty input value thing, if a user enters an empty string, then the app should run `$scope.clearSearch` instead of doing nothing or conducting a search for empty string.

shinyichen:
wrong branch. Close.

Issue #664, opened 2016-10-10
Markdown column type: Provide a link to markdown syntax or other online markdown editor


hongsudt:
@hongsudt commented on [Sat Oct 08 2016](https://github.com/informatics-isi-edu/ermrestjs/issues/233)

We decided that instead of extending the comments to contain markdown, a tooltip to the markdown syntax or online editor can be shown by record-edit.

howdyjessie:
PR #681 has been merged, closing now.

PR #681, opened 2016-10-19
Add help text below Markdown inputs

howdyjessie:
This PR addresses #664.

Assigning to @jrchudy and tagging @hongsudt to look at the text, an example of which can be seen [here](https://dev.gpcrconsortium.org/~jessie/chaise/recordedit/#1/experiments:purification_sop) at the "Procedure" column.


howdyjessie:
@jrchudy Yes totally agree. I've fixed that in the latest commit.

PR #686, opened 2016-10-21
Disable Sort for PseudoColumns

RFSH:
Since `reference.columns` is going to return a list of `Column` and `PseudoColumn`s, the sort functionality in recordset will break. Because we are using the column.name for sorting and in case of `PseudoColumn` the returned value is not valid.

This PR will disable the sort buttons for `PseudoColumn`s until we find a better solution for it.


shinyichen:
Just want to confirm. If a pseudo column is simple, we also want to disable?

RFSH:
@shinyichen yes, for any pseudo column.

Issue #657, opened 2016-10-07
Add 'search' control to recordset directive


robes:
Extends #656 with a 'search' control on the recordset directive. The search control should use the `*` column which indicates a pseudo column that is concatenated from all text columns of its table. We should use the `ciregexp` operator in the filter.

We decided that for the first round we _will not_ invoke the search automatically after a pause in typing. Instead, we want to start with an explicit 'search' or icon button to invoke the search.

Issue #656, opened 2016-10-07
Directive for recordset to include additional controls (paging, sorting, etc.)


robes:
The current 'recordset' directive is just for the table without any controls which is what we want to preserve for use in the nested tables display of the record app.

For the recordedit modal, we want another level of directive (around that one?) to add the basic navigation controls to it:
- page next
- preview
- sort
- limit

Issue #657 will add the 'search' control to this directive.

Issue #658 will add the 'create new entry' button or link to this directive.

While doing this, we should refactor recordset to use this directive so that add we add the search control we get that functionality in the recordset app too.

PR #691, opened 2016-10-25
Recordset Directive

shinyichen:
Addresses #656 and #657 to have a wrapper directive around current RecordTable directive. The new directive has a set of controls for the table; search, paging, page limit.


jrchudy:
If this is the most up to date code that I have been working with, this is good to be merged. I don't see anything in the code that needs attention.

Issue #710, opened 2016-11-01
Modal load delayed by a few seconds in some instances


jrchudy:
When the user clicks on a foreign key input field, there is a delay between when the user clicked on the field to when the modal appears.

This happens because the data for the modal is fetched before the modal is opened. 

Solution:
Change the order of loading so the modal is opened before `reference.read` is fired off.

jrchudy:
This issue was merged in PR #714.

PR #714, opened 2016-11-01
Modal Load order for recordedit

jrchudy:
The modal loads before it fetches the data now. There is a spinner that appears while the data is being fetched and disappears once it has loaded.


jrchudy:
This PR resolves issue #710.

PR #735, opened 2016-11-02
Fixed issue 271 in ermrestjs

chiragsanghvi:
This PR fixes issue [Search modal has area that is not clickable](https://github.com/informatics-isi-edu/ermrestjs/issues/271) in ermrestjs. 

 The problem was that the span which contained the markdowndidn't allow to bubble the click actions to the row above it.

I have enabled it and I am explicitly checking for if the target is an anchor or button tag don't bubble up or else propagate it as a row click event.

You can test he functionality for search modal [here](https://dev.gpcrconsortium.org/~chirag/chaise/recordedit/#1/experiments:sample) and recordset over [here](https://dev.gpcrconsortium.org/~chirag/chaise/recordedit/#1/experiments:sample)

### DIFF
--- a/common/table.js
+++ b/common/table.js
@@ -116,7 +116,12 @@
                     }
                 };
 
-                scope.rowClickAction = function(index) {
+                scope.rowClickAction = function(event, index) {
+                    var el = event.target || event.srcElement;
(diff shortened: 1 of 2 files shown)


jrchudy:
This is working as I would expect. Merging this issue.

Issue #763, opened 2016-11-10
Recordset needs to fetch the Session


jrchudy:
We have recently made changes to `record` and `recordedit` to make sure the session is available before `Ermrest.resolve()` is called. This needs to be done in `recordset` as well so we have consistent logic in each app.

PR #770, opened 2016-11-11
Handle no session properly

jrchudy:
This PR addresses issue #722. When no session is find in `recordedit` it continues on rather than throwing a not found error. This will now properly error out when checking the ACLs for the returned reference.

I also removed the error code numbers from the error messages that show up in the modal.

PR #773, opened 2016-11-14
Recordset session

jrchudy:
Recordset grabs the session before the reference is fetched to make sure there is a session or not. This helps in cases where recordset may be viewable as a public page.

This PR resolves issue #763. Assigning to @shinyichen and mentioning @howdyjessie.


howdyjessie:
This looks good to me. The changes correspond with changes made for PR #770.

Issue #720, opened 2016-11-01
Show All Related Entities should toggle state


robes:
From what I can tell "Show All Related Entities" is a one-way path to show all hidden (because they are empty) related entities. It might be a better UX if this link were to toggle from "Show..." to "Hide..."

Also, will our users understand what are "Related Entities"?

hongsudt:
I agreed. I discussed this with @howdyjessie today... She probably didn't get to it yet.

howdyjessie:
@robes @hongsudt 
I'm not entirely convinced that "Related Entities" is the best phrase here, but, without a suggestion of my own that I like, I've stuck with @hongsudt's proposed "Related Entities" for now. It would be nice to hear how users respond to this link.

Related Info/Information? More Info/Information? Nothing sounds good to me.

As long as there's this visual "sectioning-off" between the "main record" and all the related tables, whatever we choose for "Related Entities" should also be a heading to be placed between the main record and the related tables, so that users can more directly know what this link toggles (and also what all these tables even mean). Thoughts?

jrchudy:
One thing that we have been discussing is having the record page look consistent between the record info and the related tables. So I don't think adding a heading there will fit in with our _to be determined_ UX.

Issue #643 addresses this.

howdyjessie:
@jrchudy I hadn't realized the issue had been moved back into This Week today, so I thought we'd put #643 on the back burner for now. But yeah, if the tables can be appended seamlessly to the main record, then that would void the need for a heading.

howdyjessie:
With a completed #643 in mind, maybe a good link could be "Hide Empty Rows"/"Show All Rows"?

howdyjessie:
With PR #775 merged into master, I'll close this issue for now.

PR #775, opened 2016-11-15
Record: Toggle related entities tables

howdyjessie:
A relatively small PR to address #720, it makes the Show All Related Entities link a toggle (to show/hide empty related tables) instead of a one-time click.

Assigning to @shinyichen and mentioning @jrchudy.


jrchudy:
Looks good to me.

PR #781, opened 2016-11-16
Col wrap testfix

chiragsanghvi:
This PR fixes the test-case for column content wrapping failing locally.


howdyjessie:
This looks good to me.

Issue #746, opened 2016-11-04
recordedit support composit keys


hongsudt:
We can work on this after we have the current recordedit merge to master. 

Note: the current version only works on single column foreign keys.

PR #783, opened 2016-11-17
Recordedit composite keys

jrchudy:
This PR addresses issue #746.

I created a new schema in the test suite that includes a foreign key that is a composite key. I made this as a new schema so that it doesn't affect any of the other tests. The test is testing the ability to select a foreign key value and verify that the input field contains a value as well as checking that those values were defined after navigating to the `record` app. In the schema, I removed the visible-columns annotation and show all columns so that the pseudocolumn shows up along with the columns that make up the pseudocolumn.

There isn't a deployment where this can be tested yet as there are no schemas in dev that have composite keys.

Assigning to @howdyjessie and mentioning @shinyichen, @chiragsanghvi.


howdyjessie:
This looks good to me.

Issue #776, opened 2016-11-15
count of entities for nested tables is misleading


robes:
The count of entities maxes out at the limit for display in the nested entity table. For instance, you might have 100s of related entities but you configure the limit to 25. On the record page it will show you "Nested Table Name Here (25)".

This leaves the user not knowing:
- does this mean I am seeing all 25 related entities, or
- am I seeing 25 out of `some unknown number` of entities?

We know it is the latter but users would not. We may need to lazily get the count so the user has an indication that there are more records available.

bugacov:
And also when going into the "View all records" table doesn't seem to be an
idea of count in that view

hongsudt:
What about using 25+? To indicate that there are more and we don't have to try to go through all the rows to get an aggregate..

howdyjessie:
PR #797 has been merged into master, closing this issue now.

PR #797, opened 2016-11-23
Fix related table headings count

howdyjessie:
This PR addresses #776. It adds a "+" to related table headings if that table has more rows than the page size.

### DIFF
--- a/test/e2e/specs/record/helpers.js
+++ b/test/e2e/specs/record/helpers.js
@@ -167,7 +167,12 @@ exports.testPresentation = function (tableParams) {
                     // verify all rows are present
                     chaisePage.recordPage.getRelatedTableRows(displayName).count().then(function(rowCount) {
                         expect(rowCount).toBe(relatedTables[i].data.length);
-                        expect(headings[i]).toBe("- " + displayName + " (" + rowCount + ")");
+                        // The annotation_image table has more rows than the page_size, so its heading will have a + after the row count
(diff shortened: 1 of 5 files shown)


test/e2e/specs/record/helpers.js
```diff
@@ -167,7 +167,12 @@ exports.testPresentation = function (tableParams) {
                     // verify all rows are present
                     chaisePage.recordPage.getRelatedTableRows(displayName).count().then(function(rowCount) {
                         expect(rowCount).toBe(relatedTables[i].data.length);
-                        expect(headings[i]).toBe("- " + displayName + " (" + rowCount + ")");
+                        // The annotation_image table has more rows than the page_size, so its heading will have a + after the row count
+                        if (displayName == tableParams.related_table_name_with_page_size_annotation) {
+                            expect(headings[i]).toBe("- " + displayName + " (" + rowCount + "+)");
+                        } else {
+                            expect(headings[i]).toBe("- " + displayName + " (" + rowCount + ")");
+                        }
```
jrchudy: I'm guessing this function in helpers is being used for multiple specs? This is a case of a test that looks to be non-deterministic and should be written as 2 separate ones :/

howdyjessie: Yeah the if-else is just so this case will pass for the different specs. I have added comments the explain this.

Issue #772, opened 2016-11-14
Freeze the column names and the plus button so they are not scrollable


hongsudt:
The column names should be fixed (the same way that the columns are frozen in excel). And the add button should be moved out of the horizontally scroll-able area, so it is visible at all time.

PR #799, opened 2016-11-28
Freezed the column names and the plus button so they are not scrollable

chiragsanghvi:
This **PR** fixes issue #772.

You can test it here.

### DIFF
--- a/recordedit/form.controller.js
+++ b/recordedit/form.controller.js
@@ -472,6 +472,8 @@
             e.currentTarget.blur();
         }
 
+        /*------------------------code below is for fixing the column names when scrolling -----------*/
+
         var captionColumWidth = 130;
         var marginLeft = captionColumWidth - 5;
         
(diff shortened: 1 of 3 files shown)


recordedit/form.controller.js
```diff
@@ -471,5 +471,108 @@
         function blurElement(e) {
             e.currentTarget.blur();
         }
+
```
jrchudy: Could you put a comment here noting that below is code that is used for scrolling properly on the form create view.

Something like:
`/* ============================*/`
`// below is code for fixing the column names when scrolling

recordedit/form.controller.js
```diff
@@ -471,5 +471,108 @@
...
+                    if (keytdHeight > valuetdHeight) { 
+                        trs[i].children[1].height = keytdHeight;
+                    } else if (valuetdHeight > keytdHeight)  {   
+                        trs[i].children[0].height = valuetdHeight;
+                    } 
+                }
+            }
+        }
+
+        var DEBOUNCE_INTERVAL = 50; //play with this to get a balance of performance/responsiveness
```
jrchudy: What is this debounce for? It looks like it's for the height because it can change based on the # of forms and the content in each form. Maybe rename the variable to be more specific. 

We are introducing a debounce into the popup modal so it will auto search after x amount of time, and don't want to confuse that debounce with this one.

recordedit/form.controller.js
chiragsanghvi: Made the changes as requested

jrchudy:
That looks good. Just make sure the build passes before you merge this PR.

Issue #801, opened 2016-11-29
Table Directive More/Less Bug


shinyichen:
The more/less link should expand/hide the whole row instead of individual cell.

PR #802, opened 2016-11-29
Ellipses Fix #787 & #801

### LINKED ISSUE #787: Blank tables appear before data is loaded
For example:

https://cirm-dev.isrd.isi.edu/chaise/record/#1/Microscopy:User/Initials=JJG

The nested table for experiments will be displayed with all empty cells, and then after a second or two the data will be populated in the table in one flash.

shinyichen:
The ellipses directive was table cell based. This caused two problems 1) loading the table is slow because the ellipse code has to run for each cell in the table, which caused the problem addressed in #787, table appears blank when page initializes, and 2) clicking on the more/less link only expands/collapses content of one cell, but should expand/collapse the row #801.

This PR made ellipses directive row based, which addresses both issues.

### DIFF
--- a/common/templates/ellipses.html
+++ b/common/templates/ellipses.html
@@ -1,6 +1,6 @@
 <td ng-repeat="val in rowValues track by $index">
     <div ng-switch="val.isHTML">
-        <div ng-class="{'hideContent': hideContent == true, 'showContent': hideContent == false}" >
+        <div ng-class="{'hideContent': hideContent == true, 'showContent': hideContent == false}">
             <span ng-switch-when="true" class="markdown-container" ng-bind-html="val.value"></span>
             <span ng-switch-default ng-bind="val.value"></span>
         </div>
--- a/common/templates/ellipses.html
+++ b/common/templates/ellipses.html
@@ -2,7 +2,7 @@
     <div ng-switch="val.isHTML">
         <div ng-class="{'hideContent': hideContent == true, 'showContent': hideContent == false}" >
             <span ng-switch-when="true" class="markdown-container" ng-bind-html="val.value"></span>
-            <span ng-switch-default>{{val.value}}</span>
+            <span ng-switch-default ng-bind="val.value"></span>
         </div>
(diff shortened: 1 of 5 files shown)


common/templates/ellipses.html
```diff
@@ -1,10 +1,12 @@
-<div ng-switch="isHtml">
-    <div ng-class="{'hideContent': overflow == true, 'showContent': overflow == false}" >
-	    <span ng-switch-when="true" class="markdown-container" ng-bind-html="content"></span>
-	    <span ng-switch-default>{{content}}</span>
+<td ng-repeat="val in rowValues track by $index">
+    <div ng-switch="val.isHTML">
+        <div ng-class="{'hideContent': hideContent == true, 'showContent': hideContent == false}" >
+            <span ng-switch-when="true" class="markdown-container" ng-bind-html="val.value"></span>
+            <span ng-switch-default>{{val.value}}</span>
```
howdyjessie: @shinyichen You may want to use `ng-bind` here also to better avoid the problem described in #787, so that it's `<span ng-switch-default ng-bind="val.value"></span>`.

This [SO thread](http://stackoverflow.com/questions/16125872/angularjs-why-ng-bind-is-better-than-in-angular) has more info on `ng-bind` vs `{{ }}`, if you'd like to read more.

common/templates/ellipses.html
```diff
@@ -1,10 +1,12 @@
-<div ng-switch="isHtml">
-    <div ng-class="{'hideContent': overflow == true, 'showContent': overflow == false}" >
-	    <span ng-switch-when="true" class="markdown-container" ng-bind-html="content"></span>
-	    <span ng-switch-default>{{content}}</span>
+<td ng-repeat="val in rowValues track by $index">
+    <div ng-switch="val.isHTML">
+        <div ng-class="{'hideContent': hideContent == true, 'showContent': hideContent == false}" >
```
howdyjessie: Small nitpick, but there's an extra space between the `"` and `>` here.

Issue #766, opened 2016-11-10
Some e2e tests are non-deterministic


jrchudy:
Some e2e tests rely on a value being set, but if that value isn't set it runs a different set of expectations. These e2e tests need to be split into multiple tests in separate specs. Having them in separate specs will allow us to test functionality one way in one spec and a different way in the other spec.

hongsudt:
@howdyjessie @shinyichen @jrchudy @chiragsanghvi @RFSH can you guys review your test cases and generate issues if they need to be fixed...

jrchudy:
We addressed this issue before christmas and completed the changes when the `recordedit` PR was merged early in January coming back from the break.

PR #803, opened 2016-11-29
Fix non-deterministic tests

howdyjessie:
This partially addresses #766 by fixing the tests for navbar and the `headInjector` service.

**headInjector tests**: These deal with the specs that say `"should load custom CSS and document title in chaise-config.js"`
- Added this spec to Record app tests
- Modified the existing specs in Viewer, RecordEdit, and RecordSet apps to directly test the `customCSS` and `headTitle` properties of `chaise-config.js` instead of only testing _if_ these properties existed in `chaiseConfig`

**Navbar tests**:
- There are now 2 `chaise-config.js`s to test the navbar's functionalities by setting or not setting various `chaiseConfig` properties. The specific set of properties is listed at the top of these configs.

Assigning to @jrchudy, tagging @shinyichen if she'd like to take a look as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -421,7 +421,7 @@ distclean: clean
 # Rule to run tests
 .PHONY: test
 test:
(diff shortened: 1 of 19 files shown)


jrchudy:
Changes look good, merging this PR.

Issue #774, opened 2016-11-14
recordset search box: need a tooltip on multiple distunctive and conjuntive clauses


hongsudt:
We should provide a tooltip on how multiple and clauses can be specified.. e.g. 
`Use space to separate between conjunctive (i.e. and) criteria and | for disjunctive (i.e. or) criteria. `

PR #813, opened 2016-11-30
Search tooltips #774.

shinyichen:
Add a tooltip next to the search. #813

### DIFF
--- a/common/templates/recordset.html
+++ b/common/templates/recordset.html
@@ -12,6 +12,10 @@
                     <button id="search-submit" class="btn btn-primary" ng-click="search(vm.search)" role="button">
                         &nbsp;<span class="glyphicon glyphicon-search"></span>&nbsp;<!-- Added 2 non-breaking spaces to prevent btn's margins from collapsing and to center the icon-->
                     </button>
+                    <button type="button" tabIndex="-1" class="btn btn-primary btn-no-focus coltooltip">
+                        <span class="glyphicon glyphicon-question-sign"></span>
+                        <span class="coltooltiptext paragraph"><p>Use space to separate between conjunctive terms and quotations for exact phrases.</p><p>For example, typing <i><b>usc 1234</b></i> returns all records containing "usc" and "1234".</p><p>Typing <i><b>"usc 1234"</b></i> returns all records containing "usc 1234".</span>
+                    </button>
                     <button id="search-clear" class="btn btn-primary" ng-disabled="!vm.reference.location.searchTerm" ng-click="clearSearch()" type="button">Reset</button>
                 </span>
         </div>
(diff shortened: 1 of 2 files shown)


common/templates/recordset.html
```diff
@@ -12,6 +12,10 @@
                     <button id="search-submit" class="btn btn-primary" ng-click="search(vm.search)" role="button">
                         &nbsp;<span class="glyphicon glyphicon-search"></span>&nbsp;<!-- Added 2 non-breaking spaces to prevent btn's margins from collapsing and to center the icon-->
                     </button>
+                    <button type="button" tabIndex="-1" class="btn btn-primary btn-no-focus coltooltip">
+                        <span class="glyphicon glyphicon-question-sign"></span>
+                        <span class="coltooltiptext paragraph"><p>Use space to separate between conjunctive terms, | (no spaces) to separate disjunctive terms and quotations for exact phrases.</p><p>For example, <i><b>usc 1234</b></i> returns all records containing "usc" and "1234".</p><p><i><b>usc|1234</b></i> returns all records containing "usc" or "1234".</b></i></p><p><i><b>"usc 1234"</b></i> returns all records containing "usc 1234".</span>
```
howdyjessie: Some food for thought:
I debated for a while whether those `<b>` and `<i>` tags should be changed to `<strong>` and `<em>` tags respectively. Short answer is no, I think these are fine as is in this current case because the tooltip description here is just simply using bold and italic styles to visually separate the sample search terms from surrounding text — the search terms are not intrinsically more important than the surrounding text (like the way a paragraph heading would be).

Most people would say `<strong>` and `<em>` should be used instead because they're more "semantic", but [this article's details](http://stellify.net/html5-b-and-i-tags-are-going-to-be-useful-read-semantic-again/) on the W3C's HTML5 spec has me convinced `<b>` and `<i>` is correct here.

howdyjessie:
This looks fine to me, merging now.

common/templates/recordset.html
chiragsanghvi: I would just like to mention that we should get rid of these spans and start using Angular Bootstrap tooltips. We already have it in our codebase, we just need to start using them. The syntax is pretty simple.

You can check example [here](https://angular-ui.github.io/bootstrap/#/tooltip). For the above scenario the code would be 


```html
<button type="button" tabIndex="-1" class="btn btn-primary btn-no-focus coltooltip"
tooltip-placement="bottom" tooltip-append-to-body="true"  uib-tooltip-html='<p>Use space to separate between conjunctive terms, | (no spaces) to separate disjunctive terms and quotations for exact phrases.</p><p>For example, <i><b>usc 1234</b></i> returns all records containing "usc" and "1234".</p><p><i><b>usc|1234</b></i> returns all records containing "usc" or "1234".</b></i></p><p><i><b>"usc 1234"</b></i> returns all records containing "usc 1234".'>
       <span class="glyphicon glyphicon-question-sign"></span>
</button>
```

Issue #785, opened 2016-11-17
"+" button in the wrong place on edit pane when adding columns causes a scroll


carlkesselman:
If you add a large number of columns to the create widget the columns will properly scroll of the side, however, the + button ends up in a column in the middle, and not over the far right column.  

Also, the columns become too narrow, causing the text to overflow the box.

hongsudt:
We are aware of the problem and this issue is currently being worked on.

Issue #753, opened 2016-11-08
allow users to specify # of rows to be duplicated


hongsudt:
Instead of duplicate one row at a time, we allow users to specify # of rows to be duplicated. 

We used to have that but took this feature out.

hongsudt:
There should be a hard maximum limit for the # of rows to be added.. Maybe 200? @jrchudy can you try and see how responsive is if you add 100 rows at once..

PR #815, opened 2016-12-01
Number of records

jrchudy:
This PR addresses issues #753 and #785. Both of these buttons also have a tooltip added to them. The tests that I included test the implementation of:
 - clicking the button to open the input and clicking again to close it
 - making sure the error case occurs properly and alert shows up when a number greater than the max is input
 - it also verifies that multiple forms are indeed added when the user submits an appropriate value 

The tests do not test filling out the forms and submitting them to the database. I spent some time trying to use the `testPresentationAndBasicValidation` function in `../specs/recordedit/helpers.js` to fill in my forms and submit them but that function is too specific to the case of adding 1 row at a time and verifying that new one can be filled out. I would have to heavily modify that function or write a whole new function for filling out my forms, to test this functionality that is already tested in a slightly different context.

This functionality can be tested [here](https://dev.isrd.isi.edu/~jchudy/chaise/recordedit/#1/legacy:dataset). 

Assigning to @howdyjessie and mentioning @shinyichen.


howdyjessie:
- The input field is taller than the checkmark button, anyway to fix that and make them aligned?
- Should the pencil icon be that? We use it to mean editing things, which I kinda get why you'd use it in this context (the user is editing how many rows to add), but I don't find it immediately clear what I'm editing. How about a &#65049; (`&#65049;`) or &#8943; (`&#8943;`) like a context menu?  Or even `glyphicon-triangle-bottom` to indicate there's a dropdown there?

howdyjessie:
Oops I just realized you said this morning you're retracting this PR! Sorry for the premature comments.

jrchudy:
@howdyjessie, this issue is ready for review. I have since fixed the position of the input control so it doesn't get placed on top of a trash can icon and made the input group look better.

howdyjessie:
This looks good, I've merged it.

Issue #784, opened 2016-11-17
create "add" button into the search directive


hongsudt:
When the add button is clicked, a new tab is opened with recordedit. After the addition is complete, the newly added value need to be populated in the search directive. 

@robes @chiragsanghvi We discussed that a search directive can monitor some event in the submission. Please provide some detail in this issue.

shinyichen:
Update this to use cookie to monitor event.

hongsudt:
@shinyichen There should be a tooltip associated with the add button. After the add button is done, we should delete the add button from the recordedit page.

shinyichen:
I think your second suggestion about removing the button is kind of strange. Even though it's not likely user will add another record, but it's strange to prevent user from adding another record.

hongsudt:
Sorry.. That's not what I meant. I meant to remove the add button that we currently have on the recordedit page itself, so they won't be adding a new item without going to the search directive.

PR #824, opened 2016-12-02
Create "add" button into the recordset directive

shinyichen:
PR for #784 'Add' button in the recordset directive.

When the add button is clicked, a new tab is opened with recordedit.
After the addition is complete, window is switched back to search directive, read() is called to refresh the table data.


common/table.js
```diff
@@ -194,6 +198,20 @@
...
+                    // open a new tab, once focus is back to this page, reload
+                    var newRef = scope.vm.reference.contextualize.entryCreate;
+                    var appLink = newRef.appLink;
+
+                    // open url in a new tab
+                    var window = $window.open(appLink, '_blank');
+                    if (window) {
+                        window.focus();
+                    } else {
+                        $window.location.href = appLink;
```
jrchudy: What's the purpose of the else case here? Is it so if the page fails to open a new tab, it will redirect the current one?

shinyichen: I saw this logic being added to the Record app when opening app_link in a new tab, which I believed is doing what you said, so I added it here as well.

Issue #819, opened 2016-12-01
More/less ellipses non-deterministic on page load


karlcz:
See https://synapse-dev.isrd.isi.edu/chaise/recordset/#1/Zebrafish:Behavior@sort(Image%20Date::desc::)

1. On a fresh page load w/ empty browser cache, I see no more/less ellipses on the rows with thumbnail images.
2. On a reload, I see thumbnails truncated with a "...more" control below them.

One time, I even saw a mixed rendering where some rows had a truncation control and some rows were showing full thumbnails without any "...less" control (like in case 1).

So, I think there is an async initialization bug in this feature that depends on resource load times.  I see this on firefox.

jrchudy:
I cannot reproduce this issue in Chrome. 

What I think could be happening is that the renderer for firefox is rendering the angular template for the ellipsis directive before the images finish loading. So they load and show at their full size. 

But when you go to this page with the images cached everything appears properly, which leads me to believe that the images load before the ellipsis directive when the data is cached.

chiragsanghvi:
@karlcz This issue has been fixed and is in Master.

Issue #820, opened 2016-12-01
More/less ellipses needs fine-tuning


karlcz:
On https://synapse-dev.isrd.isi.edu/chaise/recordset/#1/Zebrafish:Behavior@sort(Image%20Date::desc::)

The "...less" control for an expanded thumbnail adds significantly more vertical height to the row. It has even more padding than the "...more" control. At minimum, it should be tightened up like the "...more" control.  Ideally, both controls should be even more compact so they don't add to the very problem they are trying to solve...

Also, I think the thresholds need to be a little smarter so that if a truncated value + "...more" control is nearly as tall as the original content (I'd say within 80-90%), the control is omitted and the full content shown instead. Having something truncated and replaced with this control when it basically fits is very frustrating to a user...

hongsudt:
Please review.

karlcz:
Master still has most of the bad behaviors I described originally.

1. The "less" and "more" divs are identical height but there is an extra vertical padding before the "less" which makes an expanded cell even more spread out.
2. The threshold doesn't seem to take into account the added height of the control to realize that showing the entire content would use less vertical space than a slightly truncated content plus "...more". Test with content that is just a few pixels taller than the threshold.

hongsudt:
@chiragsanghvi this issue still exists. Please fix.

chiragsanghvi:
@karlcz The threshold approach is not required. You can simply set the height to the threshold height you're looking for. 

Currently the more button height is not considered when we consider the height. The calculation considers the actual element height.

You see the more button, because the height is actually greater than the allowed height.

I tested the behavior on the above link, and the image itself has a height of 170px and the threshold is 160px. Thats the reason you see a more link. If you want it to be rendered completely then set the max-height as 170.

karlcz:
I think you're missing my point.

The version of the row with the "...more" button should *never* be taller than the original content.

So whatever the threshold, once activated this mode should truncate the content to be the threhold height *minus* the extra space needed for the "...more" control.

chiragsanghvi:
@karlcz Oh,  I got it wrong. I have made the changes to consider the more button too and added the support to disable more-less behavior as requested by @mikedarcy and @robes .

You just need to set the property `maxRecordsetRowHeight` property to the threshold height in the chaiseConfig. If this property is not set then more-less behavior will not be enabled.

[...]

PR #826, opened 2016-12-02
More less changes

chiragsanghvi:
This PR fixes issues #819 and #820. 

https://synapse-dev.isrd.isi.edu/~chirag/chaise/recordset/#1/Zebrafish:Behavior@sort(Image%20Date::desc::)

### DIFF
--- a/common/ellipses.js
+++ b/common/ellipses.js
@@ -16,32 +16,49 @@
                     scope.overflow = []; // for each cell in the row
 
                     scope.hideContent = false;
-                    scope.linkText = "more";
+                    scope.linkText = "show more";
+                    scope.maxHeightStyle = { };
 
                     // 1em = 14px
                     // 7.25em = 101.5px
-                    scope.maxHeight = "7.25em";
-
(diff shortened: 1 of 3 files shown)


karlcz:
Does this include a change to the threshold for truncation?  I am trying to test the more-less branch but not seeing the behavior anymore...

chiragsanghvi:
Yes. Hongsuda asked me to increase the default size to 160px.  I read it from chaise-config. So you just need to set the parameter `maxRecordsetRowHeight` in chaise-config to change it.

karlcz:
This is not working right on synapse-dev....  when I set the max height to 150 and then see the "... more" link, clicking it makes the row shorter rather than taller and the "... less" control is missing.

chiragsanghvi:
@karlcz Sorry about that. I was referring the wrong variable to check the overflow. Jenifer made some changes to the ellipsis plugin and I forgot to consider them in my branch. Have fixed it and updated the userdir website above

Issue #817, opened 2016-12-01
update the appropriate related table entries after the new entries are added


hongsudt:
After an add operation originating from the record page is done, it should save the created status in the cookie.

When the record app tab is focused, it should check the cookie and update the appropriate related tables with new entries.

PR #835, opened 2016-12-06
Record app related table refresh

shinyichen:
PR for #817 

When an add related table entry operation originating from the record page, a cookie is created with a unique id. When the new recordedit tab has finished creating new entries, it modifies the cookie's value. When the record tab regains focus, it checks its cookie and update the appropriate related tables with new entries.

### DIFF
--- a/record/record.controller.js
+++ b/record/record.controller.js
@@ -3,7 +3,7 @@
 
     angular.module('chaise.record')
 
-    .controller('RecordController', ['AlertsService', '$cookies', '$log', 'UriUtils', '$rootScope', '$window', function RecordController(AlertsService, $cookies, $log, UriUtils, $rootScope, $window) {
+    .controller('RecordController', ['AlertsService', '$cookies', '$log', 'UriUtils', 'DataUtils', '$rootScope', '$window', function RecordController(AlertsService, $cookies, $log, UriUtils, DataUtils, $rootScope, $window) {
         var vm = this;
 
         vm.alerts = AlertsService.alerts;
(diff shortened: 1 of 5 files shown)


shinyichen:
Assigned to @howdyjessie since I utilized the cookies code she implemented. Adding @jrchudy to take a look as well.

shinyichen:
How about this? Rename the `prefill` query string to `referrer`, cookie still has same name/data, with an extra field called `hasPrefill`.  This is actually how I modified for my next issue because for my table refresh task, I really don't need anything other than a unique identifier from the referrer. So I utilized the existing prefill ID without making the URL and parsing more complicated. 

What do you guys think?

howdyjessie:
@shinyichen That sounds fine in that I wouldn't want there to be duplicated code just to parse out another parameter. I also imagine `referrer` is generic enough to be used when we need something similar for other inter-app communications (like the Cancel button on RecordEdit?). (i.e. On initialization and on focus, each Chaise app checks for a cookie in the `referrer` parameter to see what special "instructions" were received from the referring app).

Does anyone share my apprehension about the second step in this PR description, in which the RecordEdit modifies the existing cookie? Should RecordEdit create a _new_ cookie with the info needed for Record to know which related table to refresh? Modifying the existing cookie sounds like we're starting on a path that couples Record and RecordEdit tightly together.

robes:
I have not objections. I agree with the change to `referrer`.

shinyichen:
Another PR will reflect the updated approach

Issue #823, opened 2016-12-02
add a link to toggle between the markdown table display and tabular display


hongsudt:
The tabular display will include a column that allows use to take action on each entry. But this option is now available for markdown table display. So we need a way to toggle between the two presentation format to allow editing.

howdyjessie:
PR #844 has been merged into master, so I'm closing this now.

PR #844, opened 2016-12-08
Toggle Markdown related tables to between markdown and tabular views

howdyjessie:
This PR addresses #823.

It adds a "Table Display" link to each Markdown related table, which allows the user to toggle that table between Markdown and tabular display.

I replaced the `ng-switch` logic in favor of `ng-show` because I didn't want Angular to remove and re-insert elements in the DOM every time the user toggled the display, which would cause a delay for the user.

Requesting a review from to @jrchudy and tagging @shinyichen as well. Also assigning this PR to @jrchudy because ZenHub doesn't allow filtering by Reviewer yet. :(

### DIFF
--- a/record/record.controller.js
+++ b/record/record.controller.js
@@ -88,6 +88,14 @@
             }
         };
 
+        vm.toggleRelatedTableDisplayType = function(i) {
+            console.log($rootScope.tableModels[i]);
+            if ($rootScope.tableModels[i].displayType == 'markdown') {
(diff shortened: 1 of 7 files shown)


record/record.controller.js
```diff
@@ -88,6 +88,14 @@
             }
         };
 
+        vm.toggleRelatedTableDisplayType = function(i) {
+            console.log($rootScope.tableModels[i]);
```
jrchudy: Leftover `console.log()` statement.

Issue #843, opened 2016-12-08
a wrapper function for url parameter lookup


hongsudt:
We don't need to completely rewrite this yet, but should have a function that returns all url params or perform a lookup of a certain parameter name

PR #847, opened 2016-12-09
Refresh Table Using Cookies

shinyichen:
#784 When 'Add' button on Recordset is clicked, append a unique id in the RecordEdit URL for this request. When RecordEdit completes, a cookie name by that id is added. When recordset gets focus, it checks if any of its requests has been completed by the presence of the cookies. Do a read on the table if necessary.
 
#817 When 'Add' button on the related table in Record is clicked, handle it the same way as above. Do read on the related table that has been updated.

#843 Update the URL parser to handle multiple query params, accessible using the `queryParams` object in the result context.


shinyichen:
@howdyjessie I disabled the two cookie related tests for now. Can you go ahead and review. When you test locally, you have to enable these test cases. 
1. record/helpers.js --> last test case
2. recordset/add/01-recordset.add.spec.js  --> last test case

Manually testing. 
1) Go to recordset, click the add button (one or more times) to open recordedit tab(s), create new record(s). Go back to recordset and see them in the table (depends on sorting).
2) Go to a record, click add on related table(s) to open recordedit tab(s). Same as above.

shinyichen:
Add @jrchudy to take a look as well.

howdyjessie:
@shinyichen This PR passed local tests; it looks like @chiragsanghvi has pushed some commits and now there are conflicts with master. I suggest resolving the conflicts asap and merging into master and setting aside the cookie debugging on a different branch.

Issue #790, opened 2016-11-18
Add copy functionality


hongsudt:
There should be a "Copy" button on the record page.. Upon clicking the copy button, a new recordedit page should be populate with those values in the columns. 

Note that: all the invisible columns need be reset to null, so that the proper defaults can be set in the database.

PR #850, opened 2016-12-09
Copy btn

jrchudy:
This PR addresses issue #790.

The changes include adding a `copy` button to the Record page that depends on the user being able to edit the record. This adds a url param `?copy=` and sets it to true. This url is constructed similar to our edit mode urls but uses this `copy` param to differentiate it and set `recordedit` in create mode.

You can test the chages [here](https://dev.isrd.isi.edu/~jchudy/chaise/record/#1/legacy:dataset/id=5). Assigning to myself and adding @howdyjessie as a reviewer.


record/record.controller.js
```diff
@@ -33,6 +33,14 @@
             }
         };
 
+        vm.copyRecord = function() {
+            var referenceFilter = $rootScope.reference.location.filter;
```
howdyjessie: Is `referenceFilter` being used anywhere else?/What is `referenceFilter` being used for?

recordedit/form.controller.js
```diff
@@ -7,7 +7,7 @@
         var vm = this;
         var context = $rootScope.context;
         vm.recordEditModel = recordEditModel;
-        vm.editMode = context.filter || false;
+        vm.editMode = (context.filter && !context.copy) || false;
```
howdyjessie: The `|| false` is probably unnecessary here. `context.filter && !context.copy` will evaluate to `false` if it's not `true`.

howdyjessie: Or you may elect to keep the `|| false` if you feel it's more readable. I'm ambivalent.

record/record.controller.js
jrchudy: I accidentally forgot to remove this.

Makefile
```diff
@@ -19,20 +19,27 @@ MODULES=node_modules
 BIN=$(MODULES)/.bin
 
 # Protractor scripts
+# Legacy apps tests
```
howdyjessie: These comments though. 💯

Issue #837, opened 2016-12-07
Remove Add button from recordedit


hongsudt:
After the pull request for adding a create button in the search directive is merged, please remove the add button from the recordset.

Issue #786, opened 2016-11-17
Search box behavior on edit popup


carlkesselman:
I would suggest two changes to the current behavior:

1. When the popup first appears, we should have the search box selected.  Right now if I have a bit list the first thing I'm going to do is narrow it down, but this requires me to select the search box which means I have to move the mouse and click.  This is awkward.  If I was going to scroll, I will have to move the mouse anyway, so assuming that search will be the first thing I do seems like a good assumption

~2. I think having to click the return button to get the search to take effect is not behavior that people would expect.  I would like to propose that we have dynamic update of the search result.~ We will address the second bullet 

If we could get type ahead in the search box, that would also be very nice As an added bonus, if we had type ahead, we might also provide that in the edit pane, eliminating the need to go to the selector popup in many cases

Issue #768, opened 2016-11-11
"Add Record" links to wrong table on association table linkages


robes:
When you have a model that has associated entities, the "Add Record" will take you to the recordedit page to add another of the target entities rather than to add a new association.

Based on this figure:
`[Table 1] <-- [Association Table] --> [Table 2]`

If you were on the record page for an entity from `Table 1` and you see the nested tables section for `Table 2` entities, when you click on the "Add Record" link it takes you to the recordedit page to create a new entity for `Table 2`. This is a bug. What it should do is take you to the recordedit page to create a new entity for `Association Table` instead.

robes:
@hongsudt if we could get this fixed we could have a workaround to let us move forward on the CIRM update.

hongsudt:
@RFSH Can you look into this?

hongsudt:
@robes Do you have the latest code?  Chaise/ermrestjs seems to handle the association table correctly in the gpcr use case.

robes:
We are using the chaise master branch and it should be up to date.

I showed this to @jrchudy and he confirmed it was a bug. I can show to others.

For example, go to this record:

https://cirm-dev.isrd.isi.edu/chaise/record/#1/Microscopy:Specimen/ID=20160920-mKDCrlf1IRES21d-GA-0

Under the "Specimen Gene" section click "Add a Record". It takes you to record edit for the `gene` entity. It should go to the `specimen_gene` table to add an entity to the association table.

RFSH:
The problem is that in ermrestjs, the context that is returning the association table is edit ([`ref.contextualize.entryEdit`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L1213)) while in chaise we're using create context ([`ref.contextualize.entryCreate`](https://github.com/informatics-isi-edu/chaise/blob/master/record/record.controller.js#L98)). 

I expanded the logic in ermerstjs to include all the entry contexts (entry, entry/create, and entry/edit).

hongsudt:
@howdyjessie please review this issue and make sure that the contexts are in sync.  I think what we have is now correct.. Ermrestjs should handle all editing contexts and that link in chaise should be create.

howdyjessie:
[...]

PR #851, opened 2016-12-09
Auto focus search box in recordset #768 and other minor UI adjustments

shinyichen:
* Auto focus search box #786
* Move 'items per page' next to search box
* Add tooltip for the 'Add' button
* Remove add button from recordedit #837


howdyjessie:
This all looks good to me. I do have 2 nitpicks:
- In the search box on RecordSet and the modal version, I'd think the "s" in "search [table name]" should be capitalized.
- The Items per page box is slightly shorter than the other buttons we use; is it possible to make this box the same height?

PR #866, opened 2016-12-14
Rename config showDeleteButton to deleteRecord

shinyichen:


hongsudt:
@shinyichen There is a conflict with this pull request..

PR #872, opened 2016-12-15
Re resultset

jrchudy:
This PR includes changes to the `recordedit` app so that when multiple records are created it shows a resultset table instead of redirecting to recordset.

Included in this PR is a change to the table directive to include an `enableSort` option so that we can turn sorting on and off in the directive when we need to. Sorting is expected to be unavailable for this resultset table.

This code can be tested [here](https://dev.isrd.isi.edu/~jchudy/chaise/recordedit/#1/legacy:dataset).


jrchudy:
@howdyjessie, I updated the code based on your suggestions. I verified that the build passed on travis. Could you take one last look at this before I merge it into master?

recordedit/form.controller.js
```diff
@@ -231,6 +231,9 @@
                     if (model.rows.length == 1) {
                         vm.redirectAfterSubmission(page);
                     } else {
+                        AlertsService.addAlert({type: 'success', message: 'Your data has been submitted. Showing you the resultset...'});
```
howdyjessie: "Resultset" is not an actual word; two words should be used in the UI.

Issue #852, opened 2016-12-12
e2e tests need a configurable time


jrchudy:
We need to have a configurable wait timer so that all the tests can be changed based on this option instead of having to change tests individually.

howdyjessie:
The PR was merged into master, closing now.

Issue #853, opened 2016-12-12
e2e tests should use browser.wait instead of sleep


jrchudy:
In our e2e tests we are using `browser.sleep` in some cases and `browser.wait` in others. `Sleep` will stop execution for the amount of time specified whereas `wait` will stop execution until the condition is satisfied. `Wait` is more ideal because it can break out of the timer if the condition is satisfied.

hongsudt:
The test owners should change their test to use wait instead of sleep.

hongsudt:
From our standup meeting today, it seems like we are all done with this task.. So I am closing the issue.

PR #873, opened 2016-12-16
Create configurable default timeout for Protractor tests

howdyjessie:
This PR addresses #852 by introducing a default timeout (currently 30000ms) at `browser.params.defaultTimeout`, which allows us to change the timeout for url changes and async calls in one place instead of all over the tests. To change the timeout figure, see the `params.defaultTimeout` property of the `config` object in `protractor.configuration.js`.

Most `browser.sleeps` that were used for url changes or async calls have been modified to use the new default. Some of the shorter browser.sleeps were preserved because they were put in place to pace WebDriver while the browser renders new elements, so a full 30 sec sleep is unnecessary (and will sometimes cause those tests to time out due to inaction).


test/e2e/specs/recordedit/helpers.js
```diff
@@ -360,13 +360,13 @@ exports.testPresentationAndBasicValidation = function(tableParams) {
...
 	                                // wait for the modal to open
-	                                browser.wait(EC.visibilityOf(modalTitle), 5000);
+	                                browser.wait(EC.visibilityOf(modalTitle), browser.params.defaultTimeout);
 
 	                                return modalTitle.getText();
 	                            }).then(function(text) {
 	                                // make sure modal opened
 	                                expect(text.indexOf("Choose")).toBeGreaterThan(-1);
-
+                                    // TODO: What is this 5-sec sleep for? To make sure all rows are loaded?
```
jrchudy: I believe that is correct. This is so to make sure the rows all load first. This should certainly be turned into a `browser.wait()`.

howdyjessie: All righty, I'll leave this for when work is done on #853.

PR #881, opened 2016-12-18
More threshold

chiragsanghvi:
This pr fixes issue #820 . It also adds an explicit handling for tuning off more-less feature.

To turn it off set `maxRecordsetRowHeight: false` int chaiseConfig. The default maxHeight is 160. If you want to change it then set `maxRecordsetRowHeight ` to respective height in chaiseConfig.

```js
maxRecordsetHeight: 200
```

https://synapse-dev.isrd.isi.edu/~chirag/chaise/recordset/#1/Zebrafish:Behavior@sort(Image%20Date::desc::,ID)?limit=25

### DIFF
--- a/chaise-config-sample.js
+++ b/chaise-config-sample.js
@@ -24,6 +24,7 @@ var chaiseConfig = {
     recordUiGridExportPDFEnabled: true,
     editRecord: true,
     deleteRecord: true,
+    maxRecordsetRowHeight: 160,
     tour: {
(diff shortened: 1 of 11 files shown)


jrchudy:
The changes look good. In the PR comment, you mentioned it as `maxRecordsetHeight` but the changes in the code are `maxRecordsetRowHeight`. Besides that small discrepancy, this is good.

Issue #848, opened 2016-12-09
RecordEdit "x" on the popup select input box does not work


shinyichen:
It's possible the clear button broke after the popup selected was changed from input field to div. Tooltip doesn't show and clicking on it doesn't clear the input, but opens the modal.

chiragsanghvi:
Merged in Master

PR #884, opened 2016-12-19
Foreign Key X button not clickable

chiragsanghvi:
This PR fixes issue #848 where you couldn't click on x button when you have selected a foreign key from modal on recordedit page.

The issue was with the class `form-control-feedback` which added a `pointer-events:none` css style. This property disables the element from firing any input events.

https://dev.isrd.isi.edu/~chirag/chaise/recordedit/#1/legacy:dataset


jrchudy:
This change is trivial and works as expected. This is ready to merge whenever.

PR #887, opened 2016-12-19
Split make testall into multiple jobs to avoid travis timeout

chiragsanghvi:
This PR fixes issue #886. We had to split `make testall` into separate smaller jobs because Travis has a time limit of 50 minutes per job.


hongsudt:
Please review Chirag's PR.

jrchudy:
I verified that these `test<app>` targets include all 19 of our current protractor test configurations. This is all set to merge.

Issue #791, opened 2016-11-18
Automatically update the search directive if the typing in the search box is paused by 2 s.


hongsudt:
The app should make sure that it won't keep submitting the same request. As the user continue to type, new requests should be submitted and old results should be ignored.

karlcz:
It's not a really big deal to repeat an identical query, since our
cache management should detect that as a repeat and quickly let the
browser reuse its cached result, as long as we don't repeat until the
previous response has been received.

I would suggest logic like the following, distinguishing a
"background" search we perform based on idle time and a "foreground"
search we perform due to user hitting enter or a search button:

1. Keep flags for whether background or foreground searches are in
   progress.

2. Only start a background search if we have no outstanding searches.

3. Only start a foreground search if we have no outstanding foreground
   searches.

4. If we have already started a foreground search, discard any
   background search result that arrives.

5. Ideally, lock the search input as busy while a foreground search is
   pending.

Since we cannot cancel a slow search, we shouldn't make the user think
they are still able to initiate an action. Nor should we actually
submit concurrent foreground searches since this abuses the server.


karl

PR #894, opened 2016-12-20
Search popup

chiragsanghvi:
This PR implements the search feature as described in issue #791 

The search directive has 2 variables viz. backgroundSearch & foregroundSearch to track which search is going on currently.

Changes can be seen [here](https://dev.isrd.isi.edu/~chirag/chaise/recordedit/#1/legacy:dataset)

The recordset app needs to change a bit, because it updates the whole page while you're typing text in search input box. I have added a switch to enable/disable the auto search.

The delay is 200ms for now.

### DIFF
--- a/common/ellipses.js
+++ b/common/ellipses.js
@@ -44,14 +44,13 @@
                             }
                         };
 
-                        var timerCount = 0, containsOverflow = false, oldHeights = [];
+                        var timerCount = 0, containsOverflow = false;
 
                         var resizeRow = function() {
                             if (containsOverflow == false && timerCount ++ < 500) {
                                 
                                 for (var i = 0; i < element[0].children.length; i++) {
                                     var height = element[0].children[i].children[0].clientHeight;
-                                    if (height < oldHeights[i]) continue;
                                     if (height > maxHeight) {
                                         scope.overflow[i] = true;
                                         scope.hideContent = true;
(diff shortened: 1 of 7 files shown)


chiragsanghvi:
I updated the branch width changes as suggested by @hongsudt to have same behavior for search in recordset app too.

The recordset can checked [here](https://dev.isrd.isi.edu/~chirag/chaise/recordset/#1/legacy:dataset@sort(id)?limit=15)

shinyichen:
Hongsuda wanted me to take a look at this PR even though it's already merged. Are the followings implemented?
1. If we have already started a foreground search, discard any background search result that arrives. ---- It seems every read result will update the table.
2. Ideally, lock the search input as busy while a foreground search is
pending. ---- although it says ideally, I wonder if you just missed it since it's a small change.

chiragsanghvi:
1. I have added an explicit check to discard background search whenever a foreground search is going on. That kind of takes care of the situation.

2. Whenever you fire a foreground search, the search box is behind the spinner, and hence it gets disabled automatically. This is for the popup. I am not sure about the spinner in recordset. I can change it if the spinner doesn't covers the whole page.

shinyichen:
This is the sequence I meant. 
1. background search started, search() function is called.
2. user clicks enter (bg flag is disabled, but task 1 is already reading)
3. task 1's read() comes back with result. (**The spec says this result should be discarded.)
4. Foreground search comes back with result.

Issue #632, opened 2016-09-26
Fields should be pre-filled with Default values


jrchudy:
Input fields should have their default values displayed when a page loads if those values are based on a value instead of a function.

PR #896, opened 2016-12-21
Recordedit defaults

jrchudy:
The recordedit page will now prefill basic input fields with default values as defined by the schema ass long as their are values rather than a function. This PR addresses issue #632.

There are currently no deployments that have default values defined on columns that are part of the visible column list. @hongsudt mentioned there is on dev.gpcr, but the ACLs for that deployment have changed again.


jrchudy:
I was able to confirm that the e2e tests pass from beginning to end this morning.

recordedit/recordEdit.app.js
```diff
@@ -183,6 +183,15 @@
                 } else {
                     if ($rootScope.reference.canCreate) {
                         $rootScope.displayname = $rootScope.reference.displayname;
+
+                        // populate defaults
+                        angular.forEach($rootScope.reference.columns, function(column) {
+                            // typeof column.default !== "function"
+                            if (column.default) {
```
howdyjessie: Curious about why you're not checking via `typeof` here, since defaults are only filled if they're not a function and is a value? (And if you don't need this commented line, it should be removed.)

howdyjessie: Thinking out loud: In the scenario when the the URL has a `prefill` cookie but no filter, then the app will go to `if (context.prefill)` to fill in the foreign key link's view model, and then skip `if (context.filter)` to go to its else branch and apply this `angular.forEach` loop.

Is there a chance of this loop overwriting the prefill's values? (And if so, is that a concern anyway because does it make sense for foreign key columns to have defaults?)

howdyjessie: Hmm I see a comment in your tests about adding "tests for default values for foreign keys when implemented" so perhaps my comment above is premature.

howdyjessie:
E2E tests passed.

test/e2e/specs/recordedit/data-independent/add-defaults/07-recordedit.defaults.spec.js
```diff
@@ -0,0 +1,49 @@
...
+            });
+
+            it("and redirect to a record page with the default values.", function() {
+                // After submitting 1 record in RecordEdit, the expected record
+                // page url will have a id of 1 because it'll always be the first
+                // row of this table in the new catalog created by the defaults tests.
+                var redirectUrl = browser.params.url.replace('/recordedit/', '/record/');
+                redirectUrl += ":" + testParams.table_name + '/id=1';
+
+                chaisePage.waitForUrl(redirectUrl).then(function() {
```
jrchudy: I thought the function in chaise page was defaulting to the default timeout if we didn't set one. I guess it's defaulting to `5000` instead of our set default.

Maybe we should change that..

recordedit/recordEdit.app.js
```diff
@@ -186,8 +186,7 @@
 
                         // populate defaults
                         angular.forEach($rootScope.reference.columns, function(column) {
-                            // typeof column.default !== "function"
-                            if (column.default) {
+                            if (column.default && typeof column.default !== "function" && !recordEditModel.rows[0][column.name]) {
```
howdyjessie: A comment would be useful here for why you're including a check for `recordEditModel.rows` here, and maybe also explain the check for function as well.

Issue #816, opened 2016-12-01
Convert current tooltips to Bootstrap tooltips


jrchudy:
Each tooltip needs to be replaced with the Bootstrap implementation for tooltips. This is a more robust implementation and handles the oddities we are seeing with our current ones.

This requires adding `tooltip-placement`, `tooltip-append-to-body`, and `uib-tooltip` to the elements that should have tooltips and removing the existing tooltip elements.

howdyjessie:
PR #897 has been merged, so closing this issue now.

PR #897, opened 2016-12-21
Change tooltips to use UIB tooltips

howdyjessie:
This PR addresses #816. It converts our existing CSS-based tooltips to the more flexible [tooltip tool from Angular UI](http://angular-ui.github.io/bootstrap/versioned-docs/1.3.2/#/tooltip).

 - The `tooltip` attribute in the `deleteLinks` directive has been removed in favor of applying UIB tooltips directly to the directive instead (and thus allow for customization on placement, `append-to-body` or not, etc.).

- Tooltips are appended to the `<body>` by default in each app (set via `.config` blocks) to avoid potential clashes with styles on the tooltips' parent elements. This can be reversed by setting `append-to-body` to `false` on a per-tooltip basis.

### DIFF
--- a/detailed/assets/javascripts/app.js
+++ b/detailed/assets/javascripts/app.js
@@ -2,6 +2,13 @@
 
 var chaiseRecordApp = angular.module("chaiseRecordApp", ['ngResource', 'ngRoute', 'ui.bootstrap','ui.grid', 'ui.grid.resizeColumns', 'ui.grid.pinning', 'ui.grid.selection', 'ui.grid.moveColumns', 'ui.grid.exporter', 'ui.grid.grouping', 'ui.grid.infiniteScroll', 'ngCookies', 'ngSanitize', 'chaise.utils', 'chaise.navbar']);
 
+// Configure all tooltips to be attached to the body by default. To attach a
+// tooltip on the element instead, set the `tooltip-append-to-body` attribute
+// to `false` on the element.
+chaiseRecordApp.config(['$tooltipProvider', function($tooltipProvider) {
+    $tooltipProvider.options({appendToBody: true});
+}]);
+
 // Refreshes page when fragment identifier changes
 setTimeout(function(){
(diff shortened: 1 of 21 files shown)


detailed/assets/javascripts/app.js
```diff
@@ -2,6 +2,13 @@
 
 var chaiseRecordApp = angular.module("chaiseRecordApp", ['ngResource', 'ngRoute', 'ui.bootstrap','ui.grid', 'ui.grid.resizeColumns', 'ui.grid.pinning', 'ui.grid.selection', 'ui.grid.moveColumns', 'ui.grid.exporter', 'ui.grid.grouping', 'ui.grid.infiniteScroll', 'ngCookies', 'ngSanitize', 'chaise.utils', 'chaise.navbar']);
 
+// Configure all tooltips to be attached to the body by default. To attach a
+// tooltip on the element instead, set the `tooltip-append-to-body` attribute
+// to `false` on the element.
+chaiseRecordApp.config(['$tooltipProvider', function($tooltipProvider) {
+    $tooltipProvider.options({appendToBody: true});
+}]);
+
```
howdyjessie: Note: RecordSet, RecordEdit, Viewer, and Record apps uses version 1.3.2 of Angular UI, and the tooltip provider's name in that version is `$uibTooltipProvider`. The "Detailed" app uses its own, older, version of Angular UI, and the provider for that version of tooltip is called `$tooltipProvider`, which is why this config block is different from the other apps'.

I decided to continue allowing Detailed app to use its own version of Angular UI because it's an older app we don't develop new features for anymore and I don't want to force this app to upgrade and potentially break features that use other parts of Angular UI.

hongsudt:
The plan on the detailed app sounds good.. I think all deployments have
been migrated to record 2.. We might be able to even delete this branch
after we do one final confirm during a division meeting after new year.

Issue #899, opened 2016-12-21
More/Less (Ellipses) incorrect after table reload


shinyichen:
This happens when the record table's data has been updated, we get rows with more/less link when it's not necessary. It's because the table's ellipses is not being re-calculated.  This could happen when clicking on previous/next page, or after table refreshes with a new record. 

The example in the screenshot. Before adding record 103, row 3 in table was record 263 where some of its content were wrapped with more link. After adding a new record 103, row 3 is now replaced with new content, but the ellipses did not get updated.

<img width="1356" alt="screen shot 2016-12-21 at 2 06 48 pm" src="https://cloud.githubusercontent.com/assets/636361/21407943/ef126d0a-c786-11e6-80e0-c92dcd334909.png">

chiragsanghvi:
@shinyichen I am unable to reproduce it. It is on your branch or Master ?

PR #900, opened 2016-12-22
More Less issue in tables when paginated or searched

chiragsanghvi:
This PR fixes issue issue #899. The problem was with the ellipsis directive which was not registering to listen to rowvalue changes.

Whenever the table content was updated, the ellipsis code had to be rerun which wasn't being done.

Now there is an explicit watcher on the row-values, which is triggered on change in values.

Can be checked [here](https://dev.isrd.isi.edu/~chirag/chaise/recordset/#1/legacy:dataset@sort(id)?limit=15)

### DIFF
--- a/common/ellipses.js
+++ b/common/ellipses.js
@@ -67,7 +67,14 @@
                             }
                         };      
 
-                        resizeRow();
+                        scope.$watchCollection('rowValues', function (v) {
+                            $timeout(function() {
+                                timerCount = 0;
+                                containsOverflow = false, oldHeights = [];
+                                resizeRow();
+                            }, 10);
+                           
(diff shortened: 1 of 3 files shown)


common/ellipses.js
```diff
@@ -67,7 +67,14 @@
                             }
                         };      
 
-                        resizeRow();
+                        scope.$watchCollection('rowValues', function (v) {
+                            $timeout(function() {
+                                timerCount = 0;
+                                containsOverflow = false, oldHeights = [];
```
shinyichen: I don't see `oldHeights` being set anywhere in this file. Is it suppose to be set somewhere?

PR #904, opened 2016-12-23
Recordset tests

chiragsanghvi:
This PR converts sleep to wait in recordset tests as discussed in issue #853 .


chiragsanghvi:
Merging this as all tests are passing. The total duration has reduced to 34 minutes from 49 after changing sleeps to wait in search and recordset tests.

Issue #838, opened 2016-12-07
create a pseudo-column that contains a list of actions that users can do


hongsudt:
@hongsudt commented on [Thu Dec 01 2016](https://github.com/informatics-isi-edu/ermrestjs/issues/298)

- This column should be the first column (left most). 
- The column name should be "Actions". The value should be linkable icons (with tooptips) for `view`, `unlink` or `delete`, and `edit`.
- In the compact/brief context, if the related table is derived based on pure and binary association table, `unlink` should be provided, otherwise, `delete` should be provided. `Unlink` deletes an entry in the association table. `Delete` deletes the actual row in the table. 

@robes @karlcz I am making an assumption that we can determine whether the user can delete or create at the row level, which I am not sure whether it has to eventually be an asynchronous call or not. 
   

---

@robes commented on [Fri Dec 02 2016](https://github.com/informatics-isi-edu/ermrestjs/issues/298#issuecomment-264587093)

@hongsudt I think the reference api covers row-level security but of course that is not and cannot be implemented given the current mechanism that ermrest uses based on postgres row level security.

---

@hongsudt commented on [Tue Dec 06 2016](https://github.com/informatics-isi-edu/ermrestjs/issues/298#issuecomment-265342677)

@shinyichen we decided to implement this in chaise. I am moving this issue to chaise repo.

shinyichen:
1) Is the action column always shown? For example, it may not be useful when in the popup selection modal. What should the visibility be based on? App decides?
2) Hide or disable edit/delete icons when not logged in?

hongsudt:
For Glyphicons, what about: in the compact or compact/brief context,
* use glyphicon-list for view. 
* use trashcan for delete. For a related table that is generated based on pure and binary association,  user a remove icon (an x) for unlink instead of trash. This will delete the entry in the association table. The table should be refreshed upon deletion.
* use pencil for editing. Similar to pre-filled add, the related table should be refresh after the editing is done. 
 
In the search modal, use an ok font (check) for select. The rest can be the same as the compact context..

hongsudt:
Instead of using cookies or local storage, we should try Karl's suggestion on DOM manipulation through windows.opener with the suggested flow below. With this, we don't have to deal with garbage collection. 

The general recordedit flow to avoid using cookies for signalling between parent and child:

1. parent defines entry points which consult local hashmaps etc.
   -- window.getPrefill(url) function
   -- window.childChanged(url, ...) function
[...]

PR #912, opened 2017-01-06
Action Colum

shinyichen:
PR for #838. Create an action column. 
* In recordset, column should have view|edit|delete actions. 
* In record, table have view|edit|delete or unlink.
* Update the table after delete or edit made.
* In recordedit modal, column should only have select action.
* Remove row click action from table.

### DIFF
--- a/common/ellipses.js
+++ b/common/ellipses.js
@@ -12,6 +12,7 @@
                 scope: {
                     tuple: '=',
                     rowValues: '=', // tuple's values
+                    context: "=",
                     config: '=',    // {viewable, editable, deletable, selectable}
                     onRowClickBind: '=?'
                 },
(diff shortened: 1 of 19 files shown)


shinyichen:
You can test them here.
https://dev.isrd.isi.edu/~jchen/chaise/recordset/#1/legacy:dataset

RFSH:
These are the bugs that I found while testing the behaviour on chaise.

- In case of edit on pure binary association tables, should use the same reference that is being used for unlink.
- When you edit a related reference, record page gets updated automatically but it is not the case for pure binary association tables (only updates on create and not edit).

-----
you can test it here: [https://dev.isrd.isi.edu/~ashafaei/chaise/record/#1/legacy:dataset/id=9](https://dev.isrd.isi.edu/~ashafaei/chaise/record/#1/legacy:dataset/id=9)

shinyichen:
The generated edit link (when using entryEdit context), gives you entry in the associated table.  https://dev.isrd.isi.edu/~jchen/chaise/record/#1/legacy:data_type/term=Facial%20Images%2FLandmarks
(Is that what you meant?)

shinyichen:
I pushed the changes that addressed the two issues. @RFSH

Issue #870, opened 2016-12-15
multi-entry edit


hongsudt:
We should add an edit button to the recordset and search page that will trigger a set edit.

PR #914, opened 2017-01-09
Multi edit for recordedit

jrchudy:
This PR addresses issue #870, allowing the multi-edit functionality for the recordedit page. Recordedit expects a uri in the form of `.../t:s/<column-name>=<val>;<column-name>=<val>;...`. This will show however many records that the user defines in the uri. 

After submitting the data, the user will be shown the resultset view of the recordedit app rather than being redirected to recordset. This can be tested from [here](https://dev.isrd.isi.edu/~jchudy/chaise/recordedit/#1/legacy:dataset/id=9;id=10).


howdyjessie:
Code and demo looks good to me, merging this PR..

Issue #922, opened 2017-01-13
Submit and display timestamp and timestamptz values as local time


hongsudt:
A *timestamptz* value is stored as a single point in time in Postgres. When the value is retrieved, the value is in whatever time zone the database is in.

A *timestamp* value is stored as a string with the date and time; time zone info will be dropped in Postgres.

When submitting values for timestamp and timestamptz columns, the app should just submit the values as browser's local time. (The app currently converts to UTC before submitting, which is unnecessary.)

When displaying *timestamp* value, display whatever is in the db (the date and time, no need to convert to local time because there's no time zone info attached anyway)

When displaying *timestamptz* value, convert that value to browser's local time.

howdyjessie:
Note to self: Test this in `09-recordedit.types.spec.js`

howdyjessie:
This has been addressed in PRs #934 (implementation) and #942 (tests).

Issue #863, opened 2016-12-14
Error: merdiem is null when editing a page w/ a null timestamp


karlcz:
Start with a record that has a null timestamptz column, e.g. 

https://synapse-staging.isrd.isi.edu/chaise/recordedit/#1/Zebrafish:Behavior/ID=BhvZfDsy20160616A1@sort(ID::desc::)

If I try to edit the null `Room Date` column, the `AM` button does not work when trying to enter a new date, and submission fails too.

If I click the `Now` button first to reinitialize the field, I can then edit the value and submit succeeds.

This is repeatable on Firefox and Chrome

howdyjessie:
@karlcz The change is live on synapse-dev now; could you review it when you have a chance? [This record](https://synapse-dev.isrd.isi.edu/chaise/recordedit/#1/Zebrafish:Behavior/ID=BhvZfCza20160815A1@sort(ID::desc::)) has a null timestamp column.

You can test it by attempting to enter a date and time, toggling the AM/PM button, and then submitting.

For a timestamp column that _isn't required_ (`nullok` is true), if a user enters either a date or time (but not both), should the user see an error or warning? Right now the app just submits an incomplete timestamp as `null`, which might be confusing to users who didn't know/forgot to fully fill in a timestamp.

karlcz:
I would think that the form-validation logic would flag an incomplete
field as an error and give an error message, similar to how we would
flag any other errors such as missing required fields or invalid
numbers.

karlcz:
In addition to the need for validation and error signalling, I see one problem with this version.

Clicking into the date or time field changes it from gray (disabled) text to black (enabled) text even though the control is still acting like it is unset.  I.e. I got it to a state showing:

    2017-12-31  12:00:00

but it silently treated that as incomplete/null because the `12:00:00` was not digits I typed but the result of entering and leaving the time field.

howdyjessie:
Timestamp input validation and gray vs. black text issues have been resolved in PR #934, closing this issue now.

PR #934, opened 2017-01-19
Add timestamp validator

howdyjessie:
This PR addresses comments starting [here](https://github.com/informatics-isi-edu/chaise/issues/863#issuecomment-270800433) on a related issue (#863). 

The 2 problems described are:
1. A lack of proper validation _between_ the fields of a timestamp/timestamptz column. e.g. When a user submits only a time but leaves the date blank, the app will submit it as `null` but (1) how would the user have known that?; and (2) the app should prompt the user for a date field in this case.
2. Misleading placeholder text in the timestamp fields that would turn from gray to black upon input focus and then _stay black_ even if the user types nothing and de-focuses the input. This would make users believe that a value had been automatically filled in for them when in fact the submitted values are still `null`.

To address problem 1: 
- This PR adds a new directive (it is not strictly a validator registered to an ngModel) that validates the date and time fields in a timestamp/timestamptz column's object model and shows the appropriate messages when field(s) are missing/invalid.
- It also initializes the object models for timestamp and timestamptz columns in create mode, so that the validator can check timestamp fields consistently.

For problem 2:
- If the user doesn't type anything, placeholder text returns to gray upon de-focusing a timestamp input field.
- The placeholder text for time input has been changed to from "12:00:00" to "HH:MM:DD" to further indicate to users that the field is null (and to keep it consistent with the date input's "YYYY-MM-DD").

Also, if a user fills in a valid date but has a null time, the app now fills in the time component as "00:00:00" upon submission to ERMrest.

This PR also partially addresses #922 by removing conversions to UTC when the app submits timestamp/timetamptz values to ERMrest, but tests for this will be in a later PR.


common/validators.js
```diff
@@ -97,5 +97,42 @@
                 // });
             }
         };
+    })
+
+    .directive('timestamp', function() {
+        return {
+            restrict: 'A',
```
jrchudy: Maybe this should be changed to `E` and define this as an element directive as a child element to input. I see that this directive relieson the `valicateValues` attribute. Maybe include a doc sort of comment above the directive code?

howdyjessie: @jrchudy Hmm I'm not sure I understand your comment here. What do you mean by "define this as an element directive as  a child element to input?" Is this what you mean?
```
<timestamp validate-values="someValues">
    <input date .....>
</timestamp>
```
I'm also not convinced that this should be an element directive over an attribute directive.

jrchudy: I actually meant it the other way around where this would be a child of `<input>`.  As in:
```
<input>
    <timestamp validate-values>
</input>
```
I'm more curious if you thought about this as a possible solution rather than adding an attribute directive that also relies on another attribute.

howdyjessie: Yeah but `<input> ... </input>` is not valid HTML.

From my understanding of the Angular [docs](https://docs.angularjs.org/guide/directive), attribute would be the preferred practice here, since this new directive is not meant to be used alone.
>**When should I use an attribute versus an element?** Use an element when you are creating a component that is in control of the template. The common case for this is when you are creating a Domain-Specific Language for parts of your template. Use an attribute when you are decorating an existing element with new functionality.

PR #939, opened 2017-01-20
Edit link on recordset checks permissions

jrchudy:
The edit link on `recordset` will show based on the user's permissions and `chaiseConfig.editRecord`. This is so a not logged in user won't see the edit button as well as user's who aren't allowed to modify records.


howdyjessie:
I haven't merged this yet on account of (1) perhaps @hongsudt would like to review this; and (2) Travis builds have been failing due to a hostname issue all day so there doesn't seem to be a point in merging for now.

Issue #849, opened 2016-12-09
Obey app-links annotation after record delete


karlcz:
The record app redirects to the search app unconditionally after delete confirmation, when it should obey the normal app-links configuration so it can be sent to recordset instead.

howdyjessie:
This has been addressed in PR #943 and merged into master.

PR #943, opened 2017-01-21
Use appLink when deleting a record in Record app

howdyjessie:
This PR addresses #849.

The delete record link now takes the reference to the deleted record, gets the reference to the record's table, contextualizes this table's reference to compact, and then fetches its `appLink`.

This PR cannot be merged until informatics-isi-edu/ermrestjs#321 has been merged because it needs the new `reference.table.reference...` feature in ermrestjs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,8 +68,8 @@ before_install:
   - cd ErmrestDataUtils
   - sudo npm install
   - cd ../chaise
-  - cp test/ermrest_config.json /home/ermrest/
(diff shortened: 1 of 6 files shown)


howdyjessie:
It says "All checks have failed" because I had turned on PR builds in Travis for another issue and then forgot to turn it off before I opened this PR. I canceled the started build because Travis builds have been failing all day due to what seems like a hostname issue...(?)

PR #944, opened 2017-01-23
Add Visible Columns Annotation

RFSH:
Add visible columns annotation to schemas that has been used for test cases, to avoid usage of heuristics.


jrchudy:
I talked to @RFSH about this PR today, he said it's okay to merge once updating it with master. It only changes test schemas for a change to `ermrestJS`. This PR is being merged first because it's compatible with `ermrestJS` both before and after that PR is merged.

Issue #846, opened 2016-12-09
Precondition Failed (412) Error When Deleting Record


shinyichen:
Steps to reproduce this problem:
1) Open RecordEdit and submit a new record --> redirects to record page (keep this page open)
2) Open another RecordEdit and create another record. 
3) Go back to the first (Record) page and delete this record gives 412 Precondition Failed error.

The reason:
A new etag is generated when second create request was submitted. Now the first record has an out of date etag and forces it to fail.

Proposed solutions:
1) Better display of the error/information for the user at UI level. (This still seems inconvenient for the user)
2) emrestJS handle a failure case.
3) Karl's comment : it requires UX intervention in general case. ETag mismatch is a conservative signal that something *might* be out of sync. if you reload and find that the data you based your decisions on is still the same, you can just repeat with the newer ETag you found.  if you find a difference, you would want to prompt the user as in, "<displayname> has changed. Please review the latest content and resubmit your request." then present a refreshed page where the user can either delete again or make edits again without silently clobbering changes made in parallel.

howdyjessie:
A sketch from @karlcz on how to deal with 412 errors:
![img_2144](https://cloud.githubusercontent.com/assets/4585431/23228069/19f54304-f8f0-11e6-9ce2-cf69a68afba8.JPG)

howdyjessie:
In Chaise:
On 412 in UPDATE case in recordedit, 
- Show a modal that explains that some data has changed and user must review the data.
- Highlight the inputs where the current data in the DB isn't the same as what the user entered in the input. Show current data in blue text beneath input fields (like validation messages).

howdyjessie:
PR #1006 has been merged, closing now.

PR #950, opened 2017-01-26
Improve 412 error message

howdyjessie:
This PR partially addresses #846. The corresponding https://github.com/informatics-isi-edu/ermrestjs/pull/327url) must be merged before this PR.

It adds a 412 Precondition Failed error class so that if ermrestjs is unable to handle a 412 error successfully, the 412 error will still be passed to Chaise but Chaise will display with a more user-friendly error message that tells the user to refresh the page.


howdyjessie:
After discussing this issue more in standup, it seems like a larger change and further work is needed to address 412s. Closing this PR for now.

PR #966, opened 2017-02-01
Show/hide action column buttons based on user permissions

howdyjessie:
This is a small PR to address the visibility of action column buttons. Instead of being disabled when a user doesn't have permission to perform the action, the action button is hidden.


RFSH:
@howdyjessie Doesn't this change affect the test cases?

howdyjessie:
@RFSH I didn't expect it to because the tests' ACLs allow all the action buttons to be displayed in order to test them. The tests passed for me locally.

jrchudy:
The current set of tests only cover the case when the user has all of the permissions. We have examples of tests that use different permissions in ermrestJS.

RFSH:
@hongsudt The written test cases just test the availability of buttons in different scenarios. We are not testing the opposite direction (e.g. If user doesn't have access, the edit button should not be available.)
Test cases are:
- action columns should show view button that redirects to the record page
- action columns should show edit button that redirects to the recordedit page.
- action columns should show delete button that deletes record.
- action columns should show unlink button that unlinks.

hongsudt:
Jessie will add visibility test cases.

howdyjessie:
Closing this PR to make a new one with tests and a couple other fixes.

PR #973, opened 2017-02-08
One way binding performance

jrchudy:
This PR addresses some performance issues with `record` and `recordedit`. The changes made were to reduce the number of two-way bindings that each app creates. 

On `record` page, I was able to reduce the number of listeners on initial load from about 100 to 70, a 30% decrease.

On `recordedit` page, I was able to reduce the number of listeners on initial load from about 60 to the mid 25. This is roughly a 50% decrease. This doesn't address the overall concern with performance which seems to be caused by the ellipsis directive, but this is a good step towards cleaning up the apps and limiting unnecessary events from bogging down the angular digest cycle.


jrchudy:
I turned on PR builds for this PR to verify the tests were still green after the chaos I waded through yesterday. I have since turned PR builds off again for chaise.

PR #974, opened 2017-02-08
Changed ellipsis implementation to use onload instead of polling

chiragsanghvi:
This **PR** handles the issue with performance related to the more/less behavior.

Initially it used to do polling to check for height changes for dynamic HTML elements like `image` and `iframe`. This used to take a lot of processing power.

Now, the code explicitly attaches `onload` event handlers onto these elements. This handler is called once the image and iframe is loaded and updates the height to add/removes the ellipsis.

You can test it on GPCR for recordedit . 

**Image:** 
https://synapse-dev.isrd.isi.edu/~chirag/chaise/recordset/#1/Zebrafish:Behavior@sort(Image%20Date::desc::)

**Iframe**
https://synapse-dev.isrd.isi.edu/~chirag/chaise/recordset/#1/Zebrafish:Cropped%20Image@sort(Extract%20Date::desc::,ID)?limit=10

### DIFF
--- a/common/ellipses.js
+++ b/common/ellipses.js
@@ -4,8 +4,8 @@
     angular.module('chaise.ellipses', [])
 
 
-        .directive('ellipses', ['$sce', '$timeout', 'AlertsService', '$uibModal', '$log', 'MathUtils', 'UriUtils', '$window',
-            function($sce, $timeout, AlertsService, $uibModal, $log, MathUtils, UriUtils, $window) {
+        .directive('ellipses', ['$sce', '$timeout', 'AlertsService', '$uibModal', '$log', 'MathUtils', 'UriUtils', '$window', 'UiUtils',
+            function($sce, $timeout, AlertsService, $uibModal, $log, MathUtils, UriUtils, $window, UiUtils) {
 
             return {
                 restrict: 'AE',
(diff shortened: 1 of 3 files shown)


common/ellipses.js
```diff
@@ -181,36 +181,56 @@
...
                                 }
-                                $timeout(function() {
-                                    resizeRow();
-                                }, 50);
                             }
                         };
 
+                        // Watch for change in rowValues, this is useful in case of pagination
+                        // As Angular just changes the content and doesnot destroys elements
                         scope.$watchCollection('rowValues', function (v) {
```
robes: If I understand correctly, this `function (v)...` gets called when `rowValues` changes, which in turn delays `10` ms in case Angular (or the web browser layout engine?) is working on the layout, after which time calls the `resizeRow()` function. Is that right?

And what happens if the layout isn't settled down by the (10ms) time we call resizeRow, do we have a race condition possible or do we have a way of dealing with this possibility?

chiragsanghvi: So basically, when the directive is called, the html is still not displayed. So we need to add a timeout to call resizeRow. 

As Javascript uses event queue. The `resizeRow` function is called after the html is rendered by the directive because of the timeout. Even if I make the timeout to 1, it will still work.

jrchudy:
This doesn't seem to be working correctly for some cases. If you go to https://dev.gpcrconsortium.org/~chirag/chaise/recordedit/#1/experiments:sample and select a construct, then click the next button to go to the second page, you can see that some of the cells are not being reduced in size.

image

chiragsanghvi:
I have made the changes as requested to handle race conditions during paging.

I fixed the issue with the View action button too as mentioned [here](https://github.com/informatics-isi-edu/chaise/issues/975). The problem was with the [watchCollection](https://github.com/informatics-isi-edu/chaise/blob/master/common/ellipses.js#L206) statement. It was inside the [if statement](https://github.com/informatics-isi-edu/chaise/blob/master/common/ellipses.js#L164) that checks for `maxRecordsetRowHeight` is not `false`.

We got this issue, when we set `maxRecordsetRowHeight` to `false`. Due to which the the  `watchCollection` statement was never executed.

hongsudt:
@chiragsanghvi if you have addressed Josh's comment, let's merge this PR and please assign @karlcz to review relevant issues related to this PR.

PR #983, opened 2017-02-14
Autosearch

chiragsanghvi:
This PR fixes the problem with search in recordset table as mentioned in issue #933.

The fix now waits for no user input duration to 1 second and then triggers search. In addition it will queue background search keeping the last one only, instead of firing requests for all at once.

It will reject any background request data that were fired before foreground and returned after it.

https://dev.gpcrconsortium.org/~chirag/chaise/recordset/#1/experiments:sample/*::ciregexp::GPCRUSC20161012SAM@sort(id)?limit=25


common/table.js
```diff
@@ -56,15 +56,54 @@
      *      });
      */
     .factory('recordTableUtils', ['DataUtils', '$timeout', function(DataUtils, $timeout) {
-        function read(scope) {
+
+        // This method sets backgroundSearch states depending upon various parameters
```
robes: You have not defined a 'background' vs 'foreground' search. (I can know from earlier conversations.) These need to be documented in here to explain what that means and what you are doing. Even a few sentences in a comment is good.

common/table.js
```diff
@@ -56,15 +56,54 @@
...
+        // else we should reject the data
+        function setSearchStates(scope, isBackground, searchTerm) {
+
+            // If request is background
+            if (isBackground) {
+                // If there is a term in search queue for background and there is no foreground search going on then
+                // Fire the request for the term in the queue and return false
+                // Else empty the queue and set backgroundSearch false
+                if (scope.vm.backgroundSearchQueue && !scope.vm.foregroundSearch) {
+                    scope.vm.search = scope.vm.backgroundSearchQueue
```
robes: `backgroundSearchQueue` technically is not a _Queue_ so let's not call it that. It's confusing. I gather from reading the code that it is the _pending_ background search term or something like that. Might rename it `backgroundPendingSearchTerm` or something.

common/table.js
chiragsanghvi: Added more comments.

common/table.js
chiragsanghvi: Changed the variable name to backgroundSearchpendingTerm

chiragsanghvi:
@hongsudt  @robes  Should I merge it ?

robes:
yes, looks good.

PR #984, opened 2017-02-14
Added separate method in chaise.page.js to performlogin

chiragsanghvi:
This PR adds a separate method `performLogin` to `chaise.page.js` that performs login on behalf of a user when passed a cookie.

Documentation can be found [here](https://github.com/informatics-isi-edu/chaise/wiki/E2E-tests-guide#setting-webauthn-cookies-in-test-cases)

### DIFF
--- a/test/e2e/utils/chaise.page.js
+++ b/test/e2e/utils/chaise.page.js
@@ -873,6 +873,25 @@ function chaisePage() {
     this.waitForElementCondition = function(condition, timeout) {
         return browser.wait(condition, timeout || browser.params.defaultTimeout);
     };
+
+    this.performLogin = function(cookie, defer) {
+
+        defer = defer || require('q').defer();
+
+        browser.get(process.env.CHAISE_BASE_URL + "/login/");
+        browser.ignoreSynchronization = true;
+          
+        browser.wait(protractor.ExpectedConditions.visibilityOf(element(by.id("loginApp"))), browser.params.defaultTimeout).then(function() {
+            browser.driver.executeScript('document.cookie="' + cookie + ';path=/;' + (process.env.TRAVIS ? '"' : 'secure;"')).then(function() {
+              browser.ignoreSynchronization = false;
+              defer.resolve();
+            });
+        }, function() {
+            defer.reject();
+        });
+
+        return defer;
+    };
 };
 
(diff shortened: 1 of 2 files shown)


hongsudt:
@chiragsanghvi has anyone review this PR prior to merging?

test/e2e/utils/chaise.page.js
```diff
@@ -873,6 +873,25 @@ function chaisePage() {
...
+        browser.get(process.env.CHAISE_BASE_URL + "/login/");
+        browser.ignoreSynchronization = true;
+          
+        browser.wait(protractor.ExpectedConditions.visibilityOf(element(by.id("loginApp"))), browser.params.defaultTimeout).then(function() {
+            browser.driver.executeScript('document.cookie="' + cookie + ';path=/;' + (process.env.TRAVIS ? '"' : 'secure;"')).then(function() {
+              browser.ignoreSynchronization = false;
+              defer.resolve();
+            });
+        }, function() {
+            defer.reject();
```
howdyjessie: Would it be necessary to set ignoreSynchronization to false in the failure case as well?

chiragsanghvi: Not necessarily. As this would mean that the test is unable to open the login page due to some server issue. Due to which all the tests will fail. 

We can add, but uptil now I have never encountered that issue.

howdyjessie: Hmm I see. Then my next question is why is ignoreSync set to false on L886? I can understand why it was set to true on L882 but I'm not sure why it's set to false later.

chiragsanghvi: As mentioned in the e2e test wiki regarding [syncronization issues](https://github.com/informatics-isi-edu/chaise/wiki/E2E-tests-guide#synchronization-issues) we don't use Angular Router so thats the reason we set it to true. 

Once we have visited the page and set the document cookie successfully, I set it to false on L886. This is so that we don't alter the synchronization state and bring it back to the state it was before.

chiragsanghvi:
@hongsudt The change was in the test framework and had no side effects. It was just some code optimization.

hongsudt:
OK. So this feature should now be available on the dev server, correct?

Issue #982, opened 2017-02-11
recordedit should handle 401 properly


hongsudt:
Currently, the app will just print out "Error 401 Unauthorized The requested catalog/1 is forbidden."  We need to handle this so that users won't loose the data they have entered. 

@chiragsanghvi I think you are working on this for the file upload. Maybe we should prioritize this piece and add it to general recordedit first.

PR #990, opened 2017-02-17
Implemented new window concept for Login

chiragsanghvi:
This PR fixes issue https://github.com/informatics-isi-edu/chaise/issues/982.

Whenever a 401 Unauthorized error is encountered during add/edit/delete  a login window will open. Once you're done submitting your credentials, it will close and focus back on the main tab, restarting the submit automatically.

https://dev.gpcrconsortium.org/~chirag/chaise/recordedit/#1/experiments:sample
To test the flow

-Open above link and fill up the form
-Now open the same link in another tab and press logout
-Now go back to previous page where you had filled the form and click submit
-You should be presented with a login window and on successful login you should be able to create/edit/delete the record.

The code handles non-globus authentication flow too


howdyjessie:
General question: What about the case when the user doesn't have a valid session anymore but still needs to access something from the DB? e.g. When a user wants to change foreign key input, this would result in a 401 and they'd see the spinner spinning forever. Shouldn't the new window login also cover this case?

chiragsanghvi:
@howdyjessie I was thinking about adding tests, but was not sure how will I handle the login flow as in the tests we usually set the cookie. The login flow needs the user to use Globus login locally.

In addition, we need to identify all the cases where we are fetching content after the page load. I handles those which were present in recordedit.

chiragsanghvi:
@howdyjessie I have made the changes for fetching records in recordset as well as modal in recordedit.

common/authen.js
```diff
@@ -68,6 +68,56 @@
...
+                var config = {
+                    headers: {
+                        'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8',
+                        'Accept': 'application/json'
+                    }
+                };
+
+                window.addEventListener('message', function(args) {
+                    console.log("Inside message listener");
+                    if (args && args.data && (typeof args.data == 'string')) {
```
howdyjessie: Also, unnecessary console.log?

chiragsanghvi: I am unsure about identifying the identity as tabs usually are incapable of accessing the url's of other tabs. Whenever you want to post a message to a window, you need to have  a handle to it. Which in our case the window has it in `window.opener`. So we really won't need that information.

chiragsanghvi:
@howdyjessie @hongsudt Should I merge this PR ?

hongsudt:
I have asked Josh to review this PR yesterday..

jrchudy:
The changes look good to me. I tested the functionality locally, and it works for me as well.

common/authen.js
jrchudy: Using the `referrerId` should be sufficient.

Issue #991, opened 2017-02-21
Recordset remove search phrase


jrchudy:
Recordset should have an `X` in the search box to remove the search term rather than a `Reset` button. The `X` should have the same behavior as the `Reset` button which includes reseting the `resultset` to an empty search.

Issue #907, opened 2016-12-24
Info button on recordset page


carlkesselman:
The "?" button on the recordset page seems unnecessary to me and inconsistent in behavior.  It looks like a button, but it doesn't do anything if you push it, and it has information about the search box.  I would put the contents of the tooltip for this button on the magnifying glass (which has no tool tip) and eliminate this button.

hongsudt:
@karlcz @robes @jrchudy @howdyjessie @RFSH @chiragsanghvi any objection?

jrchudy:
Seems reasonable to me and should be simple to change.

howdyjessie:
No objection here.

robes:
No objection.

PR #992, opened 2017-02-22
Recordset search cleanup

jrchudy:
This PR addresses issues #907 and #991. 
 - The info button on recordset has been removed and the corresponding information from the tooltip is now present on the search button.
 - The reset button has been removed in favor of having an `X` inside of the search input box with the same behavior.

### DIFF
--- a/common/styles/app.css
+++ b/common/styles/app.css
@@ -519,3 +519,8 @@ input.search-progress {
     background-size: 25px 25px;
     background-position-x: 99%;
 }
+
+/************* search input ****************/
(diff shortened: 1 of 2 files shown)


chiragsanghvi:
The `X` inside the input box interferes with the loader in the input control when background search is going on. We should have some way to hide the `X` when background search is going on.

jrchudy:
I changed the logic for hiding the `X` to address your comment.

chiragsanghvi:
Changes look good to me. You can merge it

PR #998, opened 2017-02-24
Changed existing login flow for 401 to use iframes instead of new window

chiragsanghvi:
As discussed in the meeting, I have added a layover.

After noticing multiple ways to allow a popup for different browsers, I thought that instead of opening  a new window I can open the login window inside an iframe on the same page. Everything else will be grayed out in background, when the login page is shown.


jrchudy:
I like the use of an iframe that you define like a modal.

karlcz:
I have several concerns with using iframe for login:

1. We don't know if every 3rd party oauth provider will allow their login flow to run in our iframe (I believe they could block it with x-iframe-options...). It's impractical for us to test every provider available to users via Globus.
2. The oauth flows are designed for top-level presentation so we could have ugly nested branding/navbar layouts?
3. It seems like the UX around error cases in the login flow could get pretty strange.

For the last point, imagine that people end up just looking at some Globus page inside this iframe because the login didn't complete. That whole page content may confuse the user versus having an static overlay on our app page reminding them that they must complete login to proceed and offering links to attempt login again.

chiragsanghvi:
I agree with what @karlcz says. I changed the implementation to iframe, because when I was creating the instructions to allow the popup to open for users, it was just too problematic and different for every browser.

chiragsanghvi:
Made the changes as requested. Here are the instructions to enable the modal for each browser


`Chrome`

chrome

`Safari`

safari

`Firefox`

firefox

`Edge`

edge

`IE`

ie

hongsudt:
@chiragsanghvi  I am making an assumption that an explicit login click-event to open a new window works for all browsers. If that's the case, we can simplify the instruction message to be below:

```
Your session has expired. Please login to continue.

                                      Login button
```
The instruction window should be centered vertically and horizontally. When a window is open, please center the window relative to the current window. 

I tested this on firefox, the window was open with pop-up disabled, and it was closed automatically after log-in.

chiragsanghvi:
@hongsudt Made the changes as discussed during the standup. Now user presses the login button explicitly. In addition the modal is not dismissible until you login.

hongsudt:
Chirag, can you make the instruction modal dismiss able by having an X at
the upper right corner?

chiragsanghvi:
Yes sure.

chiragsanghvi:
@hongsudt I have made the changes as requested.

common/authen.js
```diff
@@ -109,8 +102,74 @@
...
                         }
                         login_url += '&method=' + method + '&action=' + action + '&text=' + text + '&hidden=' + hidden + '&?referrerid=' + referrerId;
-                        $window.open(login_url, '_blank','width=800,height=600');
                     }
+
+                    var params = {
+                        login_url: login_url
+                    };
+                    
+                    params.message = "Your session has expired."
```
jrchudy: This needs to be added to the `messageMap`.

common/modal.js
```diff
@@ -25,6 +25,30 @@
...
+        var vm = this;
+        params.login_url = $sce.trustAsResourceUrl(params.login_url);
+        vm.params = params;
+
+        vm.openWindow = function() {
+
+            var x = window.innerWidth/2 - 800/2;
+            var y = window.innerHeight/2 - 600/2;
+
+            window.open(params.login_url, '_blank','width=800,height=600,left=' + x + ',top=' + y);
```
jrchudy: This opens the window in the bottom right corner of my display no matter what size my browser is.

common/templates/loginDialog.modal.html
```diff
@@ -0,0 +1,16 @@
+<div class="modal-header">
+	<button class="btn btn-default pull-right modal-close" type="button" ng-click="ctrl.cancel()">X</button>
+    <h3 class="modal-title text-center">{{::ctrl.params.message}} Please login to continue.</h3>
```
jrchudy: This needs to be added to the `messageMap`.

common/templates/loginDialog.modal.html
```diff
@@ -0,0 +1,16 @@
...
+</div>
+<div class="modal-body">
+	<div>
+		<div class="row">
+		    <div class="col-md-12 col-sm-12 col-xs-12">
+		    	<center>
+					<h4>To open the login window press</h4>
```
jrchudy: This needs to be added to the `messageMap`.

common/utils.js
```diff
@@ -466,6 +466,11 @@
             return result;
         }
 
+        function isBrowserIE() {
+            //Internet Explorer 6-11
+            return /*@cc_on!@*/false || !!document.documentMode;
```
jrchudy: not sure how this would detect if the browser is IE. It looks like it would just always return false.

common/modal.js
chiragsanghvi: This is because you extended monitor connected to your monitor.

common/utils.js
chiragsanghvi: This works, as I have tested it in IE. You can try the url in IE and validate it.

chiragsanghvi:
Made changes and added messages to the messageMap as requested.

common/utils.js
jrchudy: From [here](http://stackoverflow.com/questions/10072816/how-does-this-ie-check-work):
"If IE sees this comment, it actually compiles the contents of comment blocks as pure code."

"IE compiles the ! as if it were written into source so it handles the code as if (!false), while all other browsers treat it as if (false)."

I was more curious about understanding why it works. `cc_on` suggests that this comment needs to be conditionally compiled so it writes the `!` into the compiled source.

common/modal.js
jrchudy: Aha. It does behave differently after disconnecting my thunderbolt display.

jrchudy:
Changes approved. Make sure that the UX flow is acceptable with @hongsudt before merging this PR.

PR #1006, opened 2017-03-01
Handle 412 error on delete

howdyjessie:
This PR addresses #846. **This PR cannot be merged until informatics-isi-edu/ermrestjs#362 has been merged.**

- All Reference.delete operations now require a tuples argument, so changes in Chaise have been made accordingly.
- In Record, when a user attempts to delete the main entity and the app encounters a 412 error from ermrestjs, then a modal is shown to explain why the record can't be deleted right now and then prompts the user to refresh the page. The same thing happens in RecordEdit app.
- In RecordSet, a modal also appears but instead of asking the user to refresh the page, it tells user to review the table for the most recent values from the DB. When the user clicks OK, the table reads the page from the DB and updates the table cells.


common/errors.js
```diff
@@ -66,6 +66,9 @@
 
             if (exception instanceof ERMrest.UnauthorizedError || exception.code == errorNames.unauthorized) {
                 Session.login($window.location.href);
+            } else if (exception instanceof ERMrest.PreconditionFailedError) {
+                // A more useful message for 412 Precondition Failed
+                AlertsService.addAlert({type: 'warning', message: 'This page is out of sync with the server. Please refresh the page and try again.'});
```
jrchudy: This message should be moved to `messageMap`.

common/templates/refresh.modal.html
```diff
@@ -0,0 +1,9 @@
+<div class="modal-header">
+    <h3 class="modal-title">Page Refresh Required</h3>
```
jrchudy: This message should be moved to `messageMap`.

common/templates/uiChange.modal.html
```diff
@@ -0,0 +1,9 @@
+<div class="modal-header">
+    <h3 class="modal-title">Review Modified Record</h3>
```
jrchudy: This message should be moved to `messageMap`.

howdyjessie:
@jrchudy I've modified the 412 errors to use `messageMap`.

Issue #344, opened 2016-05-12
Record app shouldn't change case on table names when already mixed case


karlcz:
I have a model where I already set mixed (title) case in my table names in the schema.  The record app is displaying the table names to all UPPER CASE which I think is wrong.  If we're going to have a heuristic for automatic reformatting, I think it needs to be more conservative in deciding when the names need case changes.

hongsudt:
Will not fix in the current record app, but move this to the new record app.

hongsudt:
This is no longer relevant due to annotation on column name transformation.

PR #1020, opened 2017-03-14
Domain filter pattern implemented for Foreign Key select

jrchudy:
This PR addresses issue [#344](https://github.com/informatics-isi-edu/ermrestjs/issues/344) in `ermrestJS`. The changes here in `chaise` need to be merged after the changes that are made to `ermrestJS` in [this PR](https://github.com/informatics-isi-edu/ermrestjs/pull/365) are merged.

- The most notable change is factoring out the `populateSubmissionRow` function. I made sure this still works for record add and record edit (e2e tests still pass)
- in the `searchPopup` function, it calls the function `filteredRef()`, this is a newly defined function in `ermrestJS` that properly defines the uri for a reference based on a potential `domain-filter-pattern` on the `foreign-key` annotation.


recordedit/form.controller.js
```diff
@@ -382,10 +362,24 @@
...
+            }else if (context.queryParams.copy) {
+                originalTuple = $rootScope.tuples[0];
+            } else {
+                originalTuple = null;
+                editOrCopy = false;
+            }
+
+            var submissionRow = populateSubmissionRow(vm.recordEditModel.rows[rowIndex], vm.recordEditModel.submissionRows[rowIndex], originalTuple, $rootScope.reference.columns, editOrCopy);
+
+            params.reference = column.reference.filteredRef(submissionRow).contextualize.compactSelect;
```
howdyjessie: If `filteredRef` is on each `ReferenceColumn`, should it be `column.filteredRef(submissionRow).contextualize.compactSelect`?

recordedit/form.controller.js
```diff
@@ -484,6 +478,40 @@
...
+                        var foreignTableColumn = column.foreignKey.mapping.get(referenceColumn);
+                        // check if value is set in submission data yet
+                        if (!submissionRow[referenceColumn.name]) {
+                            /**
+                             * User didn't change the foreign key, copy the value over to the submission data with the proper column name
+                             * In the case of edit, the originating value is set on $rootScope.tuples.data. Use that value if the user didn't touch it (value could be null, which is fine, just means it was unset)
+                             * In the case of create, the value is unset if it is not present in submissionRows and because it's newly created it doesn't have a value to fallback to, so use null
+                            **/
+                            if (editOrCopy) {
+                                submissionRow[referenceColumn.name] = originalTuple.data[referenceColumn.name] || null;
```
howdyjessie: The comparison might be a problem if `originalTuple.data[referenceColumn.name]` was ever `0` or some other falsy value.

recordedit/form.controller.js
jrchudy: You are correct, I will change this.

recordedit/form.controller.js
jrchudy: Data will almost always be an object, so we don't need to be worried about that. If it isn't an object, it will be set to undefined.

howdyjessie: Yes but what if the actual value at `originalTuple.data[referenceColumn.name]` is `0`? Then this would always evaluate to `null`.

jrchudy:
@howdyjessie I addresses your comments. I made sure the e2e tests still pass as well.

Issue #1012, opened 2017-03-03
Refactor error handling to comform to chaise dev guideline


hongsudt:
I have updated the [Chaise Dev Guide wiki](https://github.com/informatics-isi-edu/chaise/wiki/Chaise-Dev-Guide) to reflect our discussion on Thur. We should try to review the code and refactor it to conform to the convention..  What about? 
- [x] record: @howdyjessie 
- [x] recordedit: @jrchudy  
- [x] recordset: @chiragsanghvi 
- [x] ermrestjs: @RFSH 
- [x] common: @howdyjessie 

How does the above breakdown sound?

jrchudy:
That sounds good to me. 

One thing to note: 
`ermrestJS` doesn't necessary need a robust error handler that follows the convention we outlined because `http.js` acts as an interceptor for all requests in `ermrestJS`. @RFSH do you want to think about this a bit for `ermrestJS`? 

My thoughts:
That common error handling can be managed in `http.js` for generic cases if you don't need to act on the response object in the error callback for that specific API request. If the response needs to be acted on for a failure case, that code would go in the error callback and then maybe a `catchAll` will need to be defined? So maybe we should have this same convention in `ermrestJS` to have a `finally` case defined for all errors. This finally would ultimately just re-throw the error though so that `chaise` can then act on it and have it's `catchAll` raise the proper notification for the user.

robes:
@jrchudy I'd like to hear about a specific example where such an error callback in http.js would be useful.

robes:
@hongsudt We should look into implementing a 'decorator' for angular's `$exceptionHandler` service. Not sure if that's another bullet here to be assigned or a separate issue.

References here:
https://github.com/informatics-isi-edu/chaise/wiki/Chaise-Dev-Guide#guideline-for-angular-components

jrchudy:
I found out today that you can open a PR for a branch, but none of the files will show up for comments unless you make a change to each file you want to comment on.

For `recordedit`, I created a branch and went to each `.js` file and added a return character. After I pushed my branch to github and opened a PR so I could comment on the individual lines in the `.js` files.

chiragsanghvi:
@hongsudt Recordset already makes call to catchAll in all error scenarios. The only place where we need to make change is for the 401 login handling. This code is partially going to be a part of ermrestjs as we discussed. `http.js` in Ermrestjs will have a callback that will be called on `401` error. This callback will be provided by Chaise.

[...]

PR #1024, opened 2017-03-16
Recordedit Errors

jrchudy:
This PR is opened to address the areas in `recordedit` where error handling needs to be updated to be consistent with our error handling guidelines.


jrchudy:
As I noted in the issue for the error handling (#1012), it wasn't exactly trivial how to make this available in a PR for others to comment accordingly. What I had to do was make comments inline like:
`// TODO: text here`
so that those changes would show up in git and that others can then comment using git's tools.

I asked for a review from each of you just so there can be some discussion about the changes that should be made.

jrchudy:
If git won't let you comment in a place that I haven't made a change (this happens when you want to make a git comment on a line that is nowhere near a change that was made), leave a comment on the PR with a link to that line number like this:
https://github.com/informatics-isi-edu/chaise/blob/re-errors/recordedit/form.controller.js#L31

recordedit/recordEdit.app.js
```diff
@@ -318,6 +321,7 @@
                 }
             });
         } catch (exception) {
```
RFSH: This is a universal try-catch for this app. Should we instead wrap the code in `Utils.catchWrapper`?

jrchudy: I don't think so. That would force us to move the code for the run block into a function just so we can wrap it with a try/catch.

recordedit/recordEdit.app.js
```diff
@@ -233,6 +234,7 @@
                             $log.warn(response);
                             throw response;
                         }).catch(function readCatch(exception) {
+                            // TODO: catchAll here too or should this be logic specific to recordedit?
                             ErrorService.errorPopup(exception.message, exception.code, "home page");
```
howdyjessie: I don't think a catchAll is necessary here since we're just redirecting to home page no matter what.

That is, unless we want to log it for monitoring/informational purposes. If so, then a catchAll could be put here to log it.

recordedit/form.controller.js
```diff
@@ -227,6 +234,7 @@
                             submit();
                         });
                     } else {
+                        // TODO: this needs to be removed and a .catch() added
                         vm.showSubmissionError(exception);
                         vm.submissionButtonDisabled = false;
                     }
```
hongsudt: The above code block can be removed in favor of a .catch() that will catch rejected promises and errors.

recordedit/form.controller.js
```diff
@@ -277,6 +285,7 @@
                             submit();
                         });
                     } else {
+                        // TODO: this needs to be removed and a catch added here
                         vm.showSubmissionError(exception);
                         vm.submissionButtonDisabled = false;
                     }
```
hongsudt: The above code block can be removed in favor of a .catch() that will catch rejected promises and errors.

recordedit/form.controller.js
```diff
@@ -356,6 +366,7 @@
                             ErrorService.catchAll(error);
                         });
                     } else {
+                        // TODO: rethrow error here
                         vm.showSubmissionError(response);
                     }
                 }).catch(function (error) {
```
hongsudt: pass `catchAll()` fn as the function rather than defining an anonymous function that calls it.

recordedit/form.controller.js
```diff
@@ -415,6 +426,7 @@
                 vm.recordEditModel.rows[rowIndex][column.name] = tuple.displayname.value;
             }, function noDataSelected() {
                 // do nothing
+                // TODO: need catchall here
```
hongsudt: user closed modal and we don't want to do anything. Remove the rejectCallback() and no catch is needed either.

recordedit/form.controller.js
```diff
@@ -433,6 +445,7 @@
         }
 
         function createRecord(column) {
+            // TODO: wrapper function with catch here
```
hongsudt: Remove the check for appURL. Have appLink function in common check for `undefined`

recordedit/recordEdit.app.js
```diff
@@ -233,6 +234,7 @@
                             $log.warn(response);
                             throw response;
```
hongsudt: remove rejectCallback() and let catch delegate.

recordedit/recordEdit.app.js
hongsudt: This is unnecessary with $exceptionHandler

recordedit/recordEdit.app.js
```diff
@@ -309,6 +311,7 @@
                 $log.warn(response);
                 throw response;
             }).catch(function genericCatch(exception) {
+                // TODO: Should this be factored out so that this logic is in the catchALL() or do we think this is specific to Recordedit?
```
hongsudt: This should just be the catchAll()

jrchudy:
@chiragsanghvi and @RFSH, I finished making the error handling refactor changes to the `recordedit` app. This is ready for review. 

Should the [catchAll](https://github.com/informatics-isi-edu/chaise/blob/master/common/errors.js#L120-L129) be officially removed from the code now?

recordedit/recordEdit.app.js
```diff
@@ -50,277 +50,264 @@
...
+                            }
+                        }
 
-                        notAuthorizedError.code = errorNames.unauthorized;
+                        $rootScope.displayReady = true;
+                        $log.info('Model: ', recordEditModel);
+                        // Keep a copy of the initial rows data so that we can see if user has made any changes later
+                        recordEditModel.oldRows = angular.copy(recordEditModel.rows);
+                    }, function error(response) {
+                        $log.warn(response);
```
RFSH: Do we need to log the error here? The exceptionHandler is doing this too.

recordedit/recordEdit.app.js
```diff
@@ -50,277 +50,264 @@
...
-                } else {
-                    ErrorService.errorPopup(exception.message, exception.code, "home page");
-                }
-            });
-        } catch (exception) {
-            ErrorService.errorPopup(exception.message, exception.code, "home page");
-        }
+            }
+        }, function error(response) {
+            $log.warn(response);
```
RFSH: the same comment.

RFSH:
@jrchudy If we're not calling that function anywhere in the code (which seems to be the case), we should remove it.
Also I think we should update the wiki to reflect the changes regarding not using `catcher` function, and using `window.onerror`.

PR #1026, opened 2017-03-22
Recordset Errors

jrchudy:
Opening this pull request with my comments for `recordset`.

### DIFF
--- a/common/errorDialog.controller.js
+++ b/common/errorDialog.controller.js
@@ -1,3 +1,4 @@
+
 (function() {
     'use strict';
 
@@ -6,10 +7,21 @@
     .controller('ErrorDialogController', ['$uibModalInstance', 'params', function ErrorDialogController($uibModalInstance, params) {
         var vm = this;
         vm.params = params;
+        // TODO: Wrap ok function in catcher fn 
         vm.ok = ok;
+
+        vm.details = false;
+        vm.showDetails = showDetails;
+        vm.linkText = "Show Details";
+        
+        function showDetails() {
+            vm.details = !vm.details;
+            if(vm.details) vm.linkText = "Hide Details";
+            else vm.linkText = "Show Details";
+        };
 
         function ok() {
             $uibModalInstance.close();
-        }
+        };
     }]);
 })();
--- a/common/errorDialog.controller.js
+++ b/common/errorDialog.controller.js
@@ -7,7 +7,6 @@
     .controller('ErrorDialogController', ['$uibModalInstance', 'params', function ErrorDialogController($uibModalInstance, params) {
         var vm = this;
         vm.params = params;
-        // TODO: Wrap ok function in catcher fn 
         vm.ok = ok;
 
         vm.details = false;
--- a/common/errors.js
+++ b/common/errors.js
@@ -88,7 +88,7 @@
                 // A more useful general message for 412 Precondition Failed
                 AlertsService.addAlert({type: 'warning', message: messageMap.generalPreconditionFailed});
             } else {
-                errorPopup("Oops! something went wrong.", 0, "Home Page", $window.location.origin, "The page encountered an error. See the Javascript console for technical details.");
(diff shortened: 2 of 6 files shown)


recordset/recordset.js
```diff
@@ -79,10 +79,12 @@
         $scope.navbarBrandImage = (chaiseConfig['navbarBrandImage'] !== undefined? chaiseConfig.navbarBrandImage : "");
         $scope.navbarBrandText = (chaiseConfig['navbarBrandText'] !== undefined? chaiseConfig.navbarBrandText : "Chaise");
 
+        // TODO: Session.login should already be wrapped in a try catch, maybe don't need to wrap this function with the wrapper (Utils.catchWrapper(function() {...}))
         $scope.login = function() {
```
RFSH: I don't think `login` and `logout` functions will even be needed here. Isn't the navbar responsible for these functionalities?

recordset/recordset.js
```diff
@@ -95,11 +97,13 @@
             $rootScope.location = $window.location.href;
         });
 
+        // TODO: We might not need to wrap functions if $on("error", ...) will catch errors thrown from one time function events
         $scope.$on('error', function(event, exception) {
```
RFSH: Does other apps need this event handler? It's only available in this app.

recordset/recordset.js
jrchudy: I think you are correct.

recordset/recordset.js
jrchudy: I noticed that in `ellipses.js` we are `emitting` errors when they occur and then calling the catchAll function. It's becoming more apparent to me that this is most likely unnecessary.

recordset/recordset.js
```diff
@@ -219,6 +224,7 @@
                 $log.warn(exception);
```
RFSH: I think the whole catch should be replaced by `catchAll` (assuming that `catchAll` will eventually pop-up the error if it cannot handle it)

chiragsanghvi:
@jrchudy @howdyjessie I have made the changes for global error handling as well as recordset app.

https://dev.gpcrconsortium.org/~chirag/chaise/recordset/#1/experiments:experiment

common/errorDialog.controller.js
```diff
@@ -6,10 +7,21 @@
     .controller('ErrorDialogController', ['$uibModalInstance', 'params', function ErrorDialogController($uibModalInstance, params) {
         var vm = this;
         vm.params = params;
+        // TODO: Wrap ok function in catcher fn
```
RFSH: @chiragsanghvi Is this comment still relevant?

common/errors.js
```diff
@@ -44,6 +45,7 @@
                     $window.location.replace(redirectLink);
                 }
             });
+            // TODO: Add .catch for catchAll
```
RFSH: @chiragsanghvi Same comment about here.

chiragsanghvi: Removed the comment

common/errorDialog.controller.js
chiragsanghvi: Removed the comment

RFSH:
Besides the two comments that I just made, I think the code looks good to merge. 

This is how the output of new catchAll looks like. Mentioning @hongsudt for comments about UI.
screen shot 2017-03-28 at 6 10 39 pm

hongsudt:
@chiragsanghvi We should be displaying 
Error: Terminal Error 
An unexpected error has occurred. Please report this problem to your system administrators.
Click OK to return to the Home Page, or Show Details to see technical details associated with this error. 

There should be a terminal error object that contains this information, correct?

chiragsanghvi:
@hongsudt This is how it will look for terminal errors. Other HTTP Error handling we will add later.

screen shot 2017-03-28 at 11 11 42 pm

jrchudy:
The changes look good to me. The error is detailed enough and provides the information that we had talked about for stack traces.

Issue #1034, opened 2017-03-29
copy should not copy generated columns


hongsudt:
These columns should be passed to the default list, so the default can be applied. 

* go to an existing experiment:  https://dev.gpcrconsortium.org/chaise/record/#1/experiments:experiment/id=40
* click copy
* when the new record is submitted, it seems like the values of generated columns (e.g. number of samples and number of images) also get updated. This is misleading as the new experiment shouldn't have any of these values. These generated columns should be using the default values.. 

Note: the rest of the columns will be overridden by the trigger, but it is still misleading to show the existing value. I can write a trigger to always initialize number of samples to 0, but this is probably the wrong approach. 

If it is easier, maybe we shouldn't be displaying the existing values of the generated column in the copy case.. We should show "automatically generated" instead. @karlcz @robes what do you think?

robes:
Make sense to me

PR #1042, opened 2017-04-05
Copy not generated columns

jrchudy:
This PR addresses issue #1034. When a record is copied, the generated columns are no longer copied and instead it shows "Automatically generated" in the input field.

I adjusted the test cases to account for this scenario and added a column to the table for the copy tests that is generated.

### DIFF
--- a/recordedit/recordEdit.app.js
+++ b/recordedit/recordEdit.app.js
@@ -177,9 +177,11 @@
                                     column = $rootScope.reference.columns[i];
 
                                     // If input is disabled, there's no need to transform the column value.
-                                    // if copy, don't populate the field
-                                    if (column.getInputDisabled(context.appContext) && !context.queryParams.copy) {
-                                        recordEditModel.rows[j][column.name] = values[i];
+                                    if (column.getInputDisabled(context.appContext)) {
+                                        // if not copy, populate the field without transforming it
+                                        if (context.mode != context.modes.COPY) {
+                                            recordEditModel.rows[j][column.name] = values[i];
+                                        }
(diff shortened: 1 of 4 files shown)


recordedit/recordEdit.app.js
```diff
@@ -219,7 +219,9 @@
                                             break;
                                     }
 
-                                    if (!context.queryParams.copy || !column.getInputDisabled(context.appContext)) {
+                                    // line 181 checks for input disabled (generated) and not copy
+                                    // no need to check for copy here because that case above guards against the negative case for copy
+                                    if (!column.getInputDisabled(context.appContext)) {
```
RFSH: While this would work; it's a bit difficult to follow the code. You can change it to:
```javascript

// If input is disabled, there's no need to transform the column value.
if (column.getInputDisabled(context.appContext)) {

  // in copy mode, don't copy the disabled 
  if (context.mode != context.modes.COPY) {
    recordEditModel.rows[j][column.name] = values[i];
  }

 continue;
}

// Transform column values for use in view model
..

// this last if statement is not needed anymore
recordEditModel.rows[j][column.name] = value;
```
You can also break the first `if` statement into two different statements `if` you want.

```javascript
// Don't copy the disabled value
if (context.mode == context.modes.COPY && column.getInputDisabled(context.appContext)) {
  continue;
}

// Don't transform the disabled input
if (column.getInputDisabled(context.appContext)) {
  recordEditModel.rows[j][column.name] = values[i];
  continue;
}
```
I'm not sure which one is more readable.

Issue #1044, opened 2017-04-06
Global exception handler does not catch all exceptions


jrchudy:
Exceptions that are thrown in the run block of our apps are not being caught properly. If you follow [this link](https://dev.isrd.isi.edu/chaise/record/#1/isa:dataset), you can see that the page errors out in an unexpected way. This link should produce an error stating that there is no filter defined. What is happening is that an exception is being thrown by the function `DataUtils.verify()`. It appears that the angular global exception handler, `$exceptionHandler`, is not being set properly.

It looks like `$exceptionHandler` needs to be configured as part of a config step to make sure it is defined the way we want before the `run` block runs.

PR #1046, opened 2017-04-06
Added a global error handler to fix issue #1044

chiragsanghvi:
This PR fixes issue #1044. I have added a window error handler that will catch all unhandled exceptions.

https://dev.isrd.isi.edu/~chirag/chaise/record/#1/legacy:dataset

It is not dependent on Angular for rendering the error.


common/errors.js
```diff
@@ -136,12 +136,48 @@
...
+window.onerror = function() {
+    var error = arguments[4];
+    error.stack = [
+        arguments[1],
+        arguments[2],
+        arguments[3]
+    ].join(':');
+    
+    var redirectLink = (chaiseConfig.dataBrowser ? chaiseConfig.dataBrowser : window.location.origin);
+    document.body.innerHTML = '<div modal-render="true" tabindex="-1" role="dialog" class="modal fade in" index="0" animate="animate" modal-animation="true" style="z-index: 1050; display: block;">'
```
jrchudy: It would be better to have this html defined in a template in `common/templates/`. Then you can import it using `$templateCache`.

`var template = $templateCache.get('nameOfTemplate.html');`

jrchudy: I guess this is wrong though. For this to work, the template needs to be defined inside a script tag as part of the outer html for that app.

chiragsanghvi: I didn't define it as a template, as there would be dependency in parsing it and replacing the messages.

jrchudy:
@chiragsanghvi, the tests for this branch all pass. They just stopped running because of the max build time and the ones that didn't finish running were search tests. I'm going to merge this branch into master.

Issue #1028, opened 2017-03-22
recordedit submit for cpm_file


ankit081190:
while using recordedit to submit data for cpm_file, it says success but shows a popup error saying no filter found and redirects to the main search page.

Despite saying success, the values are not being reflected in the database

URL: https://dev.gpcrconsortium.org/chaise/recordedit/#1/assets:cpm_file/id=111@sort(id)

jrchudy:
This [issue](https://github.com/informatics-isi-edu/ermrest/issues/142) was submitted to ermrest detailing that the cause might be in ermrest. After some investigation, it was determined that this was happening because of row-level security and that data wasn't allowed to be changed on the corresponding table. 

If multiple rows are submitted to be updated and 1 fails, it ignores the failed rows and returns only those that succeed. When `ermrestJS` sees a 200 response for a `PUT` request and there is no data returned, this is the case. We need to catch this case and have it throw a `403 Unauthorized Exception`.

jrchudy:
This issue has been resolved. If the user submits 1 row or multiple rows but 0 rows are updated, the user is presented with a 403 forbidden error.

If they submit multiple rows and only a subset are updated, the user is presented with 2 tables including the rows updated and the rows not updated.

PR #1054, opened 2017-04-11
PUT returns 200 with partial data

jrchudy:
This PR addresses issue #1028. When the user submits a 'PUT' request via `Reference.update` and it returns a smaller set of data than what was submitted, `chaise` will properly display that information to the user.

For this to work, you have to have the proper permissions. [This link](https://dev.gpcrconsortium.org/~jchudy/chaise/recordedit/#1/assets:cpm@sort(last_modified_timestamp::desc::,id)?limit=25) is a link to my folder on dev.gpcr. 

To test this, follow the link above and submit the data. You should be able to change entities with `Site name` as `USC`. The 2 entities associated with `iHuman` shouldn't be editable.

NOTE: The following requires the changes to `ermrestJS` as well. The above does not rely on changes to `ermrestJS`. 
You can then follow [this link](https://dev.gpcrconsortium.org/~jchudy/chaise/recordedit/#1/assets:cpm/id=1) and try submitting the data. You should be presented with an error dialog that states that editing records for table cpm is not allowed. There is no information returned in the request that states as to why editing is not allowed, so we can't present a more meaningful error message.

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -581,6 +581,54 @@
         }
 
         /**
+         * @param {ERMrest.Tuple[]} tuples - array of tuples
+         * @param {ERMrest.ReferenceColumn[]} columns - array of column names
+         * @return {Object[]} array of row value arrays [{isHTML: boolean, value: v}, ...]
+         */
+        function getRowValuesFromTupleData(tuples, columns) {
+            var rows = [];
+            for (var i = 0; i < tuples.length; i++) {
+                var tuple = tuples[i],
+                    row = [];
+
+                for (var j = 0; j < columns.length; j++) {
+                    var value,
+                        column = columns[j];
+
+                    if (column.isPseudo) {
+                        var keyColumns;
+
+                        if (column.key) {
+                            keyColumns = column.key.colset.columns;
+                        } else if (column.foreignKey) {
+                            keyColumns =  column.foreignKey.colset.columns;
+                        }
+
+                        for (var k = 0; k < keyColumns.length; k++) {
+                            var referenceColumn = keyColumns[k];
+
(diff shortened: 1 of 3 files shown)


common/utils.js
```diff
@@ -581,6 +581,54 @@
         }
 
         /**
+         * @param {ERMrest.Tuple[]} tuples - array of tuples
+         * @param {ERMrest.ReferenceColumn[]} columns - array of column names
+         * @return {Object[]} array of row value arrays [{isHTML: boolean, value: v}, ...]
+         */
+        function getRowValuesFromTupleData(tuples, columns) {
```
RFSH: I think there is a typo here. It should be "Raw" not "Row".

RFSH:
The code looks fine. I only have a UX related question. In the result page, shouldn't we present some "back" or "ok" buttons?

common/utils.js
jrchudy: I intentionally named it `Row` although technically they are the _raw_ values too.

jrchudy:
This page that shows up is actually still the `recordedit` app. Providing a back link would be odd because it would conceivably be a back link to the set of edit forms. More appropriately, there should be a link back to the originating `recordset` page that spawned this multi-edit mode. This link is present in the page title and behaves the same way that a "back" or "ok" button would behave (navigating to the recordset page for that table).

RFSH:
@jrchudy That makes sense.

common/utils.js
RFSH: You can change it to "raw" or leave it as it is. Just add a comment that emphasis on the fact that it returns _raw_ values.

Issue #1039, opened 2017-04-04
Timestamp entry is not initialized properly.. AM/PM toggle doesn't work


karlcz:
If you go to a recordedit page in record creation mode that has a timestamp entry, and start editing fields, you get validation errors until you type a valid date and time digit set.  But, the AM toggle is broken.

If you click 'Now' to intiailize the entry box and then edit the digits, the AM/PM toggle works correctly.

So, I think it must be failing to properly initialize the controller when you start with a blank record?

example page:

https://synapse-dev.isrd.isi.edu/chaise/recordedit/#1/Zebrafish:Behavior

displays the error:

```
toggleMeridiem@https://synapse-dev.isrd.isi.edu/chaise/recordedit/form.controller.js?v=c3d6045a599d2bd8479fe7b37ab4c294:631:1
anonymous/fn@https://synapse-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84 line 14432 > Function:2:541
expensiveCheckFn@https://synapse-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84:15485:18
ngEventHandler
```

even more detailed description of how to produce:
1. load the page
2. use mouse to enter the "Room Date" date field.
3. type 2017-01-02
4. use mouse to enter the "Room Date" time field.
5. type 120101 (it fills colons for you)
6. use mouse to leave the time field, e.g. clicked page background
7. click "AM" button on the "Room Date" field
8. get an "Error: Terminal Error" alert with merdiem is undefined

i'm doing this in a firefox private browser session w/ clean cache

jrchudy:
This issue has been fixed. Reassigning to @karlcz for verification.

karlcz:
ok I see that the terminal error is gone

PR #1057, opened 2017-04-13
Toggling AM/PM

jrchudy:
This PR fixes issue #1039. There was a condition for when the page was initialized to set a default value for some fields (if it was available) in the `entry/create` context. If that default wasn't available (it's a function) or it's `null`, `timestamp` and `timestamptz` subfields were initialized to `null` with a default meridiem of `AM`. 

This was then being overwritten to the default value. That should have been nested in an `else` case.

### DIFF
--- a/recordedit/recordEdit.app.js
+++ b/recordedit/recordEdit.app.js
@@ -275,8 +275,9 @@
                                             meridiem: 'AM'
                                         };
                                     }
+                                } else {
+                                    recordEditModel.rows[0][column.name] = column.default;
                                 }
-                                recordEditModel.rows[0][column.name] = column.default;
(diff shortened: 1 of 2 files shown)


RFSH:
The change looks good and fixes the issue. I just have a question about [the test case that you changed](https://github.com/informatics-isi-edu/chaise/blob/54350aeea685f093f396d37e57b4d098646c78fa/test/e2e/specs/recordedit/helpers.js#L552). 
Is it possible to break it into different test cases? I know that time limit is a concern for not breaking it into different test cases. But looking at the test case, it's hard to follow a 100+ lines unit test and understand what exactly is trying to test.

jrchudy:
After reviewing the structure of the e2e tests, it isn't possible to factor this section into smaller pieces given the way protractor scopes variables.

PR #1058, opened 2017-04-13
Heuristics for when to show related tables changed

jrchudy:
This PR addresses issue #1055. The Show/Hide empty related tables toggle is now based on whether the user can modify at least one of the related tables. If they can, it shows all empty related tables by default.


chiragsanghvi:
Is there a place where we can test the above behavior ?

jrchudy:
I don't know of any deployments that have tables with related tables that a user cannot edit. Besides that, everything should still remain the same.

jrchudy:
Link available for now (if the catalog gets deleted it won't work):

https://dev.isrd.isi.edu/~jchudy/chaise/record/#20549/defaults:defaults-table/id=1

PR #1069, opened 2017-04-19
401 handling

chiragsanghvi:
This PR adds an explicit interceptor for handling 401 errors in `authen.js`.

Whenever a 401 error is encountered the handler is called. The handler renders the login window and on successful login, fetches the new session and sets it in the `authen` module.

It also adds logic to subscribe to change in session. This is used in navbar to update the once the user has logged in.

Can be tested here https://dev.gpcrconsortium.org/~chirag/chaise/recordedit/#1/experiments:experiment

### DIFF
--- a/common/authen.js
+++ b/common/authen.js
@@ -11,6 +11,8 @@
         // Private variable to store current session object
         var _session = null;
 
+        var _changeCbs = [];
+
         return {
 
             getSession: function() {
@@ -28,6 +30,16 @@
 
             getSessionValue: function() {
                 return _session;
+            },
+
+            subscribeOnChange: function(fn) {
+                if (typeof fn  == 'function') _changeCbs.push(fn);
+            },
+
+            executeListeners: function() {
+                _changeCbs.forEach(function(cb) {
+                   cb(); 
+                });
             },
 
             login: function (referrer) {
@@ -221,6 +233,9 @@
                     // and it can continue firing other queued calls
                     Session.getSession().then(function(_session) {
                         defer.resolve();
+
+                        Session.executeListeners();
+
                     }, function(exception) {
                         throw exception;
                     });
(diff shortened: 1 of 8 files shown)


common/authen.js
```diff
@@ -8,16 +8,40 @@
...
+        // Private variable to store current session object
+        var _session = null;
+
+        var _changeCbs = [];
+
         return {
 
             getSession: function() {
                 return $http.get(serviceURL + "/authn/session").then(function(response) {
+                    _session = response.data;
```
jrchudy: Shouldn't you call `setSessionValue()` here? This seems like the only place that function would ever be called. If you don't think it should be changed, then maybe `setSessionValue()` doesn't need to be defined.

common/authen.js
```diff
@@ -8,16 +8,40 @@
...
                     return response.data;
                 }, function(response) {
                     return $q.reject(response);
                 });
             },
 
+            setSessionValue: function(session) {
```
jrchudy: See comment above.

common/authen.js
chiragsanghvi: I have removed the` setSessionValue` function. In addition I have made some more changes to the way we fetch session in each app. 

I have introduced a pub sub mechanism, where the recordedit, record and recordset apps listen to session change event once in the beginning and then stop listening it to avoid fetching it twice by authen and app.

chiragsanghvi:
@jrchudy Yes thats right. Each app's run block subscribes to session changes. But the app doesn't calls `getSession()`. It is called by the navbar.

jrchudy:
Cool. I was testing it out from a bunch of different login scenarios and it behaves properly. 

It's just a different way to think about the problem in a somewhat less linear way. By less linear I guess I mean that you have to understand how each of these components communicate together to understand the whole flow. Merge when you are ready.

Issue #1085, opened 2017-04-25
Search box not working properly


svoinea:
On `cirm-dev.isrd.isi.edu`, in the `Scan` search view, when I enter text in the search box, the following text is displayed in the sidebar: 
```
“{{FacetsData.searchFilter}}” found in {{sideBar.getEnabledFilters()}} attribute(s):
```
In the browser console, the following error is displayed:
```
Error: $scope.FacetsData.box[facet.table][facet.name] is undefined
showFacetMatch@https://cirm-dev.isrd.isi.edu/chaise/scripts/controller/ermrestSideBarController.js?v=47bd899826a9293d721b98246defef6a:246:1
anonymous/fn@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84 line 14432 > Function:2:650
$RootScopeProvider/this.$get</Scope.prototype.$digest@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84:17064:34
$RootScopeProvider/this.$get</Scope.prototype.$apply@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84:17337:13
NgModelController</this.$$debounceViewValueCommit@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84:26912:7
NgModelController</this.$setViewValue@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84:26884:7
baseInputType/listener@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/angular.js?v=ea03e809effc2b87cc7bd368e1740e84:23129:7
m.event.dispatch@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/jquery-latest.min.js?v=8101d596b2b8fa35fe3a634ea342d7c3:3:8384
m.event.add/r.handle@https://cirm-dev.isrd.isi.edu/chaise/scripts/vendor/jquery-latest.min.js?v=8101d596b2b8fa35fe3a634ea342d7c3:3:5122
```

jrchudy:
[...]

PR #1086, opened 2017-04-25
Search box not working properly

svoinea:
The pull request is related to the issue https://github.com/informatics-isi-edu/chaise/issues/1085

I have enclosed the body of the `showFacetMatch()` function from the `ermrestSideBarController.js` file in a `try/catch` statement and bypass thus the errors.

### DIFF
--- a/scripts/controller/ermrestSideBarController.js
+++ b/scripts/controller/ermrestSideBarController.js
@@ -242,19 +242,25 @@ ermSideBarController.controller('SideBarCtrl', ['$scope', '$filter', '$timeout',
 	};
 
 	this.showFacetMatch = function showFacetMatch(facet) {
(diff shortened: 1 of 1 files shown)


jrchudy:
@serban, while this fixes your specific case, this will continue to be a problem in `search` app. The error was caused because of the angular exception handler function trying to access `Ermrest` when it's not available. 

To better address this issue, we should change [this line](https://github.com/informatics-isi-edu/chaise/blob/master/common/errors.js#L125) to include this check:
```
(window.location.pathname.indexOf('/search/') != -1
```

jrchudy:
@serban, test it out at this link:
https://cirm-dev.isrd.isi.edu/~jchudy/chaise/search/#1/Microscopy:Scan

If that fixes your problem, we should go with the changes I made to the `handleException` function to ignore running in the search app.

Issue #1088, opened 2017-04-26
Add a global configuration option for the page size of the displayed results in the search view


svoinea:

svoinea:
Added the `pageSize` option in the `chaise-config.js` file.
If present, its value is an integer representing the number of records to be displayed on a page in the search view.
The default is `25`.

PR #1089, opened 2017-04-26
Added the "pageSize" option in the global chaise configuration.

svoinea:
If present, its value is an integer representing the number of records to be displayed on a page in the search view.
The default is 25.
The pull request is related to the issue https://github.com/informatics-isi-edu/chaise/issues/1088


svoinea:
The parameter name is now `searchPageSize`.

jrchudy:
Your changes look good @svoinea. Could you update the [configuration doc](https://github.com/informatics-isi-edu/chaise/blob/master/doc/configuration.md) to reflect this new property?

svoinea:
Updated the documentation for the new global parameter.

PR #1091, opened 2017-04-27
Support search on facets NULL values

svoinea:
Updated the code to support search on facets NULL values.


svoinea:
So far, the `NULL` values were excluded from the facets sidebar search page.

As the `ermrest /attributegroup` request is returning also the `NULL` distinct value, it can be added to the facets sidebar search page.

The ermrest predicate for `NULL` values is `col::null::`.

The label for the checkbox representing the `IS NULL` predicate is an empty string and it is displayed at the end of the facet's values, as the `ermrest sort()` function returns the `NULL` values at the end.

**Known limitation**: a filter can not be applied any more to a text column having the `null` string value (predicate `col=null`), as it translates now in the `col::null::` predicate.

Issue #869, opened 2016-12-15
recordedit: add a spinner while the page is waiting to be redirected


hongsudt:
We should do this for create, edit and delete.

Issue #550, opened 2016-08-24
Refactor spinner gif


howdyjessie:
We use the same code to display the spinner gif in our apps. It should be moved to the `common/` folder in the form of a new spinner directive.

``` html
<div ng-show="pageInfo.loading" id="spinner">
    <img src="spinner.gif" style="display: block; margin: 0 auto; width: 32px; height: 32px; opacity: 0.8;"/>
    <div style="margin-top: 15px;">Loading...</div>
</div>
```

PR #1092, opened 2017-04-27
Loading spinner refactor

jrchudy:
This PR addresses issues #550 and #869. The spinner now shows up over the `recordedit` form when the user has submitted the data for an update or create.

Everything seems to be working properly except in `recordedit`. When `recordedit` is loading, the spinner does not show up. I think it's related to the template not being loaded until after the DOM has started to load. Directives don't load until after the run block, so maybe this is why this is happening.

But contrary to this, `recordset` does properly show the spinner before the app loads. Maybe it's because `recordset` injects a template to be loaded with `index.html` allowing more of angular to load in the background in the mean time.


jrchudy:
I am going to investigate why the spinner is not showing up for `recordedit` some more tomorrow. Besides that, everything else appears to be working as expected.

jrchudy:
It was pretty trivial why the spinner was not working while loading the `recordedit` app. This has been fixed.

Issue #1096, opened 2017-05-02
RE View Model shouldn't be changed during submission


jrchudy:
Currently the submission function in `recordedit` modifies the values in the view model for submission in place before setting them in the submission object. We should not adjust the view model before submitting the data and instead use a new object.

PR #1097, opened 2017-05-02
Re view model

jrchudy:
This PR addresses issue #1096. The view model is no longer changed while transforming the view values for submission.

Also fixes a change to the makefile that I accidentally pushed to master.


recordedit/form.controller.js
```diff
@@ -114,12 +115,6 @@
...
                 var col = $rootScope.reference.columns[i];
 
-                // If this column is disabled, its value is already in the format
-                // needed for submission.
-                if (col.getInputDisabled(context.appContext)) {
-                    continue;
-                }
-
                 var rowVal = row[col.name];
                 if (rowVal) {
```
RFSH: This is not related to what you changed in this PR. But the [if statement in `default`](https://github.com/informatics-isi-edu/chaise/blob/261a5780b623466ac1163264d4c28271e0b46018/recordedit/form.controller.js#L136) will never be true, since we have this if statement here.

jrchudy: That is correct. I should probably remove that line.

Issue #1094, opened 2017-05-01
Confusing logout flow


karlcz:
When I use the "Logout" button on the current Chaise recordset app, using Globus Auth provider in webauthn, while logged in via a Globus ID account, I am presented with a page from auth.globus.org like this:

    Logged out of Globus Auth

    You have successfully logged out.

    If you are using a shared computer, please ensure you have also logged out of any identity providers used during your session, and shut down this browser to remove any session cookies.

    To make sure your account is secure, ou may need to log out of the following identity providers:

      - Globus ID ---  [Log out](www.globusid.org/logout)

    Continue to (https://synapse-dev.isrd.isi.edu/chaise/logout)

It's this last link that I think is wrong.  As far as I can tell, I'm already logged out of webauthn by this point, so I'm not sure why I need to go to another logout app.  This page just offers a link that invokes chaiseApp() which seems to then redirect to `/chaise/search`.

1. It seems to me we should just offer the "continue to site" option directly as the link in the globus auth logout page. Why go through two interstitials showing essentially the same verbiage about cookies and sessions?
2. I don't think the chaiseApp() thing should redirect to `/chaise/search` anyway.  Shouldn't it go to the site homepage, or some deployment-configurable location?

I'm not sure if this is solely a Chaise UX issue or relates to some confusion or misuse of the webauthn logout API?  Is there some technical reason we want/need this extra layer of interstitial page?  @ljpearlman @robes @hongsudt

jrchudy:
The plan is to change the referrer url sent with the logout request to the link that is presented on the extra logout page. This way there's only one logout page that will then redirect back to the `databrowser` chaise config option. If this option is not set, it will default to root.

jrchudy:
The changes to the logout flow have been merged into master. Reassigning to @karlcz for review.

karlcz:
This looks better. The only thing I can imagine to improve it would be to also have a config option to set a title for the post-logout link so it doesn't just show the user a plain URL as the anchor text.

PR #1099, opened 2017-05-04
Removed unnecessary extra logout page

jrchudy:
This PR addresses issue #1094. The extra logout page has been removed so that when the user logs out, they are presented with an appropriate link to go back to the defined chaiseConfig property `logoutURL`. If it is undefined, it will go to the root page.


jrchudy:
Good catch. I will update the config doc.

PR #1103, opened 2017-05-10
Repo cleanup

jrchudy:
This PR removes unnecessary files in the repo as well as updating some more of the chaise configs for e2e tests.


RFSH:
Should we replicate the documents that you removed (`karma.md` and `protractor.md`) to wiki pages? We have a wiki that talks about protractor, but `protractor.md` has some extra information.

jrchudy:
I will look through both files (the doc I removed and the existing wiki page) later this afternoon and add the information that's missing from the wiki page. The karma document is unnecessary given that we don't have any unit tests that run in chaise.

Issue #1106, opened 2017-05-11
Change .isPseudo handling to consider foreign key and asset columns


chiragsanghvi:
Currently, the code assumes that if a column has `.isPseudo` flag on, it is a foreign key and things break in recordedit and record app for asset columns. 

After @RFSH made changes to ermrestjs for the [asset annotation](https://github.com/informatics-isi-edu/ermrestjs/pull/400) refactoring `ReferenceColumn` implementation to have a base table that other `pusedo-column` classes (`KeyPseudoColumn`, `ForeignKeyPseudoColumn`, and `AssetPseudoColumn`) will extend.

PR #1107, opened 2017-05-11
Fixed issue #1106

chiragsanghvi:
This PR fixes issue #1106.


recordedit/form.controller.js
```diff
@@ -505,7 +505,7 @@
             var transformedRow = transformRowValues(modelRow);
             columns.forEach(function (column) {
                 // If the column is a pseudo column, it needs to get the originating columns name for data submission
-                if (column.isPseudo) {
+                if (column.isPseudo && column.isForeignKey) {
```
jrchudy: This case is meant for only foreign keys. Checking for column.isPseudo is redundant. Could you remove it?

Issue #1102, opened 2017-05-10
Investigate alternatives to angular.copy


jrchudy:
Angular.copy is doing a deep copy of `tuple.values` in the `recordedit` app. Should this just be a shallow copy? Should we write our own copy function for this?

hongsudt:
This deep copy cause significant performance degradation on a table with many fks. Once this is done, please check the whether we still have an alert popup to stop running the script on rbk (catalog 2, table: RNASeq:Sample_Settings).

jrchudy:
Re-assigning this issue to you @hongsudt to verify the performance is better for `recordedit`.

hongsudt:
@svoinea can you please test this and see whether you still have problem on your browser on dev?

svoinea:
I have tested it on `rbk-dev` and the delay is not present any more.

PR #1112, opened 2017-05-12
Remove deep copy in recordedit

jrchudy:
This PR addresses issue #1102. In `recordedit` we were creating a deep copy of the whole `page.tuples` array. This would recursively copy every single object in each `Tuple` in the array. This is unnecessary given that we are only ever modifying the data associated with the `Tuple`. I added a function to the `Tuple` class in ermrestJS to handle this so that the `Tuple`'s prototype can be used.

**NOTE: This PR cannot be merged until [this](https://github.com/informatics-isi-edu/ermrestjs/pull/405) PR is merged in `ermrestJS`.

### DIFF
--- a/recordedit/recordEdit.app.js
+++ b/recordedit/recordEdit.app.js
@@ -175,7 +175,7 @@
 
                                 // We don't want to mutate the actual tuples associated with the page returned from `reference.read`
                                 // The submission data is copied back to the tuples object before submitted in the PUT request
-                                var shallowTuple = tuple.deepCopyData();
+                                var shallowTuple = tuple.copy();
                                 $rootScope.tuples.push(shallowTuple);
 
                                 for (var i = 0; i < $rootScope.reference.columns.length; i++) {
--- a/recordedit/recordEdit.app.js
+++ b/recordedit/recordEdit.app.js
@@ -161,9 +161,7 @@
                             var column, value;
 
                             // $rootScope.tuples is used for keeping track of changes in the tuple data before it is submitted for update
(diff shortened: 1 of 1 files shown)


recordedit/recordEdit.app.js
```diff
@@ -173,7 +171,12 @@
                                 recordEditModel.submissionRows[j] = {};
 
                                 var tuple = page.tuples[j],
-                                values = tuple.values;
+                                    values = tuple.values;
+
+                                // We don't want to mutate the actual tuples associated with the page returned from `reference.read`
+                                // The submission data is copied back to the tuples object before submitted in the PUT request
+                                var shallowTuple = tuple.deepCopyData();
```
robes: after addressing the comments in the review of ermrestjs this line ought to read more like
```
var copiedTuple = tuple.copy();
```

recordedit/recordEdit.app.js
```diff
@@ -161,9 +161,7 @@
                             var column, value;
 
                             // $rootScope.tuples is used for keeping track of changes in the tuple data before it is submitted for update
-                            // We don't want to mutate the actual tuples associated with the page returned from `reference.read`
-                            // The submission data is copied back to the tuples object before submitted in the PUT request
-                            $rootScope.tuples = angular.copy(page.tuples);
+                            $rootScope.tuples = [];
```
robes: Side note: seems off that these things are stored directly in `$rootScope` but that doesn't need to be addressed by this PR.

Issue #1009, opened 2017-03-02
Refactor recordedit Testcases to not imitate ermrestjs logic


RFSH:
There are some test cases that try to imitate the ermrestjs logic. For instance [`getVisibleColumns`](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/utils/page.utils.js#L277) function is trying to figure out the visible columns using the annotation.

I don't think we should have these sort of logics in chaise test cases. Instead, we should just test for a fixed and known result (instead of applying a logic).

hongsudt:
@RFSH Can you identify specific test cases, so we can quickly discuss that with the test owners and have them removed?

RFSH:
All the functions that are under [`EditInputs`](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/utils/page.utils.js#L230), are trying to get the result based on some logic and that should not be the case for unit tests. These function are just used in recordedit test suit.

| link                                                                                                      | function                   |
|-----------------------------------------------------------------------------------------------------------|----------------------------|
| [1](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js#L41)   | getDisplayname             |
| [2](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js#L54)   | getVisibleColumns          |
| [3](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js#L123)  | getVisibleColumns          |
| [4](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js#L72)   | getColumnsWithComment      |
| [5](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js#L90)   | getColumnsWithRequired     |
| [6](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js)       | getDisabledColumns         |
| [7](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js#L155)  | getLongTextDataTypeColumns |
| [8](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/recordedit/helpers.js)       | getTextDataTypeColumns     |
[...]

PR #1122, opened 2017-05-19
Issue1009

chiragsanghvi:
This PR fixes issue #1009. It removes the logic that defers columns using its own logic. Now it refers the table parameters for columns.


.travis.yml
```diff
@@ -5,10 +5,6 @@ node_js:
 sudo: required
 dist: trusty
 
-branches:
-  only:
-    - master
```
jrchudy: Did you mean to remove this line?

Removing this will cause the build to run for every branch when code is pushed to it in git.

chiragsanghvi: Yes, I wanted to run the test-cases on travis. They pass on my machine. I will revert it back.

chiragsanghvi: Made the changes as requested

PR #1124, opened 2017-05-22
Fileupload

chiragsanghvi:
This PR adds file upload support to Chaise. It refers to the asset annotation defined [here](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2017-asset).

Currently the upload feature doesn't supports `sha` and `filename-ext-filter`.


common/templates/uploadProgress.modal.html
```diff
@@ -0,0 +1,77 @@
...
+	<button class="btn btn-default pull-right modal-close" type="button" ng-click="ctrl.cancel()">X</button>
+    <h3 class="modal-title">{{ctrl.title}}</h3>
+</div>
+<div class="modal-body">
+	<div class="row no-row-margin" ng-show="!ctrl.isCreateUploadJob && !ctrl.isFileExists && !ctrl.isUpload ">
+    	<div class="col-md-12 pad0-left pad0-right text-right">
+	    	<strong>Completed: {{ctrl.checksumCompleted}}/{{ctrl.noOfFiles}}</strong>
+    	</div>
+	    <div class="progress">
+	        <div class="progress-bar" role="progressbar" ng-style="{ 'width': ctrl.checksumProgress + '%' }" style="width: 0%;background-color: #8cacc7 !important;"></div>
```
jrchudy: I'm seeing `style="width: 0%;background-color: #8cacc7 !important;"` repeated quite a few times. This should be a css class instead.

common/upload.js
```diff
@@ -0,0 +1,533 @@
...
+            };
+
+            
+            /**
+             * @function
+             * @param {Object} col - Column Object
+             * @desc
+             * Creates an uploadFile obj to keep track of file and its upload.
+             */
+            var uploadFile = function(col, column, row) {
```
jrchudy: I'm curious what the reasoning is for creating this object in `chaise` instead of it being in `ermrestJS`?

While this works, this should either be it's own angular `factory` or be defined in `ermrestJS`.

recordedit/form.controller.js
```diff
@@ -141,58 +145,143 @@
...
-                // loop through model.submissionRows
-                for (var i = 0; i < model.submissionRows.length; i++) {
-                    var row = model.submissionRows[i];
-                    var data = $rootScope.tuples[i].data;
-                    // assign each value from the form to the data object on tuple
-                    for (var key in row) {
-                        data[key] = (row[key] === '' ? null : row[key]);
+            var submissionRowsCopy = [];
+
+            model.submissionRows.forEach(function(row) {
```
jrchudy: I think this was something that we needed to change. The previous code was modifying data in place. Thank you for addressing this.

recordedit/recordEdit.app.js
```diff
@@ -223,7 +224,11 @@
                                             value = (!isNaN(floatVal) ? floatVal : null);
                                             break;
                                         default:
-                                            value = values[i];
+                                            if (column.isAsset) {
```
jrchudy: For an asset column, `column.type.name` is not being set to something specific to assets? Is this because `ermrest` implements these as text fields in the database?

chiragsanghvi: Yes ermrest implements them as text fields.

common/templates/uploadProgress.modal.html
chiragsanghvi: I was planning to put it in ermrestjs, but as it sets some variables specific to upload progress parameters, that are shared I couldn't. In addition I couldn't make it a factory as it utilizes controller variables and updates them.

common/upload.js
chiragsanghvi: I was planning to put it in ermrestjs, but as it sets some variables specific to upload progress parameters, that are shared I couldn't. In addition I couldn't make it a factory as it utilizes controller variables and updates them.

PR #1140, opened 2017-06-03
added some logging for $scope.$apply

jrchudy:
I added some logging for the $scope.$apply errors. I will update this on monday with logging to a log folder on each deployment so we can later track how often this occurs and better establish a mechanism to handling it.


chiragsanghvi:
@jrchudy the try/catch wrapping needs to happen inside the `$timeout`. It won't catch those errors as they're asynchronous due to the $timeout.

Just move the try/catch inside the $timeout.

Issue #1128, opened 2017-05-24
default values are not displayed in recordedit


hongsudt:
check the difference between what's on [rbk production](https://www.rebuildingakidney.org/chaise/recordedit/#2/RNASeq:Sample_Settings) and [rbk dev](https://www.rebuildingakidney.org/chaise/recordedit/#2/RNASeq:Sample_Settings). It seems like recordedit no longer displays the default value.  

This bug is probably introduced during the recent changes to fix the timestamp issue? 

@jrchudy I forgot that the default value is used to mimic section header in recordedit. So for generated columns that have default values, we should display the default value (the way that is is used on rbk production above).

jrchudy:
The corresponding PR has been merged into master.

Reassigning this issue to @hongsudt to verify that it has been fixed.

hongsudt:
The defaults are working properly on rbk-dev.

PR #1144, opened 2017-06-06
Default values in recordedit show up properly if the input is disabled

jrchudy:
This PR addresses issue #1128. The default values show properly in `recordedit` inputs when those values are defined and not a function, i.e. those values are most likely primitives (strings are primitives in JS).

We can discuss the logic tomorrow and decide whether that should be reorganized to make more sense. I had to add a bunch of comments because I had to re-understood the reasoning behind what was there.

### DIFF
--- a/recordedit/recordEdit.app.js
+++ b/recordedit/recordEdit.app.js
@@ -224,8 +224,8 @@
                                             value = (!isNaN(floatVal) ? floatVal : null);
                                             break;
                                         default:
-                                            if (column.isAsset) {                                            
-                                                value = { url: values[i] || "" }; 
+                                            if (column.isAsset) {
+                                                value = { url: values[i] || "" };
                                             } else {
                                                 value = values[i];
                                             }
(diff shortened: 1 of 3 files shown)


recordedit/recordEdit.app.js
```diff
@@ -265,37 +265,43 @@
...
-                                                date: null,
-                                                time: null,
-                                                meridiem: 'AM'
-                                            };
-                                        }
-                                    } else if (column.isAsset) {
+                            // if column.default == undefined, the second condition would be true so we need to check both if column.default is defined and it's not a function
+                            // only want to set values in the input fields so make sure it isn't a function
+                            // check the recordEditModel to make sure a value wasn't already set based on the prefill condition
+                            if (column.default !== undefined && typeof column.default !== "function" && !recordEditModel.rows[0][column.name]) {
```
RFSH: Can `recordEditModel.rows[0][column.name]` be `false`? (Which has a value but `!recordEditModel.rows[0][column.name]` will return `true`)

jrchudy: That check is being made because you could have come from an `Add link` for a related table. This would populate `recordEditModel.rows[x][column.name]` with a value representing a link to the entity that that related table was associated with. To the best of my knowledge, a foreign key relationship should never be represented by a `boolean` value. So I think it is safe to assume if it's already set, it won't be a boolean value.

Issue #1105, opened 2017-05-11
logging: pass windowID along with CID to ermrest calls


hongsudt:
Ideally we should supply three complementary parameters:
- cid: the current client application name
- wid: a window ID stable across page loads to allow tracking of chaise app flows within one window
- pid: a page ID that changes once per chaise app restart, to track requests associated with one app instance

PR #1158, opened 2017-06-13
added support for setting the wid and pid for each server request

jrchudy:
This PR addresses issue #1105.

`Record`, `Recordset`, and `Recordedit` now set these properties when making requests to the server. e2e test cases pass locally for me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -492,7 +492,7 @@ testrecordadd:
 #Rule to run recordset app tests
 .PHONY: testrecordset
 testrecordset:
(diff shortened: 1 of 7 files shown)


RFSH:
What's the difference between `wid` and `pid`? I thought `wid` is for window. That means if we have two tabs on the same window, they should have the same `wid`, and `pid` is per page (tab).

karlcz:
No, wid should be different for each tab.  We don't care about
difference between windows and tabs.  Separate app rendering
container, i.e. a place that a DOM is rendered.

The pid should change every time an app loads/reloads/restarts even in
the same window.  But wid should be stable across page loads i.e.  due
to changes to the browser window.location

A possibly tricky part is iframes.  Ideally we'd want them to inherit
the wid of their parent.  I'm less certain about whether they should
inherit pid or also get that from parent...

Issue #1161, opened 2017-06-13
Hide related tables


jrchudy:
The "Hide Empty Related Records" button should appear when all of the related tables are empty. Currently it shows as "Show All Related Records" when the related tables are all empty.

jrchudy:
The code that fixes this issue has been merged into master. Closing this issue.

PR #1164, opened 2017-06-13
Related tables show up properly when they are empty

jrchudy:
This PR fixes issue #1161.

The related tables were not showing up by default when the user could add/edit the entities in them if those tables were empty. This is fixed by moving the logic to the `run` block. How it was defined before, it was relying on `formController.canCreateRelated()` being called. When none of the related tables had entities in them, this function was never executed in the html and therefore never updated the state for the page.


jrchudy:
Oh yes. I meant to do that and then it must have slipped my mind when I was cleaning up the code last night.

Issue #1156, opened 2017-06-12
Default Values in Recordedit


jrchudy:
The Default values need to be more robustly tested in the e2e tests at the initialization of the `recordedit` app. Include tests for each column type:
- **not** generated
  - [X] boolean
  - [X] date
  - [X] timestamp
  - [X] text
  - [X] markdown
  - [X] integer
  - [X] float
  - [X] f. key
- generated (disabled)
  - [X] boolean
  - [X] date
  - [X] timestamp
  - [X] text
  - [X] markdown
  - [X] integer
  - [X] float
  - [X] f. key

karlcz:
foreign key is not a column type but an orthogonal choice to all of those column types...  also, please make any test for this general enough so that we can easily add more column types later, i.e. `json` and `jsonb` should be added as soon as we have that feature in chaise.

jrchudy:
The changes for this issue are merged into master. Support for json columns is still in progress so those tests will be included with the chanegs to support `json` and `jsonb`. This is being tracked by issue #1041.

PR #1168, opened 2017-06-15
Test default values in recordedit

jrchudy:
This PR addresses issue #1156. Default values are now being properly set on the `recordedit` page in create mode. There was a bug with timestamp columns that had a default value that was also a disabled field. It was showing `[ Object object ]` instead of the string representation.


jrchudy:
@RFSH I addressed your comments and moved that one test to another spec (edit presentation spec). The changes for that were very minimal. The existing edit presentation spec was getting the innerHTML for column name displays properly, so I just needed to update the schema and test expectations object.

The defaults are all working properly now too. Merging this PR.

Issue #1041, opened 2017-04-05
JSON column support in chaise


jrchudy:
Chaise applications need to properly support JSON columns. If a JSON column is present in a table definition and part of the visible columns list, the column needs to be submitted back to ermrest as a JSON object rather than a string representation of that object. This is the case for `recordedit`.

karlcz:
To be explicit, any single JSON value should be supported, e.g.:

- null
- true
- false
- 2.0
- "a string"
- ["an", "array"]
- {"an": "object"}

karlcz:
This overlaps with the ancient informatics-isi-edu/ermrestjs#146

karlcz:
Oops I guess I got confused and thought this task was further along than it is.  So maybe the new #1163 is a duplicate and should be closed?

jrchudy:
Yes @karlcz. We have yet to merge json column support into chaise.

nitishksahu:
I have made certain changes and have added the JSON column support in chaise, these changes are available at:
https://dev.isrd.isi.edu/~nitish/chaise/recordset/#1/public:orders

@karlcz @hongsudt @robes any suggestions?

robes:
worked for me

nitishksahu:
As @hongsudt suggested, 

- the conversion of JSON object to string value must be done in the ermrestjs rather than to handle it in chaise, so going forward I will be updating ermrestjs code to accommodate this change for column type "JSON", where ermrestjs will stringify JSON object and provide a string value to chaise to display. 
Ermrest right now does not provide the formatted values in the case of "edit record" and passes the raw values, for that case as well we need to add a condition for JSON column type and pass the formatted value instead of raw values.

After having some discussion on this with @jrchudy and @RFSH, 

- we understand this, that chaise(recordedit) is the only application that submits data (PUT/POST) to ermrestjs and does all the manipulation of data to appropriate type before submitting it to ermrestjs, so the JSON value from the recordedit must be parsed to JSON object before submitting it to ermrestjs.
<img width="523" alt="screen shot 2017-06-15 at 3 57 08 pm" src="https://user-images.githubusercontent.com/7412235/27205047-567e7538-51e3-11e7-85e4-a0b90875ac97.png">

robes:
Is there some special case here?

From what I recall, every data type works this way. That is,
- ermrestjs returns data in both formatted and raw forms (hence, for JSON they should be returned in string, and js object)
[...]

PR #1177, opened 2017-06-29
JSON column support in chaise #1041

nitishksahu:
This PR addresses the issue#1041.

Chaise applications needed to properly support JSON/JSONB columns.


test/e2e/utils/chaise.page.js
```diff
@@ -314,6 +314,16 @@ var recordEditPage = function() {
...
 
+    this.getJSONtextAreaForAcolumn = function(name, index) {
+       index = index || 0;
+       return browser.executeScript("return $('td.entity-value textarea[name=\"" + name + "\"]')[" + index + "];");
+   };
+   
+   this.getJSONBtextAreaForAcolumn = function(name, index) {
+       index = index || 0;
+       return browser.executeScript("return $('td.entity-value textarea[name=\"" + name + "\"]')[" + index + "];");
+   };
```
jrchudy: How are these 2 functions different?

test/e2e/utils/chaise.page.js
```diff
@@ -464,6 +474,13 @@ var recordEditPage = function() {
         return browser.executeScript("return $(arguments[0]).parents('div[ng-switch-when=\"timestamp\"]').siblings('.text-danger.ng-active').find('div[ng-message=\"" + type + "\"]')[0];", el);
     };
     
+    this.getJSONInputErrorMessage = function(el, type) {
+        return browser.executeScript("return $(arguments[0]).parents('div[ng-switch-when=\"json\"]').siblings('.text-danger.ng-active').find('div[ng-message=\"" + type + "\"]')[0];", el);
+    };
+    this.getJSONBInputErrorMessage = function(el, type) {
+        return browser.executeScript("return $(arguments[0]).parents('div[ng-switch-when=\"jsonb\"]').siblings('.text-danger.ng-active').find('div[ng-message=\"" + type + "\"]')[0];", el);
+    }
```
jrchudy: if you remove the different cases for `json` and `jsonb` inputs in the html, having this second function is unnecessary.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -390,6 +472,95 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
+                                      jsonTxtArea.sendKeys(input);
+                                      chaisePage.recordEditPage.getJSONInputErrorMessage(jsonTxtArea, 'json').then(function(error) {
+                                            if(error){
+                                                expect(expectedValue).toBeFalsy();
+                                            }
+                                            else{
+                                                expect(expectedValue).toBeTruthy();                                                
+                                                }
+                                            });                                        
+                                    })(JSONTestParams[i].stringVal, JSONTestParams[i].expectedValue);
```
jrchudy: Kind of confusing reading the code when the end characters don't line up.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -390,6 +472,95 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
+                            expect(undefined).toBeDefined();
+                        }
+                    });
+                });
+            });
+            });
+            }
+
+            if (jsonbCols.length > 0) {
+                describe("JSONB fields, ", function () {
```
jrchudy: indentation in this function is off as well making it difficult to understand.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -390,6 +472,95 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
+                            browser.sleep(10);
+
+                            jsonTxtArea.column = c;
+                            
+                            for (i = 0; i < JSONTestParams.length; i++) {
+                                  jsonTxtArea.clear();
+                                  (function(input, expectedValue) {
+                                      c._value = input;
+                                      jsonTxtArea.sendKeys(input);
+                                      chaisePage.recordEditPage.getJSONInputErrorMessage(jsonTxtArea, 'json').then(function(error) {
```
jrchudy: For this you should be sending the keys to the input field and then verifying that the value in the input field matches the one that you would expect. As another test after that, you can verify whether an error was thrown or not. Currently, if an error is thrown it will still pass the expectation.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -390,6 +472,95 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
+                            chaisePage.recordEditPage.clearInput(jsonBTxtArea);
+                            browser.sleep(10);
+
+                            jsonBTxtArea.column = c;
+                            for (i = 0; i < JSONTestParams.length; i++) {
+                                  jsonBTxtArea.clear();
+                                  (function(input, expectedValue) {
+                                      c._value = input;
+                                      jsonBTxtArea.sendKeys(input);
+                                      chaisePage.recordEditPage.getJSONBInputErrorMessage(jsonBTxtArea, 'json').then(function(error) {
```
jrchudy: Same comment as above about checking the expectations.

test/e2e/utils/chaise.page.js
RFSH: And these are not different from `getTextAreaForAcolumn` function. You could just reuse that function.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -390,6 +472,95 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
                     });
                 });
             }
+            
+            if (jsonCols.length > 0) {
+            describe("JSON fields, ", function () {
```
RFSH: This test case is different from other column types. Try to make it consistent with other types. Take a look at other column types, for example the [markdown](https://github.com/informatics-isi-edu/chaise/blob/20189a1a052f85941fb30b6fd0a2962d62f0f986/test/e2e/utils/recordedit-helpers.js#L407,L474) that Amit wrote.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -50,7 +50,9 @@ var testParams = {
         { title: "Cover Image", value: "3,005", type: "int2", presentation: { type: "url", template: "{{{chaise_url}}}/record/#{{catalog_id}}/product-record:file/id=3005"} },
         { title: "Thumbnail", value: null, type: "int4"},
         { title: "Operational Since", value: "2008-12-09 00:00:00", type: "timestamptz" },
-        { title: "Is Luxurious", value: "true", type: "boolean" }
+        { title: "Is Luxurious", value: "true", type: "boolean" },
+        { title: "json_col", value:'<pre>'+JSON.stringify(null,undefined,2)+'</pre>'},
+        { title: "jsonb_col", value:'<pre>'+JSON.stringify(null,undefined,2)+'</pre>'}
```
nitishksahu: The expectation here is `expect(el.getAttribute(‘innerHTML’)).toBe(column.value);`. so the entire innerHTM is beign picked with the whitespaces that is beign introduced when the JSON value is stringified with a preetiness factor of 2, using JSON.stringify(value,undefined,2);

nitishksahu:
- The requested changes have been done and are available for review.

- The link to deployment of JSON column is [JSON deployment](https://dev.isrd.isi.edu/~nitish/chaise/recordset/#1/public:jsontest)

test/e2e/utils/recordedit-helpers.js
```diff
@@ -321,7 +347,51 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
                 });
             }
+            
+            if (jsonCols.length > 0) {
+                describe("JSON fields, ", function () {
+                    it("should show textarea input for JSON datatype and then set the value", function() {
+                        jsonCols.forEach(function(c) {
+                        chaisePage.recordEditPage.getTextAreaForAcolumn(c.name, recordIndex).then(function(jsonTxtArea) {
+                            if (jsonTxtArea) {
+                                expect(true).toBeDefined();
```
jrchudy: This check is unnecessary. Instead of doing an `if / else` check around whether the value is defined or not, do this instead:
`expect(jsonTxtArea).toBeDefined()`
or a slightly more robust check would be:
`expect(jsonTxtArea).toEqual(jasmine.any(Object))`

test/e2e/utils/recordedit-helpers.js
```diff
@@ -321,7 +347,51 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
+                                jsonTxtArea.column = c;
+                                
+                                JSONDataTypeFields.push(jsonTxtArea);
+                                var value = getRecordValue(c.name);
 
+                                if (value != undefined) {
+                                    expect(jsonTxtArea.getAttribute('value')).toBe(value, colError(c.name , "Doesn't have the expected value."));
+                                }              
+                            }
+                            else{
```
jrchudy: This else condition is unnecessary.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -321,7 +347,51 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
+                        jsonCols.forEach(function(c) {
+                        chaisePage.recordEditPage.getTextAreaForAcolumn(c.name, recordIndex).then(function(jsonTxtArea) {
+                            if (jsonTxtArea) {
+                                expect(true).toBeDefined();
+                                jsonTxtArea.column = c;
+                                
+                                JSONDataTypeFields.push(jsonTxtArea);
+                                var value = getRecordValue(c.name);
 
+                                if (value != undefined) {
```
jrchudy: You shouldn't guard this expectation clause with this check. If the value is undefined, then the expectation will fail. That should be the way to verify this.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -321,7 +347,51 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
+                 it("should only allow valid JSON values", function(){
+                     jsonCols.forEach(function(c) {
+                         chaisePage.recordEditPage.getTextAreaForAcolumn(c.name, recordIndex).then(function(jsonTxtArea) {
+                             for (i = 0; i < JSONTestParams.length; i++) {
+                                 jsonTxtArea.clear();
+                                 (function(input){
+                                     c._value = input;
+                                     jsonTxtArea.sendKeys(input);
+                                     chaisePage.recordEditPage.getJSONInputErrorMessage(jsonTxtArea, 'json').then(function(error){
+                                         if(error){
```
jrchudy: I understand that you are doing this so that _if_ it does error out, the test will fail. It just seems weird that you are testing the lack of an error appearing.

This is fine as is, unless you can think of a way to verify that the input values are not throwing an error (using an expectation).

test/e2e/utils/recordedit-helpers.js
RFSH: @jrchudy This is fine. The `value` is null if we don't provide any current value to the `testPresentationAndBasicValidation` function (when we don't want to test the current value of the inputs). For example in case of create, we don't need to test the current value.

test/e2e/utils/recordedit-helpers.js
RFSH: @jrchudy This is the way all the previous recordedit-helper test cases were written before and I didn't change them when I was refactoring it. @nitishksahu just used the same pattern of testing in here.
```javascript
if (exists) {
// do other tests knowing the field exists
} else {
// return error that the field is does not exist
}
```
We can change them to just have the `toBeDefined` as the first expect and remove `else` clause. This pattern has been used heavily in this file.

jrchudy: That's my mistake. I hadn't realized that was still the case.

test/e2e/utils/recordedit-helpers.js
RFSH: @jrchudy This is also something that has been written before and @nitishksahu has used those as an example. We can remove the `if() {} else {}` the same way that you suggested on the other comment.

test/e2e/utils/recordedit-helpers.js
jrchudy: It should be checking for something that verifies that it's in create mode then. I'm imagining a bug could appear that doesn't have the value input into the UI properly for edit mode and we won't catch that with this test case (and any others that are doing this).

For now, if we aren't going to change this case or the above one, we should add a comment about it.

test/e2e/utils/recordedit-helpers.js
RFSH: It was my mistake for just ignoring these weird test cases. I think we should change those.

nitishksahu:
All the if/else condition for all the specs in `recordedit-helpers.js` has been changed.

jrchudy:
The changes are good and I can verify that the tests still pass. Merge when you are ready.

nitishksahu:
The JSON column support in chaise changes has been merged to master.

PR #1181, opened 2017-07-07
This PR addresses the issue for navbar login button using wrong login flow

nitishksahu:
The onClick function bound to the login button in navbar was pointing to the login function that uses the old login flow (where the login form appears in the same window). Now the login button flow is adjusted to use the new modal window for login.

Files Changed for this are: 

- `common/navbar.js`

- `common/errors.js`

- `recordset/recordset.js`


recordset/recordset.js
```diff
@@ -78,10 +78,12 @@
         $scope.navbarBrand = (chaiseConfig['navbarBrand'] !== undefined? chaiseConfig.navbarBrand : "");
         $scope.navbarBrandImage = (chaiseConfig['navbarBrandImage'] !== undefined? chaiseConfig.navbarBrandImage : "");
         $scope.navbarBrandText = (chaiseConfig['navbarBrandText'] !== undefined? chaiseConfig.navbarBrandText : "Chaise");
-
+        var reloadCb = function() {
+                            window.location.reload();
+                        }; 
         $scope.login = function() {
```
RFSH: I don't think these login and logout function are even being used. You can completely remove these two functions.

common/navbar.js
```diff
@@ -17,7 +17,9 @@
             }
 
             var q = [root];
-
+            var reloadCb = function() {
```
RFSH: Fix the indentation.

common/errors.js
```diff
@@ -76,12 +76,15 @@
                 delete modalProperties.backdrop;
                 params.canClose = true;
             }
-
+            
             var modalInstance = $uibModal.open(modalProperties);
-
+            var reloadCb = function() {
```
RFSH: Fix the indentation.

nitishksahu:
The changes have been done to reflect the two cases -

1. Click on Login link on the navbar, opens a new pop window directly to log in.

2. When a user is in middle of some operation (adding/ editing record) and the session expires, in that case, a modal window will be shown, with the message of "expired session" and asked to log in again.

> The second case could not be verified end to end, as we get a cross-origin error thrown when a frame is loaded from dev.isrd.isi.edu/~username. Here `~username` cause the cross-origin error to occur.

common/authen.js
```diff
@@ -20,6 +20,148 @@
...
+                            cb();
+                            closed = true;
+                        }
+                    }
+                });
+            }
+        };
+        
+        var logInHelper = function(logInTypeCb, cb){
+            console.log(cb);
```
RFSH: Remove console.logs

common/authen.js
```diff
@@ -20,6 +20,148 @@
                 _changeCbs[k]();
             }
         };
+        
+        var popUpLoginCb = function(params, referrerId, cb){
```
RFSH: This function and function below it share a lot of code. Can we merge them together? Instead of having two different function, we can have a function that as a parameter will determine to open a modal or not.

PR #1184, opened 2017-07-10
Collapse table

amitjha21:
Addresses Issue #1101 
All related tables collapse after no of these tables crosses cut-Off value. The value is maintained in `maxRelatedTablesOpen` variable in `chaise-config.js`


record/record.app.js
```diff
@@ -112,7 +112,8 @@
 
                 $rootScope.tableModels = [];
                 $rootScope.lastRendered = null;
-
+                var cutOff = (angular.isUndefined(chaiseConfig.maxRelatedTab) || chaiseConfig.maxRelatedTab === null || chaiseConfig.maxRelatedTab == " ")?Infinity:chaiseConfig.maxRelatedTab;
```
jrchudy: This conditional seems overly expressive. You could verify this by checking `maxRelatedTab > 0 ? maxRelatedTab : Infinity`.

test/e2e/specs/all-features/record/presentationRelatedTab.spec.js
```diff
@@ -0,0 +1,91 @@
+var chaisePage = require('../../../utils/chaise.page.js');
+
+var EC = protractor.ExpectedConditions;
+var testParams = {
```
jrchudy: What is the purpose of defining this whole `testParams` object? It looks like you are only using a small subset of these values. There's no reason to bloat the spec with unnecessary variable definitions.

test/e2e/specs/all-features/record/presentationRelatedTab.spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+            var url = browser.params.url + "/record/#" + browser.params.catalogId + "/product-record:" + testParams.table_name + "/" + keys.join("&");
+            browser.get(url);
+            accordionSet = element.all(by.css('.related-table-heading'));
+            chaisePage.waitForElement(chaisePage.recordPage.getEntityTitleElement(), browser.params.defaultTimeout);
+        });
+
+        it('should collapse related tables after cutoff value',function(){
+            browser.executeScript("return chaiseConfig;").then(function(chaiseConfig) {
+            accordionSet.count().then(function(accrCount){
```
jrchudy: You should use promise chaining here instead of nested promises.
```
browser.executeScript().then(function() {
  return accordionSet.count();
}).then(function() {
  expectation
});
```
And then you should attach a `.catch()` to the end of that promise chain in case something errors out.

test/e2e/specs/all-features/record/presentationRelatedTab.spec.js
```diff
@@ -0,0 +1,91 @@
...
+                if(accrCount>chaiseConfig.maxRelatedTab){
```
jrchudy: Please make this expectation deterministic. You should know how many related tables are visible and you should have a chaise config option to check that against.

Makefile
```diff
@@ -40,6 +40,7 @@ E2EDrecord=test/e2e/specs/all-features-confirmation/record/presentation-btn.conf
 E2EDrecordCopy=test/e2e/specs/all-features/record/copy-btn.conf.js
 E2ErecordNoDeleteBtn=test/e2e/specs/delete-prohibited/record/no-delete-btn.conf.js
 E2EDrecordRelatedTable=test/e2e/specs/default-config/record/related-table.conf.js
+E2EDrecordCollapseRT=test/e2e/specs/all-features/record/presentationRelatedTab.conf.js
```
jrchudy: Make sure to add this to target to `make testall` and `make test`.

jrchudy:
- I removed the search target from `make testparallel` as it is included in the default config set of parallel tests. 
 - Created a config file and new schema document for the new feature. This new schema has 5 related tables instead of 3. This was done because the `multi-permissions` tests had 4 related tables that required them to be open for their respective tests. Adjusted the `maxRealtedTablesOpen` config option to 4 for this case
 - moved the test case for this new feature into the `copy-btn` spec. This new feature is really simple and doesn't warrant it's own spec file. Instead `copy btn`  spec handles both by importing 2 separate config documents (see `copy-btn.dev.json`). Maybe the copy btn files should be renamed more appropriately now?
 - changed `travis.yml` so it isn't running the navbar tests multiple times now. Navbar tests are each part of separate configs (`all-features` and `all-features-confirmation`).

PR #1187, opened 2017-07-12
fixes, cleanups, and install can install remotely

robes:
Trying to make the dev/test workflow easier. You can install to your ~userdir like this.

First set this:
```
export CHAISEDIR=USER@HOSTNAME:public_html/chaise
```

Then run this:
```
make install
```

Your local code gets pushed to your userdir on the test server.


hongsudt:
@jrchudy if this is good, please close the issue.

robes:
This is the PR that we need to enable PR/branch tests in travis to see if we can get the webdrive/protractor installation right.

Issue #1098, opened 2017-05-03
chaise record page should enforce uniqueness of its entity reference


robes:
This is a subtle issue. Notice these two links:

https://www.facebase.org/data/record/#1/vocabulary:anatomy/term=Head
 
https://www.facebase.org/data/record/#1/vocabulary:anatomy/iri=http%3A%2F%2Fpurl.org%2Fsig%2Font%2Ffma%2Ffma7154

They point to exactly the same base record, however, in the first one you will see a lot of related entities (which is actually a bug) while in the second you will not.

The problem is that when we form the URLs for the related references we are using the base query pattern `term=Head` which is _not unique_ and so we are linking on a record **set** not a single unique record.

This was exposed by a funny corner case where a user entered a chaise URL by hand -- something that I think we actually always wanted users to do.

My feeling is that the bug is _not_ the related record linking. The join is actually working correctly.

The bug is that we have not enforced uniqueness in chaise + ermrestjs.

I think the fix is
- [ ] ermrestjs should implement the skeleton API that I wrote for checking uniqueness (simple introspection will do it)
- [ ] chaise record app should use the ermrestjs API and throw an exception then the condition fails

This is the companion issue https://github.com/informatics-isi-edu/ermrestjs/issues/399

karlcz:
An alternative if you want to allow unique-in-practice query terms would be to just change
the record fetch to use limit=2 and throw an error (or redirect to recordset) if a second
result is available?

hongsudt:
I like a more flexible approach that @karlcz suggested.. This will allow user to continue as usual if there is no ambiguity.. And if there are duplicate, provide them a link to recordset..

nitishksahu:
An implementation of how multiple records are going to be handled in record page is in place:

[A record page with more than 1 record](https://dev.isrd.isi.edu/~nitish/chaise/record/#1/vocabulary:anatomy/term=Head)

@robes @hongsudt @karlcz Any suggestions??

nitishksahu:
The changes to address this issue have been merged with the master branch.

Instances of the issue in different deployment:

- [DEV Isrd](https://dev.isrd.isi.edu/chaise/record/#1/vocabulary:anatomy/term=Head)
- [RBK](https://dev.rebuildingakidney.org/chaise/record/#2/Protocol:Protocol_Author/Rank=1)
[...]

PR #1190, opened 2017-07-12
chaise record page should enforce uniqueness of its entity reference

nitishksahu:
This PR addresses the issue where more than one entity is fetched in the record, then the record page displayed a lot of unrelated entities.

With this PR: 
An error modal window will be displayed if there are more than 1 entity fetched for a given constraints and on the click of OK button, the page will redirect to recordset page to display all the fetched results appropriately.

An implementation of this is available [here](https://dev.isrd.isi.edu/~nitish/chaise/record/#1/vocabulary:anatomy/term=Head@sort(id))


record/record.app.js
```diff
@@ -89,6 +89,14 @@
                     var noDataError = ErrorService.noRecordError(context.filter.filters);
                     throw noDataError;
                 }
+                else if(page.tuples.length > 1){
+                    var recordPageLink = page._ref.appLink;
```
RFSH: Since javascript doesn't have a notion of private and public, to define a private property we add `_` before the name of it. That means `_ref` is a private property of Page and you should not use it. You can use `page.reference` instead.

Also don't replace the link, use applink function to get the link for other contexts. In this case:
```javascript
var recordSetLink = page.reference.contextualize.compact.appLink;
```
recordset app is using the `compact` context. That's why you should first contextualize reference to the appropriate context and then get the `appLink` for that.

record/record.app.js
```diff
@@ -89,6 +89,14 @@
...
+                    var recordSetLink = recordPageLink.replace("record","recordset");
+                    AlertsService.addAlert('More than one records found for your query filters, Redirecting you to Recordset Page.....', 'warning');
```
RFSH: This alert is redundant, you're showing the error modal anyways.

common/templates/errorDialog.modal.html
```diff
@@ -5,7 +5,8 @@ <h3 class="modal-title">Error: {{::ctrl.params.errorCode}}</h3>
 <div class="modal-body">
     {{::ctrl.params.message}}
     <br><br>
-    Click OK to return to the {{::ctrl.params.pageName}}. 
+    Click OK to {{ctrl.params.errorCode == 'Multiple Records Found'? 'redirect ': 'return '}} to the {{::ctrl.params.pageName}}.
```
RFSH: I think saying "return" in all the cases is fine. You don't need this check

record/record.app.js
```diff
@@ -33,8 +33,8 @@
         $uibTooltipProvider.options({appendToBody: true});
     }])
 
-    .run(['constants', 'DataUtils', 'ERMrest', 'ErrorService', 'headInjector', 'MathUtils', 'Session', 'UiUtils', 'UriUtils', '$log', '$rootScope', '$window',
-        function runApp(constants, DataUtils, ERMrest, ErrorService, headInjector, MathUtils, Session, UiUtils, UriUtils, $log, $rootScope, $window) {
+    .run(['constants', 'DataUtils', 'ERMrest', 'ErrorService', 'headInjector', 'MathUtils', 'Session', 'UiUtils', 'UriUtils', '$log', '$rootScope', '$window','AlertsService',
+        function runApp(constants, DataUtils, ERMrest, ErrorService, headInjector, MathUtils, Session, UiUtils, UriUtils, $log, $rootScope, $window,AlertsService) {
```
RFSH: Since the alert that you added is unnecessary don't forget to remove it from here too.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -134,4 +140,29 @@ describe('View existing record,', function() {
...
+        });
+        
+        it('A error modal window should appear with multiple records found error with correct title', function(){
+            var modalTitle = chaisePage.recordPage.getErrorModalTitle();
+            expect(modalTitle).toBe(testParams.multipleData.title, "The title of multiple record error pop is not correct");
+            
+        });
+        
+        it('On click of OK button the page should redirect to recordset page', function(){
+            chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
```
RFSH: I don't know if we prefer `executeScript` or not. But you could rewrite it this way:
```javascript
chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
    return btn.click();
}).then(function () {
}).then(function (currentUrl) {
    expect(currentUrl).toContain("recordset", "The redirection from record page to recordset in case of multiple records failed");
}).catch( function (err) {
    console.log(error);
}); 
```
@jrchudy Is the way this test case is written fine (using executeScript for getting current url)? or should we change it?

common/errors.js
```diff
@@ -125,26 +133,32 @@
...
             var reloadCb = function() {
                 window.location.reload();
             }; 
             if (exceptionFlag || window.location.pathname.indexOf('/search/') != -1 || window.location.pathname.indexOf('/viewer/') != -1) return;
 
             if (ERMrest && exception instanceof ERMrest.UnauthorizedError || exception.code == errorNames.unauthorized) {
                 Session.loginInAPopUp(reloadCb);
-            } else {
+            }
+            else if (window.location.pathname.indexOf('/record/') !== -1 && exception.code && exception.code.indexOf(messageMap.multipleDataErrorCode) !== -1){
```
RFSH: You don't need to check for certain app.

common/templates/errorDialog.modal.html
nitishksahu: Yesterday, Hongsuda suggested this change in the verbose, so I introduced this condition, if we need `return` in all case then I can remove this condition

RFSH: I talked with @hongsudt , she prefers using "go" instead of "return" or "redirect". So just change it to "Click OK to go to the".

common/errors.js
```diff
@@ -125,26 +133,32 @@
...
             var reloadCb = function() {
                 window.location.reload();
             }; 
             if (exceptionFlag || window.location.pathname.indexOf('/search/') != -1 || window.location.pathname.indexOf('/viewer/') != -1) return;
 
             if (ERMrest && exception instanceof ERMrest.UnauthorizedError || exception.code == errorNames.unauthorized) {
                 Session.loginInAPopUp(reloadCb);
-            } else {
+            }
+            else if (exception.code && exception.code.indexOf(messageMap.multipleDataErrorCode) !== -1){
```
RFSH: This `exception.code` is usually a number (409, 401, etc.). Since these `multipleRecordError` and `noRecordError` are custom errors that we made, we either have to change the code to something meaningful (an actual error code number) and then add `sub-code`, or we can accept the code to be string in some cases.

In either cases, this check `exception.code.indexOf(messageMap.multipleDataErrorCode) !== -1` can cause problem, change it to  `exception.code == errorNames.multipleRecords`.

hongsudt:
@nitishksahu if the code for this has been merged, please close this issue. 

If you want to have the featured reviewed by others, assign a reviewer to this issue. I think we have a separate issue already for the feature review.

nitishksahu:
This PR is related to [issue#1098](https://github.com/informatics-isi-edu/chaise/issues/1098)

@hongsudt this issue is waiting on your queue to verify, if everything looks good, we can close this.

hongsudt:
This looks good. You can close the issue.

Issue #1195, opened 2017-07-13
Static page for markdown help


amitjha21:
We should have help pages hosted along with the application. These pages will be customized for internal applications and kept locally to avoid any future disruptions.

amitjha21:
@hongsudt 
Please review the markdown help page.

hongsudt:
The help page looks really nice.

PR #1206, opened 2017-07-18
Markdown help page

amitjha21:
Addresses issue #1195 .
A new page added in `recordedit` app. The page opens in a new window with static help content. 

- mdHelp.html
- mdhelpStyle.min.css
- mdHelp.controller.js

Example: [Link](https://dev.rebuildingakidney.org/~amit/chaise/recordedit/mdHelp.html)


RFSH:
@amitjha21 It seems like you're not `markdown-container` class here for the markdown results. It's not showing the bullet points.

amitjha21:
Added `li` styling for bullets.

test/e2e/specs/default-config/recordedit/add-defaults.spec.js
```diff
@@ -168,7 +177,37 @@ describe('Record Add with defaults', function() {
             expect(foreignKeyInput.getText()).toBe(values.foreign_key_value, "Foreign key input default is incorrect");
             expect(foreignKeyDisabledInput.getText()).toBe(values.foreign_key_disabled_value, "Foreign key disabled default is incorrect");
         });
-
+        //mdhelp page
```
RFSH: I know that we want to avoid having multiple test specs, but this is just too confusing for me. You added this test case to `add-defaults` specs. But this has nothing to do with "add defaults" apart from being invoked from record-edit. I think we should have a test spec for these kind of common test cases, instead of randomly adding them to a spec.

We can combine `footer`, `navbar` and also this test case into a "data-dependent" (or any other name) spec.

amitjha21: The reason I added it in `default-config` was that it does not require any special config for testing. We can certainly refactor mentioned test cases to avoid confusion in future.  
I will create another spec inside `all-features\recordedit` for markdown help page.

jrchudy: `default-config` is more specifically meant to test features that are affected by config options when those options are not defined. In this case, the markdown page will always show up regardless of any configuration. It almost doesn't matter what config set this test goes in. So we should focus on putting this test in a spec that seems appropriate.

With that being said, we actually don't have a test that is meant to just test the presentation of the `recordedit` app. We test that presentation as part of the add spec (`/specs/all-features-confirmation/recordedit/add.spec.js`). This seems like the most appropriate place for this test case.

hongsudt:
By default, the privacy statement should never show up unless there is a
footerMarkdown config. Not all deployments have the privacy statement
present.

jrchudy:
@hongsudt this is about the markdown page, not the footer privacy policy.

.gitignore
```diff
@@ -21,6 +21,6 @@ index.html
 sftp-config.json
 npm-debug.log
 version.txt
-
+mdHelp.html
```
RFSH: Is this necessary?

common/vendor/MarkdownEditor/bootstrap-markdown.js
```diff
@@ -1294,7 +1294,8 @@
                   glyph: "glyphicon glyphicon-question-sign"
                 },
                 callback: function(e) {
-                    window.open('http://commonmark.org/help/');
+                    // window.open(window.location.origin+window.location.pathname+'mdHelp.html');
```
RFSH: Remove this.

test/e2e/specs/all-features-confirmation/recordedit/add.spec.js
```diff
@@ -212,10 +221,43 @@ describe('Record Add', function() {
                 }
             });
         });
-
+        describe('Markdown Editor Help button is clicked, ', function() {
```
RFSH: Mention in a comment that the test cases inside this `describe` assume that,

1. We are in recordedit page.
2. There's at least a column with markdown type.

RFSH: On the second thought, it's better to bring it outside the outer `describe`. It's currently inside "When url has a prefill query string param set," which shouldn't be. Then you can open the recordedit page yourself and don't need to add those assumptions as it will be clearly visible from code that it's going to which app and table.

.gitignore
amitjha21: As all `index.html` files have been added in `.gitignore`; I assume it is only appropriate to add `mdHelp.html` too.

RFSH: You're right.

test/e2e/specs/all-features-confirmation/recordedit/add.spec.js
amitjha21: I put this inside the outer `describe` just to avoid opening browser for this one case.

test/e2e/specs/all-features-confirmation/recordedit/add.spec.js
```diff
@@ -218,4 +227,57 @@ describe('Record Add', function() {
         });
     });
 
+    describe('Markdown Editor Help button is clicked, ', function() {
+        var testCookie = {};
+        beforeAll(function() {
+            // Refresh the page
+            browser.get(browser.params.url + "/recordedit/#" + browser.params.catalogId + "/product-add:accommodation");
+            chaisePage.waitForElement(element(by.id("submit-record-button"))).then (function () {
```
RFSH: This wait (adding cookie) is not relevant for this test case.

test/e2e/specs/all-features-confirmation/recordedit/add.spec.js
```diff
@@ -218,4 +227,57 @@ describe('Record Add', function() {
...
+        }).then(function() {
+            // - Go back to initial Record page
+            browser.close();
+            browser.switchTo().window(allWindows[0]);
+        }).catch(function(error) {
+            console.dir(error);
+            expect('Something went wrong with this promise chain.').toBe('Please see error message.','While checking markdown help page');
+        });
+    });
+    afterAll(function() {
```
RFSH: It's not relevant to this test case

test/e2e/specs/all-features-confirmation/recordedit/add.spec.js
```diff
@@ -218,4 +227,57 @@ describe('Record Add', function() {
...
+                    rowname: {
+                        value: chance.sentence()
+                    },
+                    keys: {id: 1}
+                };
+                browser.manage().addCookie('mdtest', JSON.stringify(testCookie));
+            });
+
+        });
+    //mdhelp page
```
RFSH: Indentation is off by one level.

Issue #462, opened 2016-06-21
review/update the structure of chaise config


hongsudt:

robes:
It will be some type of Angular provider, probably a config provider.

howdyjessie:
Maybe a value provider or a constant provider, depending on whether we need Chaise config during config phase.

It'd be nice to introduce some hierarchy to the Chaise config object so it's easier to differentiate between which settings are global (e.g. `customCSS`) and which ones are app- (e.g. `defaultAnnotationColor`) or context-specific (e.g. `editRecord`). Or at least have a matrix in our [configuration guide](https://github.com/informatics-isi-edu/chaise/blob/master/doc/configuration.md) so users can quickly see which settings apply in which apps and how.

Remove deployment-specific defaults in `chaise-config-sample.js`.

robes:
Here's my recommendation. We should move the chaise config to the catalog (i.e., to the database) and serve it up from ermrest. We create a database structure that roughly mirrors the current chaise config although we could improve that structure. I would agree with Jessie's general observations above: to separate global from app-specific settings. Pushing this into the catalog would also give us: catalog-specific configurations and the ability to do online editing of its values.

Sub-tasks:
- go through each property of the config and categorize as: global vs <app or component> specific
- define a db schema for `chaise` or `deriva` if we want this more general, with tables
  - `global_config`
  - `record_config`, `recordset_config`,...
  - `navbar_config`, other component-specific,...
- define additional SQL views over the schema where it would simplify retrieval, if necessary. For example, the navbar menu might be structured across multiple tables for menus, menu items, and sub-menus etc given its hierarchical structure. This is one candidate that crosses my mind as something that _might_ be simpler to serve up with a SQL view rather than a ermrest path.
- within chaise (js code) there should probably be an angular service for getting the config and it should get the global config entity, the app specific config entity, and any relevant component-specific config entities.

robes:
If we want to do this, we can probably spec out another task but this is what I'd suggest as a review of this issue.

karlcz:
Once the ermrest/acls-everywhere branch is merged, we'll have catalog-level annotations which could be used to store a chaise config JSON document.

robes:
That's another option yes. But its not easy to edit those annotations online. If its part of the db schema you could actaully edit it online easily. Alternately, we'd either have to punt on that or implement some general annotation editing feature.

karlcz:
[...]

PR #1216, opened 2017-07-22
Recordedit no changes before submission

jrchudy:
When the user tries to submit data without making any changes to the form, they will be given a warning message stating that no changes were actually made.

Current warning message:
image

This issue relies on PR [#462](https://github.com/informatics-isi-edu/ermrestjs/pull/462) in `ermrestJS`.


jrchudy:
@RFSH I have finished writing the test for this. I also updated the configuration for the `multi-edit` test to use the data in `multi-edit` folder so as to not mix data for `multi-add` and `multi-edit`. I noticed the data in `multi-add` was not consistent with the schema definition for `multi-add:multi-add-table`.

Issue #878, opened 2016-12-16
add total number of records found in the recordset


hongsudt:
On the resultset page, there should be a total number of records that match the specified criteria. This should be implemented asynchronously, so that the initial page load doesn't have to wait for it. 

To support this, we probably need an ermrestjs api to make an aggregate call to ermrest.

robes:
We probably also need to consider making this optional either by config or by annotation

On large tables this is going to add a lot of overhead.

karlcz:
We already have the async API to do it via promises.  We just need the app to launch this extra query and add the count to the page when it is ready rather than delaying page rendering until the count returns.

karlcz:
yes, particularly a large table with row-level security can be 10-20x slower than one without...

hongsudt:
By default, this feature should be on though since most of the tables do not have performance issue..

hongsudt:
This is pending on informatic-isi-edu/ermrestjs#304.

jrchudy:
Before total count comes back:
Displaying X records
After:
Displaying X from Y records

jrchudy:
the corresponding PR has been merged. Closing this issue.

PR #1224, opened 2017-07-31
Rs aggregate count

jrchudy:
This PR adds functionality to the `recordset` directive to put a full entity count for the current reference at the top of the table. This can be seen in the `recordset` app and the `recordedit` foreign key popup selector.

This addresses issue #878.

### DIFF
--- a/common/table.js
+++ b/common/table.js
@@ -342,6 +342,17 @@
                 window.updated = function() {
                     updated = true;
                 }
+
+                // get the total row count to display above the table
+                scope.$on('recordset-loaded', function($event, hasLoaded) {
+                    if (hasLoaded) {
+                        scope.vm.reference.getAggregates([scope.vm.reference.aggregate.countAgg]).then(function getAggregateCount(response) {
+                            scope.vm.totalRowsCnt = response[0];
+                        }, function error(response) {
+                            throw response;
+                        });
+                    }
+                });
             }
         };
     }]);
--- a/common/table.js
+++ b/common/table.js
@@ -344,7 +344,7 @@
                 }
 
                 // get the total row count to display above the table
-                scope.$on('recordset-loaded', function($event, hasLoaded) {
+                scope.$on('get-table-aggregate-count', function($event, hasLoaded) {
                     if (hasLoaded) {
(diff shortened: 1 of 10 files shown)


common/table.js
```diff
@@ -100,6 +100,7 @@
         */
         function read(scope, isBackground) {
 
+            if (scope.vm.search == '') scope.vm.search = null;
```
jrchudy: This is done so that the check on line 351 can compare like values. When the user uses back space to clear the search term, the search term field is `''` and the one on the new location object is set to `null` when it sees empty string as a search term.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -613,6 +613,9 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
                                     }).then(function(ct) {
                                         expect(ct).toBe(col.count, colError(col.name, "number of foreign key rows are not as expected."));
 
+                                        return chaisePage.recordsetPage.getTotalCount().getText();
+                                    }).then(function(text) {
+                                        expect(text).toBe("Displaying 5 of " + col.count + " Records", "The total count display in the foreign key popup is incorrect");
```
RFSH: The way I wrote the test cases `col.count` was showing the number of rows that are visible. So it's not actually the total number. This shouldn't be a problem if the count is less than 25 (maximum page size allowed).  You can make a note of it in the jsdoc of this function, or add another `totalCount` variable.
Also change the exception message to use `colError`  which will add column name and index.

```javascript
// if you're adding the note that count should not be more than 25.
var expected = "Displaying " + col.count + " of " + col.count + " Records";

// if you're adding the new variable
var expected = "Displaying " + col.count + " of " + col.totalCount   + " Records";

expect(text).toBe(expected, colError(col.name, "The total count display in the foreign key popup is incorrect");
```
Is the maximum number of foreignkeys 25? if it's not please change it accordingly.

jrchudy:
@RFSH I addresses your comments and added another few test cases to make sure the displayed value is updated when the user triggers foreground searches.

Issue #1211, opened 2017-07-19
enable multi-select in the table directive


hongsudt:
Currently, the table directive (i.e. the one that is used for users to select fks) assumes that only one record will be selected. We should extend the API to allow multi-select option.

PR #1228, opened 2017-08-03
Multi-select functionality for recordset tables

jrchudy:
The recordset directive can now be used in a multi selection mode. This will allow for future functionality to be developed for these tables to include more fine grained multi-edit functionality and better modal popup selectors that allow for multiple rows to be selected. The changes made in this PR include the following:
 - Ellipses directive takes in an extra argument, `selected`
 - Table config option `selectable` changed to `selectMode` with 3 modes, ('no-select', 'single-select', 'multi-select')
 - a set of pills is available at the top of the `recordset` page when in multi-select mode
 - action columns is changed to a select column in multi-select mode with an `All` and `None` button visible
 - recordsetModel (model for recordset app, but also for the recordset directive vm) has a selectedRows array defined for it now
 - `record-bookmark-container` class was renamed to `bookmark-container` so the bar on `recordset` looks similar to the one on `record`

This PR resolves issue #1211. This PR has to be merged after `<insert ermrestJS PR here>`.

### DIFF
--- a/common/templates/recordset.html
+++ b/common/templates/recordset.html
@@ -1,7 +1,7 @@
 <div ng-show="vm.initialized" class="row" style="margin-bottom: 10px;">
-    <div class="row total-filters">
+    <div ng-if="vm.selectedRows.length > 0" class="row total-filters">
         <div class="col-lg-12 col-md-12 col-sm-12 col-xs-12">
-            <a ng-if="vm.selectedRows.length > 0" class="remove-link" ng-click="removeAllPills()">
+            <a class="remove-link" ng-click="removeAllPills()">
                 <span class="filter-label label label-primary">
                     clear all
                 </span>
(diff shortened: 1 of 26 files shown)


common/templates/recordset.html
```diff
@@ -1,4 +1,17 @@
 <div ng-show="vm.initialized" class="row" style="margin-bottom: 10px;">
+    <div ng-if="vm.selectedRows.length > 0" class="row total-filters">
```
RFSH: Can you add a configuration to disable showing these? In case of faceting entity picker, we don't want to show the selected rows here and instead on the panel title.

recordset/recordset.js
```diff
@@ -88,6 +89,32 @@
...
 
+        // TODO: This function is used for multi-select functionality.
+        /**
+         * Creates a displayObj with a unique identifier to store the selected rows
+         *  displayname =   used for the display value in the pills for which which row is selected`
+         *  key =           unique identifier that is composed from each shortest key column's value
+         *
+         * Facilitates the multi select functionality for multi edit in the future
+         */
+        // $scope.onSelect = function(tuple) {
```
RFSH: Shouldn't this be a function for `recordTable ` directive?

jrchudy:
@RFSH I moved the `onSelect` functionality into `table.js` in the table directive. I also made sure the selected rows filters only show up when the table is also in `multi-select` mode.

Issue #643, opened 2016-09-30
prototype related entities layout to look like it is part of the main table


hongsudt:
- [ ] if the inbound fk is in the visible column, display this pseudo-column like any other regular column. The table display name should be used as the column name (this should be provided by ermrestjs already). The column value should be a table or what's specified in the table annotation.
- [ ] Add edit and view all capability. We will need to experiment with the UI.
- [ ] change the default display to be a list of row name instead of a tabular format. We will need to have an annotation to show a tabular format.

amitjha21:
This has been merged to master. Assigning @hongsudt to verify and close it.

PR #1232, opened 2017-08-08
Related Layout Table

amitjha21:
This PR addresses issue #643.
Prototype related table should now look like part of main table. The action bar items appear on the top right  corner. 
For non-tabular display content wraps around the action items; for tabular display action items appear on the top corner.


record/record.app.js
```diff
@@ -40,6 +40,8 @@
         var session,
             context = {};
         $rootScope.displayReady = false;
+        $rootScope.recDisplayReady = false;
+        $rootScope.inbFK = false;
```
RFSH: Is this something that you're using?

record/record.app.js
```diff
@@ -58,6 +60,53 @@
...
+                model.rowValues = DataUtils.getRowValuesFromPage(page);
+                model.config = {
+                    viewable: true,
+                    editable: $rootScope.modifyRecord,
+                    deletable: $rootScope.modifyRecord && $rootScope.showDeleteButton,
+                    selectable: false
+                };
+                // return model;
+                callback(model);
+        },function readFail(error) {
```
RFSH: And also please add comments on your functions or pieces of code that might be necessary.

record/record.app.js
```diff
@@ -58,6 +60,53 @@
 
         ERMrest.appLinkFn(UriUtils.appTagToURL);
 
+        function getPageSize(obj){
+            return ((!angular.isUndefined(obj) && obj.display.defaultPageSize)?obj.display.defaultPageSize:constants.defaultPageSize);
+        }
+
+        function getRelatedTableData(refObj,displaytype,accordionOpen,callback){
```
RFSH: Isn't display type something that can be derived from the reference object? I don't think you need to pass the "displaytype". Also add documentation for this function.

common/templates/recordAction-ent.html
```diff
@@ -0,0 +1,20 @@
+<div ng-class="{row:rowStyle}">
```
amitjha21: `rowstyle` is a `bool` which sets the `row` class of `bootstrap`. When this is set to `true` it keeps action bar on top line and table just below it. When `false` action bar appear in the same line as markdown bubbles.

jrchudy: So rather than updating `rowStyle` in the controller, change the conditional in the HTML to check what mode the table display is in.

common/record.js
```diff
@@ -1,22 +1,77 @@
...
+    angular.module('chaise.record.display', ['chaise.record.table'])
 
     .filter('trustedHTML', ['$sce', function($sce){
         return function(text) {
             return $sce.trustAsHtml(text);
         };
     }])
-
-    .directive('recordDisplay', [function() {
+     /* *Directive record-display is called to display related table as part of entity layout.
```
RFSH: Fix the indentation please. When you're changing code make sure the indentation is correct.

Also jsdoc only understands the following structure for comments:
```javascript
/**   DESCRIPTION HERE   */
```
Notice the two `*`s (without space in between). It's also cleaner if you leave the first line empty and also start each line with a `*`. If you're using atom, when you type `/**` the line before and then press enter it will generate a template of jsdoc which you can then edit.

```javascript
/**
 * DESCRIPTION
 * @param {string} param parameter's explanation
 */
function test(param) {}
```

`@param` syntax is not the way that you used it here. Take a look at [jsdoc documentation](http://usejsdoc.org/) for more information about the tags. I think we should add a very simple jsdoc explanation to our dev guides.

record/record.app.js
```diff
@@ -57,7 +58,59 @@
         context.pageId = MathUtils.uuid();
 
         ERMrest.appLinkFn(UriUtils.appTagToURL);
-
+        /* @desc getPageSize(obj) returms page size of the obj.           */
+        function getPageSize(obj){
+            return ((!angular.isUndefined(obj) && obj.display.defaultPageSize)?obj.display.defaultPageSize:constants.defaultPageSize);
+        }
+        /*@desc  getRelatedTableData(refObj, accordionOpen, callback) returns model object with all required component values
```
jrchudy: This isn't being formatted like a jdoc because there isn't an extra `*`. i.e. `/** @desc ...`.

record/record.app.js
```diff
@@ -57,7 +58,59 @@
...
+        *       that is needed by <record-table>, <record-action-bar> and <record-display> diretives.
+        *       refObj: Reference object with component details
```
jrchudy: add `@param` and `{<type>}` here with the proper type of the expected attribute.

record/record.app.js
```diff
@@ -57,7 +58,59 @@
...
+        *       accordionOpen: bool value
```
jrchudy: add `@param` and `{<type>}` here with the proper type of the expected attribute.

record/record.app.js
```diff
@@ -57,7 +58,59 @@
...
+        *       callback: to be called after function processing
```
jrchudy: add `@param` and `{<type>}` here with the proper type of the expected attribute.

record/record.app.js
```diff
@@ -57,7 +58,59 @@
         context.pageId = MathUtils.uuid();
 
         ERMrest.appLinkFn(UriUtils.appTagToURL);
-
+        /* @desc getPageSize(obj) returms page size of the obj.           */
```
jrchudy: This isn't being formatted like a jdoc because there isn't an extra `*`. i.e. `/** @desc ...`

jrchudy:
When I run the tests, there are tests failing in `record/presenentation.spec.js`. Can you run the tests and make sure they are still passing for you so I know whether it's something in my environment or not.

jrchudy:
Everything looks good now. The e2e tests pass from beginning to end for me.

Issue #1227, opened 2017-08-02
making "View" function more obvious


crisaless:
Users aren't 'getting' that they should click the "View" icon to see a details page. Two options:

- Make button more obvious: change one of the classes on the "a" tag from btn-default to btn-primary.

- In addition, could replace <span class="glyphicon glyphicon-eye-open"></span> with the text "View Details"

amitjha21:
Added `btn-primary` class in the ellipses template with `View Details` as the tool-tip for demo.

### [Example](https://dev.rebuildingakidney.org/~amit/chaise/recordset/#2/RNASeq:Sample@sort(Last_Modified_Timestamp::desc::,Accession_ID))

![image](https://user-images.githubusercontent.com/9977014/29146997-321d4438-7d19-11e7-9c39-1f0b548cfbf3.png)

hongsudt:
can you make all three buttons look the same.

amitjha21:
Added `btn-primary` for all buttons.
![image](https://user-images.githubusercontent.com/9977014/29149624-36f06362-7d2a-11e7-8aae-efe389100037.png)

amitjha21:
This has been merged into master. Assigning to @crisaless to verify and close it.

crisaless:
looks good, thanks.

PR #1235, opened 2017-08-10
View button panel

amitjha21:
This PR adresses #1227 .
`ellipses` directive has been changed. All buttons now belong to `btn-primary` class.

_File Changed:_

- chaise/common/templates/ellipses.html 
### [Example](https://dev.rebuildingakidney.org/~amit/chaise/recordset/#2/RNASeq:Sample@sort(Last_Modified_Timestamp::desc::,Accession_ID))

### DIFF
--- a/common/templates/ellipses.html
+++ b/common/templates/ellipses.html
@@ -10,7 +10,7 @@
             <span class="glyphicon glyphicon-remove-circle"></span>
         </button>
         <button ng-if="config.deletable && delete && tuple.reference.canDelete" type="button" class="delete-action-button btn btn-xs btn-primary btn-no-focus" ng-click="delete()" tooltip-placement="bottom" uib-tooltip="Delete">
-            <span class="glyphicon glyphicon-trash" style="color:white"></span>
+            <span class="glyphicon glyphicon-trash"></span>
         </button>
         <button ng-if="config.selectMode == 'single-select'" type="button" class="select-action-button btn btn-xs btn-primary btn-no-focus" ng-click="onSelect()" tooltip-placement="bottom" uib-tooltip="Select" >
             <span class="glyphicon glyphicon-check"></span>
--- a/common/templates/ellipses.html
+++ b/common/templates/ellipses.html
@@ -1,18 +1,18 @@
 <td ng-if="config.viewable || config.editable || config.deletable || config.selectMode != 'no-select'" class="block actions-layout">
     <div class="btn-group" style="display: flex">
-        <a ng-if="config.viewable" type="button" class="view-action-button btn btn-xs btn-default btn-no-focus" ng-href="{{viewLink}}" tooltip-placement="bottom" uib-tooltip="View">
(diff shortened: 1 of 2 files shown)


common/templates/ellipses.html
```diff
@@ -1,18 +1,18 @@
...
         </button>
-        <button ng-if="config.deletable && unlink && associationRef.canDelete" type="button" class="delete-action-button btn btn-xs btn-default btn-no-focus" ng-click="unlink()" tooltip-placement="bottom" uib-tooltip="Unlink">
+        <button ng-if="config.deletable && unlink && associationRef.canDelete" type="button" class="delete-action-button btn btn-xs btn-primary btn-no-focus" ng-click="unlink()" tooltip-placement="bottom" uib-tooltip="Unlink">
             <span class="glyphicon glyphicon-remove-circle"></span>
         </button>
-        <button ng-if="config.deletable && delete && tuple.reference.canDelete" type="button" class="delete-action-button btn btn-xs btn-default btn-no-focus" ng-click="delete()" tooltip-placement="bottom" uib-tooltip="Delete">
-            <span class="glyphicon glyphicon-trash" style="color: black"></span>
+        <button ng-if="config.deletable && delete && tuple.reference.canDelete" type="button" class="delete-action-button btn btn-xs btn-primary btn-no-focus" ng-click="delete()" tooltip-placement="bottom" uib-tooltip="Delete">
+            <span class="glyphicon glyphicon-trash" style="color:white"></span>
```
amitjha21: `.glyphicon-trash {
    color: #428bca;
}`

`button .glyphicon-trash {
    color: #fff;
}`

**These classes are maintained in `app.css`.**

jrchudy: Ah thank you. So in the case that `glyphicon-trash` is **NOT** being used in a button, we are changing it to blue. And then when we want to use it as a button, we change it back to white, but that's assuming the button is a color where white would show _well_.

We should adjust `.glyphicon-trash { color: #428bca; }` to be a styling on a class we created like `glyphicon-trash-link` or change it to `a .glyphicon-trash { color: #428bca; }`. Look at `delete-link.html`. And then we should remove `button .glyphicon-trash { color: #fff; }` from `app.css`.

jrchudy: And this blue color can be easily seen in the bar below the navbar on the record page.

jrchudy:
@amitjha21:

Ah thank you. So in the case that `glyphicon-trash` is **NOT** being used in a button, we are changing it to blue. And then when we want to use it as a button, we change it back to white, but that's assuming the button is a color where white would show _well_.

We should adjust `.glyphicon-trash { color: #428bca; }` to be a styling on a class we created like `glyphicon-trash-link` or change it to `a .glyphicon-trash { color: #428bca; }`. Look at `delete-link.html`. And then we should remove `button .glyphicon-trash { color: #fff; }` from `app.css`.

jrchudy:
Not sure if you saw that comment chained on to your previous one because of it being collapsed.

amitjha21:
Added `a` tag for class and default `#fff` has been removed.

Issue #1234, opened 2017-08-10
Handle base type in input display mode for recordedit


jrchudy:
The `columnToDisplayType` function in `recordedit/form.controller.js` needs to be aware that if we don't know how to handle the type of a column, we should check it's base type and see if we can display that. This is so we can support user defined domain types better.

jrchudy:
Closing this issue because the corresponding PR has been merged into master.

PR #1236, opened 2017-08-11
Recordedit input basetype

jrchudy:
This PR adds functionality to the `recordedit` app to check for a column's base type if we don't know how to create an input for it's type if it's a user defined domain type.

This is for issue #1234.


jrchudy:
I created a new catalog and schema on `dev.isrd`, added a domain type to that schema that `recordedit` did not understand. That domain type had a base type of date. I was then able to create a table with a column using that new domain type. When that column was displayed in the `recordedit` form, it used the `date` input we have defined instead of the text input.

Issue #806, opened 2016-11-30
create reusable directive for a number input


hongsudt:
This directive can be used in the recordedit and faceting. 
- integer
- float

jrchudy:
While writing this directive, I realized that float and integer can't be handled in the same directive. Float will require a separate directive for showing properly.

jrchudy:
Moving this issue to round 2.

RFSH:
An [input-switch](https://github.com/informatics-isi-edu/chaise/blob/master/common/inputs.js#L360) has been recently implemented that will create an input for any type of columns. This includes numbers too.

PR #1240, opened 2017-08-16
Integer input directive

jrchudy:
This PR addresses issue #806.


Makefile
```diff
@@ -134,33 +134,34 @@ JS_DEPS=$(JS)/vendor/jquery-latest.min.js \
...
-	$(JS)/facetsModel.js \
-	$(JS)/facetsService.js \
 	$(JS)/controller/ermrestDetailController.js \
 	$(JS)/controller/ermrestFilterController.js \
 	$(JS)/controller/ermrestInitController.js \
 	$(JS)/controller/ermrestLoginController.js \
 	$(JS)/controller/ermrestResultsController.js \
 	$(JS)/controller/ermrestSideBarController.js \
 	$(JS)/controller/ermrestTourController.js \
+	$(JS)/ermrest.js \
```
RFSH: It seems like you have changed the order of the resources. I'm just curious, was that necessary? or you just wanted to organize them?

common/inputs.js
```diff
@@ -0,0 +1,56 @@
...
+                        scope.max = 9223372036854775807;
+                        break;
+                    default:
+                        break;
+                }
+
+                if (!ngModel) return;
+
+                scope.onChange = function(){
+                    ngModel.$setViewValue(scope.value);
```
RFSH: I can see that at some point we might want to call a function on input change. We can pass a callback to the directive which here will be called.

Makefile
jrchudy: I was organizing them better and didn't realize I changed the `JS_RESOURCE` at first. The order does matter for some of the resources though (jquery needs to be defined first before angular for instance).

common/inputs.js
jrchudy: I agree. I will make this a callback.

jrchudy:
This PR is no longer relevant.

PR #1241, opened 2017-08-16
Association table refresh in markdown display

nitishksahu:
When we add a relationship to an association table, the heading of the table (showing n results) gets updated but the related entities are not updated.

A 2-way binding has been put in place to resolve this issue.


RFSH:
Can you add test cases for this? We have a test case for tabular mode, but not markdown mode.

test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -532,5 +533,68 @@ describe('View existing record,', function() {
...
+            beforeAll(function() {
+                var keys = [];
+                keys.push(testParams.key.name + testParams.key.operator + testParams.key.value);
+                browser.ignoreSynchronization=true;
+                var url = browser.params.url + "/record/#" + browser.params.catalogId + "/" + testParams.schemaName + ":" + testParams.table_name + "/" + keys.join("&");
+                browser.get(url);
+                chaisePage.waitForElement(element(by.id('rt-heading-association_table_markdown')));
+            });
+            
+            it("on adding new relationship should update the table display", function (){
```
RFSH: Since you have multiple `expect` statements inside this `it`, can you please add error message to each of them?

Issue #266, opened 2016-03-29
Consolidate `index.html.in` templates


robes:
We only need one `index.html.in` template. We can inject assets and templates to produce each page based on `make` targets. We should factor out the title into a configurable setting and inject it into the page with angular, but that can be a separate issue.

robes:
@hongsudt This is another one of those makefile & related issues.

hongsudt:
I am moving this back to backlog..

jrchudy:
Closing and moving this to wishlist.

PR #1250, opened 2017-08-29
Added a way to reuse index.html for all other pages

chiragsanghvi:
This PR fixes issue https://github.com/informatics-isi-edu/chaise/issues/266 by consolidating `index.html.in` templates into 1 for reuse. across different apps 

https://dev.isrd.isi.edu/~chirag/chaise/recordset/#1/isa:dataset


chiragsanghvi:
The build is passing. Can I merge it.

Makefile
```diff
@@ -577,40 +577,82 @@ distclean: clean
 html: $(HTML)
 
 # Rules to attach JavaScript and CSS assets to the head
-search/index.html: search/index.html.in .make-asset-block .make-template-block
+search/index.html: .make-asset-block .make-template-block
```
robes: `search/search.html.in` should be a dependency, right?

Makefile
```diff
@@ -577,40 +577,82 @@ distclean: clean
...
 	sed -e '/%ASSETS%/ {' -e 'r .make-asset-block' -e 'd' -e '}' \
 		-e '/%TEMPLATES%/ {' -e 'r .make-template-block' -e 'd' -e '}' \
-		search/index.html.in > search/index.html
+		search/search.html.in > search/index.html
 
-login/index.html: login/index.html.in .make-asset-block
+login/index.html: .make-asset-block
```
robes: and `index.html.in` should be a dependency here and throughout the rest of the rules, right?

Makefile
```diff
@@ -577,40 +577,82 @@ distclean: clean
...
-detailed/index.html: detailed/index.html.in .make-detailed-asset-block .make-detailed-template-block
+		-e 's/%APPNAME%//' \
+		-e 's/%PAGE%/Chaise :: Login/' \
+		-e 's/%NAVBAR%//' \
+		-e 's/%FOOTER%//' \
+		-e 's/%TEMPLATES%//' \
+		-e '/%CONTENT%/ {' -e 'r login/login.html.in' -e 'd' -e '}' \
+		index.html.in > login/index.html
+
+viewer/index.html: .make-viewer-asset-block
```
robes: these blocks all repeat... seems like they could be collapsed into a single pattern based rule?

chiragsanghvi:
@robes Should I merge this. I have made the changes as you requested.

robes:
Hmmm, I don't know. This seems more complicated. Like I commented before, I was thinking that a "pattern rule" might simplify this. Meaning you just need one rule something like:

```
%/index.html: index.html.in .make-%-asset-template
    sed -e '/%ASSETS%/ {' -e 'r .make-$*-asset-block' -e 'd' -e '}' \
        -e '/%TEMPLATES%/ {' -e 'r .make-$*-template-block' -e 'd' -e '}' \
        index.html.in > $*/index.html
```

This is what I had in mind with my earlier suggestion. I personally would find this change to the defined sequence `generateHtml` hard to follow and maintain.

hongsudt:
@chiragsanghvi have you addressed @robes's coments? If so, then please resolve the conflict and merge the code.

hongsudt:
@chiragsanghvi can you resolve this with @robes and finish it?

RFSH:
Since this PR was created we changed the structure of our apps and therefore this PR is not valid anymore. I am going to close this.
I am going to keep the issue that is connected to this PR open, so we can discuss later whether it's worth doing this with the new structure.

Issue #1231, opened 2017-08-07
support adding a batch of pure and binary association entries 


hongsudt:
For pure and association entries, when a user click "add", it should open up a multi-select modal and allow users to select a set of entities to be added at once. This feature will be useful for adding members to a collection.. 

We will probably have to deal with users choosing a subset of entities that have been previously added to the table.

amitjha21:
Steps:

1. When `Add` button is clicked of pure and association related table it should open a modal pop up.
2. Modal pop up will display table with all permitted domain value for the field.
3. Multi-select checkbox should appear in the left most column in pop-up table .
4. Modal would be modified to accommodate **Submit** button (Existing `modal` directive to be modified).
5. After clicking **Submit** button pop-up disappears and records are added in the table in the `record` page.

**Note: User Control never goes out of `record` page.**

### Example of Modal pop-up
![image](https://user-images.githubusercontent.com/9977014/29379494-8ba34436-8277-11e7-8d51-c43eb037e839.png)

amitjha21:
Few suggestions:

1. Avoid the UI shifting when selecting your first pick. Maybe put a placeholder at the top so that the page doesn't shift.
2. after submitting the main record page's nested entity does not reflect the change.
3. If you pick something that has already been picked you get a 409 : This needs more discussion on how to fetch only permissible domain values. 

Future Changes:
1. Having pre-checked permissible values in the modal box.

amitjha21:
Refresh issue for added record fixed. (_Bullet 2 in above comment_)

amitjha21:
This has been merged. Assigning to @hongsudt to verify and close it.

PR #1251, opened 2017-09-02
Batch add association

amitjha21:
This PR addresses  #1231 
It adds popup in `record` app  to add record without leaving the app for pure and binary association entries.


common/recordCreate.js
```diff
@@ -0,0 +1,300 @@
...
+                     * After uploading files, the returned submissionRowsCopy contains
+                     * new file data. This includes filename, filebyte, and md5.
+                     * The following makes sure that all the data are updated.
+                     * That's why this for loop must be after uploading files and not before.
+                     * And we cannot just pass submissionRowsCopy to update function, because
+                     * update function only accepts array of tuples (and not just key-value pair).
+                     */
+                    for (var i = 0; i < submissionRowsCopy.length; i++) {
+                        var row = submissionRowsCopy[i];
+                        var data = $rootScope.tuples[i].data;
```
jrchudy: Same comment as above about passing `tuples` in as a parameter and making sure it's not on the model beforehand.

common/recordCreate.js
```diff
@@ -0,0 +1,300 @@
...
+                        // assign each value from the form to the data object on tuple
+                        for (var key in row) {
+                            data[key] = (row[key] === '' ? null : row[key]);
+                        }
+                    }
+
+                    // submit $rootScope.tuples because we are changing and
+                    // comparing data from the old data set for the tuple with the updated data set from the UI
+                    fn = "update", fnScope = $rootScope.reference, args = [$rootScope.tuples];
```
jrchudy: Same comment about passing `reference` and `tuples` as a parameter.

common/recordCreate.js
```diff
@@ -0,0 +1,300 @@
...
+
+                    var page = result.successful;
+                    var failedPage = result.failed;
+
+                    // the returned reference is contextualized and we don't need to contextualize it again
+                    var resultsReference = page.reference;
+                    if (isUpdate) {
+                        for (var i = 0; i < submissionRowsCopy.length; i++) {
+                            var row = submissionRowsCopy[i];
+                            var data = $rootScope.tuples[i].data;
```
jrchudy: Same comment about `tuples`.

common/recordCreate.js
```diff
@@ -0,0 +1,300 @@
...
+                }
+
+                fnScope[fn].apply(fnScope, args).then(function success(result) {
+
+                    var page = result.successful;
+                    var failedPage = result.failed;
+
+                    // the returned reference is contextualized and we don't need to contextualize it again
+                    var resultsReference = page.reference;
+                    if (isUpdate) {
```
jrchudy: This if condition is doing almost the same exact thing as L100-L107. This loop that looks through `submissionRowsCopy` should be factored out as a function that's used in both places.

common/recordCreate.js
```diff
@@ -0,0 +1,300 @@
...
+         */
+        var addPopup = function(ref, rowIndex, derivedref, isModalUpdate) {
+            var column = ref;
+
+            var originalTuple, nullArr = [],
+                editOrCopy = true,
+                params = {};
+
+            if (viewModel.editMode) {
+                originalTuple = $rootScope.tuples[rowIndex];
```
jrchudy: `Tupels` needs to be passed in as a parameter.

recordedit/form.controller.js
```diff
@@ -71,7 +71,9 @@
         vm.prefillCookie = $cookies.getObject(context.queryParams.prefill);
         vm.makeSafeIdAttr = DataUtils.makeSafeIdAttr;
 
-
+        this.testShare = function(){
```
jrchudy: I don't think we need this function.

common/modal.js
```diff
@@ -66,11 +66,13 @@
         vm.params = params;
         vm.ok = ok;
         vm.cancel = cancel;
+        vm.submit = submitMutliSelection;
 
         var reference = params.reference;
         vm.hasLoaded = false;
         var reference = vm.reference = params.reference;
-
+        var selectMode = params.selectMode?params.selectMode:"single-select";
```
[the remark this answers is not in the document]
amitjha21: This has been defaulted to *single-select* to support backward compatibility. (Any existing piece of code which does not leverage this attribute). It is, in my opinion, a fair assumption to have single-select in case no value is passed. 
I will move this string value to a `constant` module.

common/modal.js
```diff
@@ -60,16 +60,18 @@
...
 
         vm.params = params;
         vm.ok = ok;
         vm.cancel = cancel;
+        vm.submit = submitMutliSelection;
 
         var reference = params.reference;
         vm.hasLoaded = false;
         var reference = vm.reference = params.reference;
+        var selectMode = params.selectMode?params.selectMode:modalBox.singleSelectMode;
```
jrchudy: Again, we should make sure this property is required for every popup select modal rather than having conditional logic for it.

We don't need to worry about backwards compatibility for this feature because only we are leveraging it and can make the proper changes in our current codebase to account for it being required.

common/recordCreate.js
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        /**
+         * uploadFiles - uploading files
+         *
+         * @param  {array} submissionRowsCopy   data that is going to be uploaded
+         * @param  {object} rootScope           object contains global values from calling function
+         * @param  {object} onSuccess           callback
+         */
+        function uploadFiles(submissionRowsCopy, rootScope, onSuccess) {
```
jrchudy: Let me be more explicit, the reference should be passed into this function not rootscope. We should not pass the entire rootscope into this function, there are far more functions and properties on the rootscope that are not needed.

common/recordCreate.js
```diff
@@ -0,0 +1,308 @@
...
+         * addRecords - Function that calls ermrestjs method to add data
+         *
+         * @param  {bool} isUpdate              flag
+         * @param  {object} derivedref          derived referene of the column that is being added
+         * @param  {array} recordEditModel      data array
+         * @param  {bool} isModalUpdate         if updating through record app
+         * @param  {object} rootScope           object contains global values from calling function
+         * @param  {object} onSuccessFunction   callback
+         */
+        function addRecords(isUpdate, derivedref, recordEditModel, isModalUpdate, rootScope, onSuccessFunction) {
```
jrchudy: Same comment here as above. The whole rootscope is not needed, only the `tuples` and `reference`. Only pass those into this function as separate parameters.

Make sure these 2 values are not already present on `recordEditModel` though.

jrchudy: I guess I didn't notice that the `queryParams` were being used as well (`rootscope.context.queryParams`). This should be another parameter to the `addRecords` function.

common/recordCreate.js
```diff
@@ -0,0 +1,308 @@
...
+                    args = [submissionRowsCopy];
+                var fnScope = isModalUpdate ? derivedref.unfilteredReference.contextualize.entryCreate : rootScope.reference.unfilteredReference.contextualize.entryCreate;
+
+                if (isUpdate) {
+
+                    var data = checkUpdate(submissionRowsCopy, rootScope);
+
+                    // submit rootScope.tuples because we are changing and
+                    // comparing data from the old data set for the tuple with the updated data set from the UI
+                    fn = "update", fnScope = rootScope.reference, args = [rootScope.tuples];
```
jrchudy: Is tuples already an array? I'm curious why it needs to be defined in brackets here?

jrchudy:
After running the tests locally, I can verify that the errors shown in travis are caused by your changes, and that those errors need to be addressed in this PR before it can be merged.

amitjha21:
5c672ca includes:
The behavior of `record` and `recordedit` apps after `Submit`. In `recordedit` app if one row has inserted the page redirects to `resultset`. In case of multi-edit, it created a `header` tag with no of records created.
For `record` app it stays on the same page with a message appearing on the top.
Test case failure was attributed to this. This commit should take care of them. I am running test to check the outcome.

amitjha21:
All cases passed @ travis.

jrchudy:
I will review this tomorrow morning along with PR #1261.

Issue #1255, opened 2017-09-07
The default display of inline inbound fk should be a list


hongsudt:
Currently, a tabular format is displayed instead of a list of row names. Please fix asap. 
https://dev.rebuildingakidney.org/chaise/record/#2/RNASeq:Sample/Internal_ID=d18_batch4

Also when it is in a tabular format, the position of Add | View More should be on top of the table, not next to the table.

amitjha21:
There is no accordion for tables that are appearing as part of the main entity; if `Add | View More` comes on top of the table it will create extra white space on the top. To utilize this space we had decided to add all action items next to the table.

amitjha21:
1. Change action items position: Keep the items on top of the table. It is ok to have extra white spaces.
2. If no annotation is defined, default display should be bullets of row names. Row names should be available through `ermrestjs`.

amitjha21:
If no annotation is defined then `ermrestjs` returns `table` type as default display type.

amitjha21:
This is merged now. Assigning to @hongsudt to verify it.

PR #1261, opened 2017-09-12
Inbfk alignment

amitjha21:
This PR address #1255 .
This has a dependent PR in `ermrestjs` which deals with creation of `compact/brief/inline` context and rendering custom list in case no annotation is defined.


amitjha21:
Investigating test cases failure.

amitjha21:
Merge conflicts were removed after `CHAISE` was updated with faceting code.

test/e2e/specs/all-features-confirmation/record/delete-btn.spec.js
```diff
@@ -101,8 +101,9 @@ describe('View existing record,', function() {
                 chaisePage.waitForElement(element(by.id('tblRecord'))).then(function() {
                     return browser.executeScript('return chaiseConfig;');
                 }).then(function(chaiseConfig) {
-                    config = chaiseConfig
-                    return chaisePage.recordPage.getDeleteRecordButton().click()
+                    config = chaiseConfig;
+                    browser.sleep(1000);
```
jrchudy: Maybe we want to wait for the delete button to be clickable using a `browser.wait()` then click the button.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -114,16 +114,20 @@ describe('View existing record,', function() {
 
             browser.wait(function() {
                 return chaisePage.recordPage.getRelatedTablesWithPanel().count().then(function(ct) {
+                    console.log("ct: "+ct);
```
jrchudy: Please remove.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -114,16 +114,20 @@ describe('View existing record,', function() {
...
                     return (ct=testParams.no_related_data.tables_order.length);
                 });
             }, browser.params.defaultTimeout);
             var showAllRTButton = chaisePage.recordPage.getShowAllRelatedEntitiesButton();
 
             chaisePage.recordPage.getRelatedTablesWithPanel().count().then(function(count) {
+                browser.sleep(2000);
+                console.log(count);
```
jrchudy: please remove

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -114,16 +114,20 @@ describe('View existing record,', function() {
...
             browser.wait(function() {
                 return chaisePage.recordPage.getRelatedTablesWithPanel().count().then(function(ct) {
+                    console.log("ct: "+ct);
                     return (ct=testParams.no_related_data.tables_order.length);
                 });
             }, browser.params.defaultTimeout);
             var showAllRTButton = chaisePage.recordPage.getShowAllRelatedEntitiesButton();
 
             chaisePage.recordPage.getRelatedTablesWithPanel().count().then(function(count) {
+                browser.sleep(2000);
```
jrchudy: Maybe wait for the `loading...` element to not be visible anymore? That element should show up at the very bottom of the page and disappear once all related tables are loaded. But this should be done before you even try to get their count. a sleep here shouldn't make any difference at all since it's in the promise after the value was already fetched.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -114,16 +114,20 @@ describe('View existing record,', function() {
...
+                console.log(count);
                 expect(count).toBe(testParams.no_related_data.tables_order.length, "Number of related tables is not correct");
 
                 return chaisePage.recordPage.getRelatedTableTitles();
             }).then(function(headings) {
+                browser.sleep(2000);
```
jrchudy: A sleep statement here won't make a difference. The `headings` have already been fetched. This sleep is just delaying the check against the already available `headings` value.

If the problem is that headings aren't always correct, maybe we need to wait for all the related tables to be loaded like I stated above.

test/e2e/specs/default-config/recordedit/domain-filter.spec.js
```diff
@@ -170,15 +170,15 @@ describe("Add a record,", function() {
...
                     chaisePage.recordEditPage.getForeignKeyInputButton(constraintColName, 0).click().then(function () {
                         browser.wait(EC.visibilityOf(modalTitle), browser.params.defaultTimeout);
-
                         return modalTitle.getText();
                     }).then(function(text) {
                         expect(text.indexOf("Choose")).toBeGreaterThan(-1);
 
                         rows = chaisePage.recordsetPage.getRows();
-
+                        browser.sleep(3000);
```
jrchudy: This sleep also should not make a difference since you are fetching the rows before this sleep statement. If the problem is that the rows aren't completely proper because they haven't all loaded yet, we should have a `browser.wait()` before fetching the rows.

test/e2e/utils/record-helpers.js
```diff
@@ -58,6 +58,8 @@ exports.testPresentation = function (tableParams) {
 	it("should render columns which are specified to be visible and in order", function() {
 		var columns = tableParams.columns.filter(function(c) {return c.value != null;});
 		chaisePage.recordPage.getAllColumnCaptions().then(function(pageColumns) {
+			browser.sleep(2000);
+			browser.pause();
```
jrchudy: remove this

test/e2e/utils/record-helpers.js
```diff
@@ -58,6 +58,8 @@ exports.testPresentation = function (tableParams) {
 	it("should render columns which are specified to be visible and in order", function() {
 		var columns = tableParams.columns.filter(function(c) {return c.value != null;});
 		chaisePage.recordPage.getAllColumnCaptions().then(function(pageColumns) {
+			browser.sleep(2000);
```
jrchudy: This sleep shouldn't be here and doesn't appear to be making a difference in the tests either.

PR #1263, opened 2017-09-13
Profile card for chaise

nitishksahu:
- A dropdown is added to the profile name that's displayed on the navbar.

- Upon clicking the profile name, a profile card appears with all the other information found in 
   webauthn's client object.

- The card has all the group names, the user is a part of, which is clickable and will navigate to 
   globus group detail page when clicked.

### DIFF
--- a/common/templates/navbar.html
+++ b/common/templates/navbar.html
@@ -13,18 +13,27 @@
                     <span id="brand-text" ng-bind="brandText || 'Chaise'"></span>
                 </a>
             </div>
-
+            
             <!-- Collect the nav links, forms, and other content for toggling -->
             <div class="collapse navbar-collapse navbar-inverse" id="fb-navbar-main-collapse">
                 <ul class="nav navbar-nav navbar-right">
                     <li>
                         <a id="signup-link" ng-if="user === null && signUpURL" ng-href="{{signUpURL}}">Sign Up</a>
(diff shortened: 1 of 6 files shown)


jrchudy:
Where are the tests for this feature?

common/templates/navbar.html
```diff
@@ -40,7 +49,10 @@
...
                     </li>
                 </ul>
-            </div><!--/.nav-collapse -->
+            </div><!--/.nav-collapse -->    
         </div>
     </nav>
 </header>
+
+<!--Modal Window -->
+<div profilemodal id="myModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true" class="modal fade" ></div>
```
jrchudy: Why does this modal element need to be defined here? 

In all other cases we use the modal popup, the modal is a template that is rendered into the DOM with only having to specify it as part of the `$uibModal.open` function.

jrchudy: In my local code, removing this element from the DOM doesn't appear to change any of the functionality.

nitishksahu:
@jrchudy the requested changes have been done and the PR is now available for review.

nitishksahu:
The changes have been merged to master, assigning this to @hongsudt to verify and close.

PR #1315, opened 2017-09-27
Disable login button after it is clicked

nitishksahu:
When the button is clicked twice, a terminal error is generated. Code changes have been done to disable the `Login` button when clicked once.


RFSH:
But this means if you click on the login button and didn't actually login, you cannot login anymore until you reload the page. Also I cannot see the terminal error.

Issue #1320, opened 2017-09-28
Range Picker min > max validator


jrchudy:
When the min value in the range picker is greater than or equal to the max value, a validation error should be thrown.

Issue #1319, opened 2017-09-28
Range Inputs need better validators


jrchudy:
The integer and float inputs need to have separate validators defined for them.

PR #1326, opened 2017-10-05
Range inputs validators

jrchudy:
This PR addresses issues #1319 and #1320.


jrchudy:
@RFSH, this PR is ready for review. In it I addressed:
 - min/max validations for integers
 - min > max for each input type
 - float properly errors when an improper decimal is entered (`<input type="text">` instead of "number" fixed that. Will have to do the same for `recordedit`)
 - when integers are too large/small, the user can press the down arrow and it changes the value into scientific notation. So I disabled the default behavior for up/down arrow actions on the integer inputs

PR #1327, opened 2017-10-06
403 error code now handled in catch all

nitishksahu:
A terminal error was shown on recordedit when users were not authorized to edit the record. A modal window is shown now with proper error message.


common/utils.js
```diff
@@ -47,7 +47,9 @@
             title: "Your need to be logged in to continue.",
             message: "To open the login window press"
         },
-        "tableMissing": "No table specified in the form of 'schema-name:table-name' and no Default is set."
+        "tableMissing": "No table specified in the form of 'schema-name:table-name' and no Default is set.",
+        "unauthorizedMessage" : "You are not authorized to Update entities.Please report this problem to your system administrators.",
```
RFSH: Since this is going to be a general message for any 403 errors, I think we should say something in line of "You are not authorized to perform this action." Instead of being too specific and saying "Update entities".

nitishksahu: Error message has been changed

common/errors.js
```diff
@@ -142,12 +142,16 @@
...
-
+            
             if (ERMrest && exception instanceof ERMrest.UnauthorizedError || exception.code == errorNames.unauthorized) {
                 Session.loginInAModal(reloadCb);
             }
             else if (exception.code && exception.code == errorNames.multipleRecords){
                 errorPopup(messageMap.multipleDataMessage, messageMap.multipleDataErrorCode,"Recordset ", exception.redirectUrl);
+            }
+            
+            else if (ERMrest && exception instanceof ERMrest.UnauthorizedError || exception.code == errorNames.forbidden) {
```
jrchudy: Can you change this to have some parenthesis to group the logic statements together? Right now it evaluates from left to right which is probably correct, but it's easier to understand with parenthetic grouping.

jrchudy: I think it's supposed to be read as:
```
( (ERMrest && exception instanceof ERMrest.UnauthorizedError) || exception.code == errorNames.forbidden )
```

common/errors.js
```diff
@@ -142,12 +142,16 @@
                 window.location.reload();
             }; 
             if (exceptionFlag || window.location.pathname.indexOf('/search/') != -1 || window.location.pathname.indexOf('/viewer/') != -1) return;
-
+            
             if (ERMrest && exception instanceof ERMrest.UnauthorizedError || exception.code == errorNames.unauthorized) {
```
jrchudy: While you are making changes here in the `handleException` function, can you change this line like the one you added with better parenthetic grouping?

jrchudy: Should read something like:
```
( (ERMrest && exception instanceof ERMrest.UnauthorizedError) || exception.code == errorNames.unauthorized )
```

recordedit/recordEdit.app.js
```diff
@@ -263,14 +263,12 @@
                             throw response;
                         });
                     } else if (session) {
-                        var notAuthorizedMessage = "You are not authorized to Update entities.";
                         var notAuthorizedError = new Error(notAuthorizedMessage);
```
RFSH: You removed the `notAuthorizedMessage`, but it's still being used in this line. The following is not really related to your code, but since you're changing this piece of code it's better if you refactor it a bit.

Instead of creating a javascript `Error` object and manually changing the `code` attribute. You can use ERMrestjs. You can just say `new ERMrest.UnauthorizedError()` or `new ERMrest.ForbiddenError()`. You need to pass the `status`, and `message` to this function, but since we're not showing those attributes it might not be necessary. 

So please change all these error creations in this file to use ermrestjs error classes. You also missed the [error cases for create](https://github.com/informatics-isi-edu/chaise/blob/5e4ccaaf4ffb7baa0a34f8126d1d62f0528e3ccf/recordedit/recordEdit.app.js#L333,L346).

jrchudy: Since you removed the variable above, this case will fail/error out with no message defined.

recordedit/recordEdit.app.js
```diff
@@ -263,14 +263,12 @@
...
                         // user logged in but not allowed (forbidden)
                         notAuthorizedError.code = errorNames.forbidden;
 
                         throw notAuthorizedError;
                     } else {
-                        var notAuthorizedMessage = "You are not authorized to Update entities.";
```
jrchudy: Since you removed the variable above, this case will fail/error out with no message defined.

common/errors.js
RFSH: If you refactor those errors to use ERMrestjs, then this check for `expection.code == errorNames.forbidden` won't be necessary. Although some other piece of code might still be using this error name. So you can just add a TODO that we should remove it when we get to refactor error classes in chaise.

recordedit/recordEdit.app.js
nitishksahu: I added the message to the message map, which will solve this for this case

https://github.com/informatics-isi-edu/chaise/blob/issue%231223/common/utils.js/#L51

recordedit/recordEdit.app.js
nitishksahu: I have added the error message code back for this case

nitishksahu:
Changes have been made to use ermrest error objects rather than javascript error objects, the test cases also passed in my local machine. @RFSH  @jrchudy the PR is available for review.

recordedit/recordEdit.app.js
jrchudy: Maybe you misunderstood what @RFSH had said. The errors in `ermrestJS` are ways of wrapping JS errors, but you have to still provide the status and message that will be displayed. See [`errors.js`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/errors.js) or the [`.doc`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/doc/api.md#ERMrest.TimedOutError).

You need to be cautious though. In the case of Unauthorized (401), we use an unauthorized message. In the case of Forbidden (403), we still use an unauthorized message and status. We don't want to tell the user they are "forbidden" instead we use "unauthorized" to notify them in both cases.

We still want to use the proper error type (Unauthorized or Forbidden) but want the message to be the same in both cases.

nitishksahu:
The changes have been made to pass the error message and error status to the ermrest error object constructor. @jrchudy @RFSH The PR is available for review.

recordedit/recordEdit.app.js
```diff
@@ -332,17 +330,12 @@
                         $rootScope.displayReady = true;
                         // if there is a session, user isn't allowed to create
                     } else if (session) {
-                        var forbiddenMessage = "You are not authorized to Create entities.";
-                        var forbiddenError = new Error(forbiddenMessage);
-
+                        var forbiddenError = new ERMrest.ForbiddenError(messageMap.unauthorizedErrorCode, messageMap.unauthorizedMessage);
                         forbiddenError.code = errorNames.forbidden;
```
RFSH: You should remove these lines that we manually are changing the error code. The `ERMrest.ForbiddenError` constructor should take care of that for us. So we don't need any of the `forbiddenError.code = errorNames.forbidden;` and `notAuthorizedError.code = errorNames.unauthorized;` that are in this file.

nitishksahu: The requested changes are now done.

Issue #1253, opened 2017-09-05
Incorrect display of immutable asset column in the edit context


hongsudt:
In the edit mode, if the file url exist, it should be shown under the uri column. On GPCR, the current uri (though exists) is not shown. 
https://staging.gpcrconsortium.org/chaise/recordedit/#1/assets:cpm/sha256=00220d82f5b98a3f043eeac13b027b33a23dc9ad5ad597621fda16b064852103@sort(last_modified_timestamp::desc::,id)

Also, the (*) is not shown even though the uri column is NOT NULL. 

Note: I only see this issue on gpcr and not on RBK.  https://dev.rebuildingakidney.org/chaise/recordedit/#2/Protocol:Protocol/ID=2;ID=5@sort(ID)

jrchudy:
After debugging the root cause of this issue, I noticed that the underlying column in GPCR has the `immutable` annotation. 

This uncovered some other bugs in `recordedit` though. 
 - The `*` for required is not showing up properly
 - The input field is not disabled properly (button as well)
 - The input field does not show the original value properly

PR #1367, opened 2017-11-09
Immutable upload file input

jrchudy:
This PR addresses issue #1253. The upload input was **NOT** being properly disabled in `recordedit` when the column had the immutable annotation defined. The input was also not having it's previous value showing properly in the input in this case.


jrchudy:
@RFSH I added tests for disabled inputs in "edit" mode of `recordedit`. The spec I added those tests to is not the "best" place but does reduce the runtime of our tests by not creating a new spec. I can rename the spec to be `recordedit-disabled-inputs` instead of `recordadd-defaults`.

test/e2e/specs/default-config/recordedit/add-defaults.spec.js
```diff
@@ -200,3 +216,48 @@ describe('Record Add with defaults', function() {
...
+            keys.push(testParams.edit_key.name + testParams.edit_key.operator + testParams.edit_key.value);
+            browser.ignoreSynchronization=true;
+            browser.get(browser.params.url + "/recordedit/#" + browser.params.catalogId + "/defaults:" + testParams.table_name + "/"  + keys.join("&"));
+
+            chaisePage.waitForElement(element(by.id("submit-record-button")));
+        });
+
+        for (var i=0; i < testParams.re_column_names.length; i++) {
+            (function(columnName) {
+                if (columnName == "foreign_key_disabled") {
```
RFSH: I think using switch case is better than if else here.

jrchudy: 👍

jrchudy:
@RFSH I made the changes to rename the files and add the switch statement instead of the lengthy `if/else if/else` case. I will merge this once my other branch for the ACLs changes passes and is merged to master.

Issue #1383, opened 2017-11-21
fix position of the `loading` in record app


hongsudt:
Currently, the word is on top of the main entity table.

Issue #1381, opened 2017-11-17
Browser Print 


jrchudy:
We haven't added anything to our code to specifically handle using the browser print functionality. Recent changes broke this function so that it only prints the first page of content when there are multiple pages. This is especially bad in the case of `record`.

For `record` app, we need to add css rules under the `@media print {...}` query. This way we can manage some of the html and css when printing out the html. A few simple rules to add:
```
#main-content {
  height: auto !important;
}
    
#footer {
  display: none;
}
```

These 2 rules will remove the fixed height and remove the footer. This will fix the problem with the all the pages not printing. There are some other issues with the printed content as well that can hopefully be addressed.
 - remove the "urls" that are being printed next to the buttons and links with `href` attributes defined
 - reduce the spacing between the column names and column values by changing the `min-width` property and making sure column names still properly wrap the text.

PR #1384, opened 2017-11-21
Media print rules added in CSS

jrchudy:
This PR addresses issues #1381 and #1383.

`@media print {...}` was added to change some of the styling when print is used from the browser tools. This should make printing documents look better than before and remove the limitation that bound it to just the first page. 

Extra padding was added to the main container so that the `Loading...` text shows up in a better place now.

### DIFF
--- a/common/styles/app.css
+++ b/common/styles/app.css
@@ -1444,7 +1444,7 @@ input[type=number]::-moz-placeholder {
 
     .entity-key {
         width: auto;
-        max-width: 100px;
+        max-width: 120px;
     }
 
     .entity-value {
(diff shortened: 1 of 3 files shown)


RFSH:
Is there a reason for `#rt-loading` span to be outside the `main-container`? You can just move it up inside the `main-container` and then remove the extra padding that you added.

jrchudy:
@RFSH I moved the "Loading..." div to inside the main container. This extended the space of that container and worked as you suggested, but the text was cut off by the footer so I added another `10 px` of padding.

Issue #1287, opened 2017-09-14
Make Sure Chaise Apps Work With Facet And Filter


jrchudy:
To do this, we need the following set of changes for each app:

- Record: 
  - We are using filters for error purposes in this app [[1](https://github.com/informatics-isi-edu/chaise/blob/master/record/record.app.js#L140),  [2](https://github.com/informatics-isi-edu/chaise/blob/master/record/record.app.js#L166)]. We can change the first one to look for facet as well as filter. And the other case can be removed for now.
- Recordedit:
  - We don't need any changes for this app.
- Recordset:
  - Add the multi-edit link in chaise.
  - ~Change ermrestjs parser to convert filter into a facet. This requires changes to [this function](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/parser.js#L73). In the process we should remove all the filter related APIs. Although this change will affect all the other apps too, but other apps can be functional without this change too.~ This has been moved to [ermrestJS issue #558](https://github.com/informatics-isi-edu/ermrestjs/issues/558)

PR #1385, opened 2017-11-21
Apps understand url with Faceting

jrchudy:
This PR addresses issue #1287.

Each app can be used with either a filter in the url or the facet blob in the url. Added back the Edit link on `recordset`.

### DIFF
--- a/record/record.app.js
+++ b/record/record.app.js
@@ -161,7 +161,7 @@
                 *  recordSetLink should be used to present user with  an option in case of no data found/more data found(>1)
                 *  This could be link to RECORDSET or SEARCH.
                 */
-                var recordSetLink = page.reference.unfiltered.contextualize.compact.appLink;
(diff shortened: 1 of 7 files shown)


record/record.app.js
```diff
@@ -163,10 +164,11 @@
                 var recordSetLink = page.reference.contextualize.compact.appLink;
 
                 if (page.tuples.length < 1) {
-                    throw new Errors.noRecordError(context.filter.filters, recordSetLink);
+                    throw new Errors.noRecordError({}, recordSetLink);
```
RFSH: I think it's better if we use the unfiltered url (`page.reference.unfiltered.contextualize.compact.appLink`) in this case for going to recordset. Currently it will go to a recordset page that has no data in it.

RFSH:
We can also add a test case for record app with faceting. You can use the [`ERMrest.createPath`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/parser.js#L38) function that given facet object will create a path for you.

PR #1387, opened 2017-11-28
Add showNull to Recordset Directive Config

RFSH:
To differentiate between empty string and null value. It is only used by scalar faceting for now.

### DIFF
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -761,8 +761,7 @@
                                     if (i !== -1) {
                                         return;
                                     }
-                                    
-                                    // currentValues[value] = true;
+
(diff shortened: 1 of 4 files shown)


RFSH:
@jrchudy I couldn't find any test cases for the modal. If we're going to add those test cases in the round 2, testing this feature would belong there.

jrchudy:
I see what you are talking about. I wrote an it in `ind-facet.spec.js` that opens the modal and verifies the selected row count and then selecting another one and returning form there. I did not verify all of the contents of the rows displayed in that modal (where I was thinking we could test this feature).

PR #1400, opened 2017-12-05
add $fkeys support to domain_filter_pattern

RFSH:
This include changes to add $fkeys support to domain_filter_pattern. 

The changes that I made for each scenario are:

1. For edit/copy mode: Attach `linkedData` (data from foreignkeys) to `vm.foreignKeyData`. So it can be used for `filteredRef`.

2. For create mode: Generate extra requests to get the actual value of foreign keys that have default value (Show spinner and also make the add multiple button disabled).

3. For pre-fill mode: Pass extra information so we can get the actual value of the foreignkey, instead of passing all the values of that table (Show spinner and also make the add multiple button disabled).

4. changing the foreignkey: Update the `foreignKeyData`. If foreign key data is loading (default or pre-fill) ignore that and just use the user-selected value.

5. clearing a foreign key: Update the `foreignKeyData`.

Test Cases:
I had to update some testcases based on the changes that I made. I also added extra test cases to the `domain-filter.spec.js` for create and edit mode.

_You need [ermrestjs#565](https://github.com/informatics-isi-edu/ermrestjs/pull/565) for testing this._

### DIFF
--- a/recordedit/recordEdit.app.js
+++ b/recordedit/recordEdit.app.js
@@ -38,7 +38,7 @@
     .config(['$uibTooltipProvider', function($uibTooltipProvider) {
         $uibTooltipProvider.options({appendToBody: true});
     }])
-    
+
     //  Enable log system, if in debug mode
     .config(['$logProvider', function($logProvider) {
         $logProvider.debugEnabled(chaiseConfig.debug === true);
@@ -49,7 +49,7 @@
 
         var session,
             context = { booleanValues: ['', true, false] };
-        
+
         $rootScope.showColumnSpinner = [{}];
 
         $rootScope.displayReady = false;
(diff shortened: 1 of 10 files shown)


recordedit/recordEdit.app.js
```diff
@@ -341,23 +377,42 @@
...
+                        throw new ERMrest.UnauthorizedError(messageMap.unauthorizedErrorCode, messageMap.unauthorizedMessage);
                     }
                 }
             }, function error(response) {
                 throw response;
             });
         });
 
+
+        function getForeignKeyData (rowIndex, colName, fkName, fkRef) {
```
jrchudy: Does this work defining the function here? I would imagine it needs to be defined above where it's used in the script. But maybe because it's in a callback, it's fine here.

RFSH: It does work. It's inside the run block, and I could move it inside the resolve callback. But I thought that it is going to make it more difficult to read.

Issue #1353, opened 2017-11-01
View More should not use appLink


jrchudy:
The view more button for the related tables on `record` app uses the appLink function. If the appLink annotation is defined for the `compact` context with `search` as the desired application, the view more link would bring the user to a broken `search` page.

View more should always navigate to `recordset`.

amitjha21:
This has been merged into `master`. Assigning to  @hongsudt  @jrchudy  to verify and close it.

PR #1411, opened 2017-12-14
View More always link to recordset

amitjha21:
This PR addresses #1353 
View More should redirect to `recordset` instead of `search` app.
Test case was tweaked in `all-features/record/related-table-link.spec.js`

### DIFF
--- a/test/e2e/specs/all-features/record/related-table-link.spec.js
+++ b/test/e2e/specs/all-features/record/related-table-link.spec.js
@@ -32,7 +32,8 @@ var testParams = {
     association_count: 1,
     associationDisabledRows: ["1"],
     page_size: 2,
-    price: "247.00"
+    price: "247.00",
+    schedule: "schedule"
 };
 
 describe('View existing record,', function() {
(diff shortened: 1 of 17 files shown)


test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -378,6 +378,8 @@ describe('View existing record,', function() {
                         return browser.driver.getCurrentUrl();
                     }).then(function(url) {
                         expect(url.indexOf('recordset')).toBeGreaterThan(-1);
+                        // View more should not redirect to Search app
```
jrchudy: This test is unnecessary. The line above tests exactly this case. If it's anything but `recordset` the above expectation fails. So if it's search, it will fail.

test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -378,6 +378,8 @@ describe('View existing record,', function() {
                         return browser.driver.getCurrentUrl();
                     }).then(function(url) {
                         expect(url.indexOf('recordset')).toBeGreaterThan(-1);
```
jrchudy: Add a comment to the expectation clause in the case it fails.

amitjha21:
Have changed the test case with annotation defined in `product-unordered-related-tables-links` schema.
Also since no of related table increased, `maxRelatedTablesOpen` in `copy-btn.spec.js` started failing (it has limit of 5 tab earlier), I have enhanced the test case by adding entries in tables. Increased the limit to 6 to test the open tabs.

amitjha21:
Updated the branch with the `master` to avoid any sync failure during testing.

RFSH:
@amitjha21 The test case that you have added was failing for me because of the race condition. I just added a check to make sure the button is clickable.

Issue #1354, opened 2017-11-01
New facet-mode for presence/not-null?


karlcz:
We already have an entity mode versus scalar mode distinction for facets.

Some use-cases would benefit from a "presence" test where the projected facet column is tested `!column::null::`  (or `!column::null::;column!=null` for json/jsonb columns).

1. In my use cases, the column holds asset URLs and/or other stuff that is meaningless to the user. So I'd really want a presence-mode that just offers a streamlined UX for a checkbox w/o displaying any of the present values.
2. In other use cases, I think DBAs may want to offer the presence test as an additional checkbox beyond whatever choices/ranges/search UX is also offered.

I think we should try to find a single annotation feature and implementation mechanism that can cover both scenarios by mixing it with the other features.  I suspect we'd want to allow the DBA to control the labeling of the presence test checkbox so they can introduce an appropriate semantic label that complements their choice of facet name.

Perhaps this could be as simple as a new optional field in a facet object (sibling to `source` and `markdown_name` etc.) which looks something like:  `"presence":` _label_  ?

@hongsudt @ljpearlman @robes

hongsudt:
We discussed this issue today and decided to introduce a not-null flag in the current facet syntax for dealing with not null checking. 

For UX simplicity, I am proposing to show a checkbox for `records with values` only in the compact/select (e.g. view more) mode. If the column or fk is not null, then this option should not be shown.

RFSH:
`not_null` has been added as a new attribute to the facets. I will update the annotation and also the facet wiki page.

PR #1414, opened 2017-12-19
Facet Not Null Support

RFSH:
This PR includes:

1. Adding "Records With Value" (not-null) option in the recordset selector for faceting. The logic for not-null filter is as follows:
   -  In modal: If you choose "Records With value" all the selected rows must be deselected and all the rows should be disabled. Users should be able to click on `clear all`, `x` besides the "Records With Value", and also the checkmark for "Records With Value" to remove the filter. After removing it, all the rows should become selectable.
   - In Facet panel: If you end up in the recordset with not-null filter, it should always show at the top of the list, and other options must be disabled. Deselecting it, should make all the other options enabled. If you select the not-null again, it should clear and disable all the other filters.

2. Extracting logic of recordset and table into the recordTable factory which then can be used in multiple places.

- ~~**Since I changed the structure of the recordset table, some of the test cases are failing. I created this PR so we can just review the change, and discuss if this approach of extending directive is acceptable or not.**~~**Test cases seem stable now.**

- **This PR needs e2e testing**.

*We need to merge the ermrestjs PR first.*

Related Issue: #1354

### DIFF
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -4,7 +4,7 @@
     angular.module('chaise.faceting', ['plotly'])
 
         .directive('faceting', ['recordTableUtils', '$timeout', '$rootScope', function (recordTableUtils, $timeout, $rootScope) {
-            
+
             return {
                 restrict: 'AE',
                 templateUrl: '../common/templates/faceting/faceting.html',
@@ -13,16 +13,16 @@
                 },
                 controller: ['$scope', function ($scope) {
                     var ctrl = this;
-                    
+
                     ctrl.childCtrls = []; // child controllers
-                    
+
                     ctrl.facetingCount = 0;
-                    
+
                     // register a children controller in here
                     ctrl.register = function (childCtrl, facetColumn, index) {
                         ctrl.childCtrls[index] = childCtrl;
                         ctrl.facetingCount++;
-                        
+
                         $scope.vm.facetModels[index] = {
                             initialized: false,
                             isOpen: false,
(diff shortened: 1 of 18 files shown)


common/faceting.js
```diff
@@ -698,86 +711,92 @@
...
-                            if (scope.reference.uri !== uri) {
-                                defer.resolve(false);
+                }
+
+
+                // read new data if neede
+                (function (uri) {
+                    scope.reference.read(appliedLen + tableConstants.PAGE_SIZE).then(function (page) {
+                        // if this is not the result of latest facet change
+                        if (scope.reference.uri !== uri) {
```
jrchudy: Shouldn't this be something like `page.reference.uri !== uri`?

The anon. function you defined passes in a uri variable which is `scope.reference.uri`.

common/table.js
```diff
@@ -426,24 +411,336 @@
...
+                }
+                if (tuples.length > 0) {
+                    callOnRowClick(scope, tuples, true);
+                }
+            };
+
+
+            // Facilitates the multi select functionality for multi edit by storing the tuple in the selectedRows array
+            scope.onSelect = function(args) {
+                console.log(args);
```
jrchudy: remove please. Or you can change it to $log with debug flag so it only shows up in that mode.

common/templates/ellipses.html
```diff
@@ -17,19 +17,19 @@
         </button>
     </div>
     <div ng-if="config.selectMode == multiSelect" class="chaise-checkbox">
-        <input type="checkbox" ng-checked="selected || selectDisabled" ng-click="onSelect()" ng-disabled="selectDisabled">
+        <input type="checkbox" ng-checked="selected" ng-click="onSelect()" ng-disabled="selectDisabled">
         <label />
     </div>
 </td>
-<td ng-repeat="val in rowValues track by $index">
+<td ng-if="config.showNull || tuple.uniqueId" ng-repeat="val in rowValues track by $index">
```
jrchudy: What tuple would not have a unique ID?

RFSH: The value of the key can be null for faceting (for both scalar and entity), and therefore the uniqueId would be empty string. But I needed to differentiate between empty string and null. I know it's not ideal for uniqueId to be null, but I couldn't think of a better way to handle it.

common/table.js
RFSH: I forgot to remove this. I added this to test that this way of extending the directive works or not.

common/faceting.js
RFSH: This is not something that I changed in this PR. But I think it's correct this way. If the facet changes then we are changing the `scope.reference`, therefore the `scope.reference.uri` is getting changed. But the `uri` is the link that we read the result with. So the `page.reference.uri` will always be equal to `uri`.

If you look at the other checks for flow-control in `table.js` and faceting.js, I'm always checking the captured `uri` is the same as `reference.uri` that is attached to `scope`/`vm` (and not the `page.reference.uri`).

jrchudy: I see what you are saying. `scope.reference` is updated to a different reference which is the reason we are capturing it's value beforehand.

RFSH:
@jrchudy I finished the test cases for this PR. They are passing for me locally. The PR test case for travis will fail here since it needs the emrestjs changes. We have to merge the ermrestjs PR first and then rerun the travis. But before that please take a look at the test cases and let me know if you think we need more tests or this is good enough.

Then we can merge the ermrestjs PR first and then let this build go through.

jrchudy:
Looks like these tests are good enough. You are capturing the functionality and then going through a couple other expected use cases for that functionality and how it should behave in the modal and with the facet panel.

Issue #1409, opened 2017-12-13
Add Pure And Binary Testcases for Inline


RFSH:
We have a test case for inline display of a related entity, but we don't have any test cases for inline pure and binary association related entities. We should add a test case for this.

Edited body:

### [Schema document](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/data_setup/schema/product.json) changes:
 - [x] a new table needs to be added to the product schema document to act as an association between `Accommodation` and `File`
   - [This](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/data_setup/schema/recordset/faceting.json#L224-L284) is a good example of how an association table is defined. A pure and binary association table must have 2 foreign keys, and a key that is created from those two foreign key columns. This concept is essentially used to describe a many to many relationship between two tables.
   - In the following I wrote the table definition which you can directly use (Please don't copy the table definition blindly and try to familiarize yourself with the concept of association table). The  product schema also already has an association table called `accomodation_image`, this table is going to be very similar to that (minus the extra attributes).
    - data needs to be added [here](https://github.com/informatics-isi-edu/chaise/tree/master/test/e2e/data_setup/data/product) for that new association table. 
    - Note: The json document filename has to have the same name as the table in the schema
- [x] modify the [visible columns list](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/data_setup/schema/record/product.json#L363-L399), so that the new table shows up as an inline related table
     - You can take a look at the annotation document for details about how to define a pseduo-column. But based on the given schema you need to add this to the list: 
       ```json
         {
            "source": [ 
              {"inbound": ["product-record", "fk_accom_image_assoc_1"]},
              {"outbound": ["product-record", "fk_accom_image_assoc_2"]},
              "id"
           ]
         }
       ```

Schema document:
```json
"accommodation_image_assoc": {
  "comment": "Association Table between accomodatiion and image",
  "kind": "table",
  "table_name": "accommodation_image_assoc",
  "schema_name": "product-record",
  "keys": [{
	  "unique_columns": [
		"image_id",
		"accommodation_id"
	  ]
  }],
  "entityCount": 0,
  "foreign_keys": [
	{
	  "names" : [["product-record", "fk_accom_image_assoc_1"]],
	  "foreign_key_columns": [
		{
		  "table_name": "accommodation_image_assoc",
		  "schema_name": "product-record",
		  "column_name": "accommodation_id"
		}
	  ],
[...]

PR #1418, opened 2018-01-03
Pure and binary#1409

amitjha21:
This PR addresses #1409 .

### DIFF
--- a/test/e2e/utils/record-helpers.js
+++ b/test/e2e/utils/record-helpers.js
@@ -239,32 +239,32 @@ exports.testPresentation = function (tableParams) {
         })
     });
 
-		it("click event on image_id in inline display should open new tab with file details",function(){
-				var allHandle;
-				chaisePage.waitForElement(element(by.id('entity-booking'))).then(function(){
-						return browser.executeScript("return $('#divRecordSet > table > tbody > tr:nth-child(1) > td:nth-child(5) > div > div > span > a')");
-				}).then(function(imageLinks){						
-						browser.executeScript("return window.open(arguments[0], '_blank')", imageLinks[0]);
-						return browser.getAllWindowHandles();
-				}).then(function (handles){
-						allHandle = handles;
-						browser.switchTo().window(allHandle[1]);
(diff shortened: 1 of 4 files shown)


test/e2e/utils/record-helpers.js
```diff
@@ -239,7 +239,34 @@ exports.testPresentation = function (tableParams) {
         })
     });
 
-    it("visible column related table with inline inbound fk should display 'None' in markdown disply mode if no data was found.",function(){
+    it("click event on image_id in inline display should open new tab with file details",function(){
+        var allHandle;
+        chaisePage.waitForElement(element(by.id('entity-booking'))).then(function(){
+           return browser.executeScript("return $('#divRecordSet > table > tbody > tr:nth-child(1) > td:nth-child(5) > div > div > span > a')");
```
jrchudy: Add a comment saying what element this selector is selecting. 

A better alternative would be to add specific ID's to elements or general classes so you don't have to crawl through the html structure. If we change the html structure, it will break this test case so it might be better to identify the element rather than crawling the HTML.

test/e2e/utils/record-helpers.js
```diff
@@ -239,7 +239,34 @@ exports.testPresentation = function (tableParams) {
...
+        }).then(function(imageLinks){
+            browser.executeScript("return window.open(arguments[0], '_blank')", imageLinks[0]);
```
jrchudy: I didn't think of using `window.open` to open a link in another tab. 👍

test/e2e/utils/record-helpers.js
amitjha21: I remember pointing this way back last year that we should not rely upon nested html for capturing elements. But was told that it's what is expected so that even if a slight change happens in the DOM it should fail and we get to know the change.
Nonetheless, I tweaked it to reduce to just 3 levels and added a comment.

jrchudy:
The changes look good but the build is failing now. Could you merge the errors PR first and merge it into this branch and let the build re-run?

amitjha21:
Synced up and E2E executed with the latest master. Will merge once the dev.Isrd VM is up.

Issue #1421, opened 2018-01-11
markdown table doesn't display cell borders


hongsudt:
- We should check the style so it supports the expected behavior
- Provide a way to over-ride the markdown table styling
- Create a wiki page on how to over-ride the styling. Based on my experience, chaise looks for styling under /assets/styles/chaise.css. You can check with @RFSH on the detail. 

Also, the markdown rendering of preview and the actual render in the app looks different.

RFSH:
If we want our markdown to look like github markdown by default, we can replicate the same css styles that github is applying. We don't need to figure out the css ourself, we can use [this](https://github.com/sindresorhus/github-markdown-css/blob/gh-pages/github-markdown.css) one. In this css files they are using `markdown-body` class which we should just change it to `markdown-container` and then use it.

For custom css for deployment, it's a chaise-config attribute (`customCSS`).

amitjha21:
Added `gihub` css in our repo.
Here is the demo link:
https://dev.isrd.isi.edu/~amit/md/record/#1/isa:dataset/id=269

amitjha21:
I have tested above link with markdown entries and also with custom CSS in assets/css/chaise.css.
It works as expected and provides a way to override styling by changing classes in custom CSS.

However, adding styles through specific classes that applies only during rendering in `record` app and not during preview in `recordedit` app (or vice versa) will cause expected differences in appearance (as can be seen in current dev rbk portal).

RFSH:
What are those specific styles that we are applying just for `record` app? Can't they be just based on `markdown-container` so it would work on any apps?

amitjha21:
We do not have anything app specific. It is all on `.markdown-container`.

RFSH:
You said "adding styles through specific classes that applies only during rendering in record app and not during preview in recordedit app (or vice versa) will cause expected differences in appearance (as can be seen in current dev rbk portal)", I'm just asking what are these styles that we are applying only to record app in rbk?

amitjha21:
Here are some  examples I found in rbk portal custom CSS:
``` css
#tblRecord td.entity-value table {
  
}

#tblRecord td.entity-value table>tbody>tr>td, #tblRecord td.entity-value table>tbody>tr>th, #tblRecord td.entity-value table>tfoot>tr>td, #tblRecord td.entity-value table>tfoot>tr>th, #tblRecord td.entity-value table>thead>tr>td, #tblRecord td.entity-value table>thead>tr>th {
[...]

PR #1424, opened 2018-01-18
Markdown table styling 

amitjha21:
This PR addresses markdown styling issue #1421.
Markdown preview and rendering is now based on GitHub markdown styles.


RFSH:
I think having these css rules in an extra file is a better approach. It's true that we're using the github markdown css, but there might some styles that we don't like and we want to change. So I think it's better if we add these github css in a separate file (the same way that bootstrap css is separated from ours). Then if we want to change that, we can overwrite those rules in the app.css .

If your concern is performance, we should eventually add a step to Makefile to concatenate and minify all the css files into one.

amitjha21:
Created a separate file for styling. `common/vendor/MarkdownEditor/styles/github-markdown.css` has all unminified styling css.

RFSH:
Thanks. Since now we're treating it as a third-party library you can minify it. Just add the link to its source (the repository that you borrowed the code from), and also add a comment that the container class is `markdown-container` (the source had a different class for it). Also I think you missed recordedit in your resources. Our makefile is a bit messy, I think I should create an issue to clean it up.

amitjha21:
The `recordedit` resource is there via MarkdownEditor entry in the `Makefile`.

PR #1429, opened 2018-01-25
Make Sure Disabled File Upload Works Properly

RFSH:
I'm not sure if the test case that I added was enough. Please let me know if I need to add extra test cases

### DIFF
--- a/recordedit/form.controller.js
+++ b/recordedit/form.controller.js
@@ -517,6 +517,8 @@
                     return '';
                 } else if (column.isForeignKey) {
                     return 'Select a value';
+                } else if (column.isAsset) {
+                    return "No file Selected";
                 }
             } catch (e) {
                 $log.info(e);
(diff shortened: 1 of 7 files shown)


recordedit/form.controller.js
```diff
@@ -517,6 +517,8 @@
                     return '';
                 } else if (column.isForeignKey) {
                     return 'Select a value';
+                } else if (column.isAsset) {
+                    return "No file Selected";
```
jrchudy: Should we move this logic to `column.getInputDisabled()` so that it returns `disabled` as an object with a message in it?

RFSH: This is the case that the column is enabled. We associated `false` to be the value for enabled (the column.inputDisabled returns `false` if the column is enabled). If I change this one to an object, that expectation would be different (I also have to do it for foreign key too). And also I don't think it adds that much value.

Issue #1382, opened 2017-11-21
prototype related records section that shows a list of related tables 


hongsudt:
* We should reuse the components that we have for recordset to create 2 independent panes on the record app---one for the table of content style that lists related tables, and one that shows the main content. 
* When I click on each related record, it should direct me to the relevant related record table in the main pane. It should automatically expand that table. 
* The TOC pane should be adjustable and collapsable. There should be a GUI control or a parameter to collapse TOC. 
* We might want to consider moving TOC to the right to visually distinguish the record app from recordset.

hongsudt:
@robes @carlkesselman @ljpearlman I think this prototype should help with many of the UX issue in the case of multiple related tables.

amitjha21:
I have added a demo link [here](https://dev.isrd.isi.edu/~amit/toc/record/#1/isa:dataset/id=269). Let me know about any changes.

robes:
Looks promising. Couple thoughts:
- is the TOC heading using a standard font/size? It looks a little different but I can't tell exactly.
- the expand/collapse icon introduces a whitespace column/gap. It would be good to tighten that up.
- do we want the nested tables to be collapsed now by default? That said (or asked), I do like how when you follow a TOC link it not only scrolls to the position (as expected) but also opens the nested table.

jrchudy:
Maybe there should be a "Top of Record" link?  (Top of Page?) Seems useful now that some users may navigate around the page quickly using the sidebar instead of scrolling.

amitjha21:
The heading is using inherited font, however, size is different (with `h2` tag) to give it a header feel.
Default behavior for accordion panels are based on a `chaise.config` entry named `maxRelatedTablesOpen`. If a page has more related table entries than this value then all panels will be closed by default. For the demo, the `maxRelatedTablesOpen` is set at 15.

amitjha21:
This has been merged into the `master`. Assigning to @hongsudt to verify and close it.

PR #1434, opened 2018-02-02
Table of Contents in Record

amitjha21:
This PR addresses #1382 .
It adds Table of Contents on the right side of the `record` app.
Chaise config entry `hideTableOfContents` is maintained to show/hide the TOC.
If the value is not present in the config then pan would be visible by default.
E.g.
https://dev.isrd.isi.edu/~amit/chaise/record/#1/isa:dataset/id=269


test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -94,10 +101,10 @@ describe('View existing record,', function() {
 
         });
 
-        describe("Presentation ,", function() {
-            var params = recordHelpers.testPresentation(testParams);
-        });
-
+    //     describe("Presentation ,", function() {
```
RFSH: uncomment this.

RFSH:
@amitjha21 I made changes to the logic and how the button should look like. This UX is good for this iteration. Can you please fix the test cases? The default behavior is having sidebar open and not collapsed.

RFSH:
Please avoid using css selectors directly in your test cases. Instead use the chaise.page.js to define a function and then call that. It makes it reusable and also easier to follow the code.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+        });
+        it('Table of contents should be collapsed by default', function(){
+            var recordSidePan = chaisePage.recordPage.getElementById('record-side-pan');
+            recordSidePan.isDisplayed().then(function (bool) {
+                expect(bool).toBeFalsy();
+            });
+        });
+
+        it('Table of contents should match all related table', function(){
+            var  allPan = element.all(by.xpath("//div[contains(@id,'recordSidePan')]"));
```
jrchudy: Please write this as a `chaisePage.recordPage` function. This can be done with a statement similar to:
`element.all(by.css('.some-class-you-added')` 
Or use a parent element that you can fetch and get all children of that parent:
`element(by.css('faceting-container)).all(by.tagName('li'))`

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+            }).then(function(){
+              expect(allPan.count()).toBe(testParams.sidePanelTest.tocCount, "Table of contents count did not match!");
+            }).catch( function(err) {
+               console.log(err);
+           });
+        });
+
+        it('On click of Related table name in TOC, page should move to the contents and open the table details', function(){
+            var rtTableHeading = chaisePage.recordPage.getRelatedTableHeading(testParams.sidePanelTest.tableToShow),
+                lastRow = element(by.id('rt-heading-Categories_5')).all(by.tagName('tr')).get(5);
```
jrchudy: This selector:
`chaisePage.recordPage.getRelatedTableHeading(testParams.sidePanelTest.tableToShow)`
and this one:
`element(by.id('rt-heading-Categories_5'))`

We have a chaisePage functions that fetches all the related table rows based on the displayName of the related table. `chaisePage.recordPage.getRelatedTableRows(displayname);`

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+
+            recSidePan.getAttribute('class').then(function(className) {
+                return recSidePan.click();
+            }).then (function (){
+              // related table should be visible
+                rtTableHeading.isDisplayed().then(function (bool) {
```
jrchudy: You have a promise inside a promise, then you chain another promises off the outer promise. Either this promise should be a callback:
```
  ...
  return rtTableHeading.isDisplayed();
}).then(function (bool) {
  expect(bool).toBeTruthy();

  return rtTableHeading.getAttribute("class");
...
```
OR
change it to a single expectation clause on one line:
```
expect(rtTableHeading.isDisplayed()).toBeTruthy()
```

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+                  expect(bool).toBeTruthy();
+                });
+                chaisePage.waitForElement(lastRow);
```
jrchudy: This `waitForElement` may not work properly if the inner data (the table rows) load after the panel opens. How this was done in other parts of our application was by waiting for a certain number of related table entries to be present before continuing.
```
browser.wait(function () {
  return chaisePage.recordPage.getREaltedTableRows(displayname).count().then(function (ct) {
    return (ct == numberOfRecords);
  });
});
```

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+                return rtTableHeading.getAttribute("class");
+            }).then (function(className) {
+                expect(className).toContain("panel-open", "Related table panel is not open when clicked through TOC.");
+            }).catch( function(err) {
+                console.log(err);
+            });
+        });
+
+        it('Record count should match for pan and related table content', function(){
```
jrchudy: All you need to test in this test case is if the side panel is displaying the proper count. We already know what the count should be (it should be defined as part of our expected data at the top of the spec). You do not need to count the rows and verify it's represented by the same value in the text of the side panel.

Simply grab the heading you are verifying and make sure the proper text is displayed.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+            var rtRow = browser.executeScript("return $('#rt-Categories_5 tbody tr')"),
```
jrchudy: this selector is unnecessary if you change the test case the way I suggested.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
jrchudy: Now that I think about it more, you should verify the full list of headings, you should be making sure each heading shows up properly and in the right order. [See here](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/utils/record-helpers.js#L180-L185).

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+                rtPanRecordCount = element(by.id('recordSidePan-heading-5')).all(by.tagName('span')).get(1),
```
jrchudy: This should be changed to fetch the full title of each heading in the side panel.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+            }).catch( function(err) {
+               console.log(err);
+           });
+        });
+
+        it('On click of Related table name in TOC, page should move to the contents and open the table details', function(){
+            var rtTableHeading = chaisePage.recordPage.getRelatedTableHeading(testParams.sidePanelTest.tableToShow),
+                lastRow = element(by.id('rt-heading-Categories_5')).all(by.tagName('tr')).get(5);
+
+            recSidePan.getAttribute('class').then(function(className) {
```
jrchudy: What's the point of this promise? Class name is not being tested for anything. You can ignore getting the `class` attribute and just call `click()`.

test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -378,7 +378,7 @@ describe('View existing record,', function() {
                     expect(relatedTableLink.isDisplayed()).toBeTruthy();
 
                     browser.wait(EC.elementToBeClickable(relatedTableLink));
-                    relatedTableLink.click().then(function() {
+                    browser.executeScript("return $('#rt-heading-schedule record-action-bar a.more-results-link').click();").then(function() {
```
jrchudy: If there is something wrong with the selector that we have defined for this, please fix the selector instead of changing the code. The selector is currently fetching the element with `#rt-heading-schedule .more-results-link` which should be working fine unless the more results link is no longer a child to the "related table heading".

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+            }).then(function(panRecordCount){
+               expect(parseInt(panRecordCount.slice(1, 2))).toBe(rtCount.length, "Record count did not match with side pan");
+             }).catch( function(err) {
+                console.log(err);
+            });
+        });
+
+        it('Side panel should hide/show by clicking pull button', function(){
+            var rtTableHeading = chaisePage.recordPage.getRelatedTableHeading(testParams.sidePanelTest.tableToShow),
+                recPan =  chaisePage.recordPage.getElementById('record-side-pan');
```
jrchudy: same comment as above, this should have it's own `chaisePage` selector.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+
+            recSidePan.getAttribute('class').then(function(className) {
```
jrchudy: What is the point of this function? You are not using `className` as part of it's callback.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+              return  fiddlerBtn.getAttribute("class");
+            }).then (function (className){
+                expect(className).toContain('glyphicon glyphicon-forward', 'Side Pan Pull button is not pointing in the right direction');
+                recPan.isDisplayed().then(function (bool) {
```
jrchudy: Same comment as above, you have a promise inside a promise chain that isn't chaining with the rest of the promises.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +170,92 @@ describe('View existing record,', function() {
...
+                    expect(bool).toBeTruthy();
+                });
+              return fiddlerBtn.click();
+            }).then (function() {
+              return  fiddlerBtn.getAttribute("class");
+            }).then(function(className){
+                expect(className).toContain("glyphicon glyphicon-backward", "Side Pan Pull button is not pointing in the right direction.");
+                recPan.isDisplayed().then(function (bool) {
```
jrchudy: same comment again.

test/e2e/utils/chaise.page.js
```diff
@@ -775,6 +778,10 @@ var recordPage = function() {
     this.getLoadingElement = function () {
         return element(by.id("rt-loading"));
     }
+
+    this.getElementById = function (id) {
```
jrchudy: This selector is unnecessary. It doesn't add any value to call `chaisePage.recordPage.getElementById(id)` instead of `element(by.id(id))` (which is much shorter). Our `chaisePage` functions are meant to be reusable functions that we use multiple times to fetch specific elements on the page.

jrchudy:
Above you stated:
> "Chaise config entry `showtableOfContents` is maintained to show/hide the TOC."

Is that the name of the property or is it `hideTableOfContents`?

You should have a test case that verifies the value of `chaiseConfig` that you expect to be defined. [See here](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/all-features-confirmation/record/presentation.spec.js#L83-L95).

Can you fix the indentation in `...schema/record/product.json`? It looks like more tables were added but git is showing that one table was changed a lot and the spacing is confusing.

@amitjha21 before assigning reviewers to a PR, can you look over your changes in `git` to make sure everyone is as you think it should be. There is one spot where you commented out a huge portion of the tests for one spec and didn't "uncomment" it after.

RFSH:
@amitjha21 @jrchudy About the chaise config, I changed it from `showtableOfContents` to `hideTableOfContents` since the default behavior should be showing the TOC, and chaise config can just hide it (I forgot to update the PR's description).

amitjha21:
There were multiple tables added to `product.json` at the same place, therefore for `git` show them all with the green highlight in a continuous sequence. I have added one extra line after each table definition to add some visibility. (_Although this extra line also appears in green highlight_)
> Can you fix the indentation in ...schema/record/product.json? It looks like more tables were added but git is showing that one table was changed a lot and the spacing is confusing.

For chaiseConfig, we want to check if the `hideTableOfContents` is not defined then it should show the panel by default. So not defining it in chaiseConfig, makes it a case by itself for default visibility testing. 
>You should have a test case that verifies the value of chaiseConfig that you expect to be defined. See here.

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +172,72 @@ describe('View existing record,', function() {
...
+
+        it('Side panel should hide/show by clicking pull button', function(){
+            var recPan =  chaisePage.recordPage.getSidePanel();
+
+            fiddlerBtn.getAttribute("class").then(function(classNameRight) {
+                expect(classNameRight).toContain('glyphicon glyphicon-triangle-right', 'Side Pan Pull button is not pointing in the right direction');
+                return recPan.getAttribute("class");
+            }).then(function(openPanel){
+                expect(openPanel).toContain('open-panel', 'Side Panel is not visible when fiddler is poining in right direction');
+                fiddlerBtn.click();
```
RFSH: You need to return the promise (click event) here and chain it to make sure that you have clicked on the button and then check for the attribute.

Also I think you can change all these getAttribute test cases in just one statement. `expect(recPan.getAttribute("class")).toContain(CLASS, ERR_MESSAGE)`

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +172,72 @@ describe('View existing record,', function() {
...
+                return fiddlerBtn.getAttribute("class");
+            }).then(function(classNameLeft){
+                expect(classNameLeft).toContain("glyphicon glyphicon-triangle-left", "Side Pan Pull button is not pointing in the left direction.");
+                return recPan.getAttribute("class");
+            }).then(function(closePanel){
+                expect(closePanel).toContain('close-panel', 'Side Panel is not hidden when fiddler is poining in left direction');
+            }).catch( function(err) {
+                console.log(err);
+                expect('Encountered an error').toBe('Please check the log', 'Inside catch block');
```
RFSH: I know that this is what we are doing in most of our test cases . But this is wrong and we should not continue writing it the same way. What you can do is add `done` function as a parameter for the it, and then calling `done.fail()` will fail the test case and `done()` means successful and the end of test case.

```javascript
it ("DESC", function (done) {
   promise1.then(function () {
      return promise2;
   }).then(function () {
      return promise3;
   }).then(function () {
       // ...
      done();
   }).catch (function (err) {
       console.log(err);
       done.fail();
   })
})
```

test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -163,4 +172,72 @@ describe('View existing record,', function() {
...
+            });
+        });
+
+        it('Record count along with heading should match for the panel and related table content should be in correct order', function(){
+            chaisePage.recordPage.getSidePanelTableTitles().then(function(tableNames){
+              for (var i=0; i<tableNames.length; i++){
+                tableNames[i] = tableNames[i].replace(/ +/g, " ");
+              }
+                expect(tableNames).toEqual(testParams.sidePanelTest.sidePanelTableOrder,"Order is not maintained for related tables in side panel");
+                expect(tableNames.length).toEqual(testParams.sidePanelTest.tocCount, "Count mismatch for number of related tables in side panel");
```
RFSH: This seems unnecessary since the previous one is comparing both arrays. You can first test the count and then whole array. This way if the length fails, it won't check the other `expect`. But that's up to you. Either remove this second expect or just move it up.

amitjha21:
Added done() for these test cases.

amitjha21:
If there are no further comments then can I go ahead and merge it?

RFSH:
Can you please just add a simple test case to another spec that has `hideTableOfContents`? to just make sure in that case it will be closed by default. You don't need to open the panel or anything, just make sure that its in the closed state. You can add it to `no-delete-btn.spec.js` spec for example. If you can find a better name (`no-delete-no-toc` maybe?) for this spec, change it. Otherwise you can leave it the way it is. We should eventually clean these spec and folder names.

Issue #1332, opened 2017-10-10
Recordset Should Handle Faceting Errors Differently


RFSH:
Recordset should treat errors that parser throws regarding invalid facets differently. We should show soft alerts. And upon catching the error it should read the unfiltered reference (to remove the facets and avoid the error).

hongsudt:
@RFSH can you identify errors that are not terminal?  Here is a list that I can think of: 

- [ ] invalid sorting criteria

RFSH:
Any part of error that can easily be ignored. These are what I can think of:

1. Invalid sort criteria

2. Invalid page criteria

3. Invalid facet

4. Invalid sort

This task includes changing ermrestjs to return appropriate errors too. For example for invalid sort criteria we don't have specific type of error.

There are some errors that we are throwing while we're parsing the url which still can be ignored, but I'm not sure how we can handle them. Since if error happens while parsing the url we won't have a reference to get the unfiltered version of it. For instance if the faceting blob cannot be turned into a JSON, we're throwing an error while parsing the url. But if one of the columns is invalid, we're throwing an error when we do the read. In both cases just removing the facet blob can solve the problem. We might have to change the way we're generating the error or let the chaise deal with those. Or we can ignore those cases.

amitjha21:
As per discussion, messages will appear in Modal with OK button:

**Type** | **Message** | **Link formation**
------- | ------- | -------
Invalid sort criteria | Generated at `Ermrestjs` | Strip `sort` and `after, before` 
Invalid page criteria | Generated at `Ermrestjs` | Strip `after, before` 
Invalid facet | Generated at `Ermrestjs` | Strip `facet` and other conditions can co-exist
Invalid filter | Generated at `Ermrestjs` | Strip `filter` and other conditions can co-exist

These operations can be done at `ermrestjs` level.
For `record`  and `recordedit` apps, in case of no filters, it should go to `recordset`.

amitjha21:
Few examples of error modal:
![image](https://user-images.githubusercontent.com/9977014/35466569-cd0b7930-02b9-11e8-8680-6802b4fe85f4.png)


![image](https://user-images.githubusercontent.com/9977014/35466588-fac9ec12-02b9-11e8-9645-766f7fef4f02.png)

amitjha21:
This is merged into master. Assigning to @hongsudt and @RFSH  to verify and close it.

PR #1436, opened 2018-02-05
Facet errors

amitjha21:
This PR addressed #1332 
It has made changes to `common/errros.js` to catch these newly created errors (`InvalidFacetSorting` and `InvalidPageCriteria`.) and generate error modal accordingly.

**Type** | **Message** | **Link formation**
------- | ------- | -------
Invalid sort criteria | Generated at `Ermrestjs` | Strip `sort` and `after, before` 
Invalid page criteria | Generated at `Ermrestjs` | Strip `after, before` 
Invalid facet | Generated at `Ermrestjs` | Strip `facet` and other conditions can co-exist
Invalid filter | Generated at `Ermrestjs` | Strip `filter` and other conditions can co-exist

Depending upon which error is being thrown the modal header changes and asks user to click on the OK button to reload the page without the faulty condition.
You can play around with the condition [here](https://dev.isrd.isi.edu/~amit/chaise1/record/#1/isa:dataset/id=269@after(id)) to see the behavior in error modal.

Minor modifications are done in `common/modal.js` and `utils.js` to accommodate modal body changes.
Test cases are written in [`errors.spec.js`](https://github.com/informatics-isi-edu/chaise/blob/facetErrors/test/e2e/specs/all-features-confirmation/errors/errors.spec.js)


amitjha21:
Test cases were added. Could not test at Travis since timestamp issue is failing.

RFSH:
@amitjha21 If you remember in the meeting we mentioned that we don't want this logic to be in catch all. You cannot assume this error is being thrown for the read request on the main entity. It should be on the catch function (or error callback) for the read request on the main entity. More specifically these reads: [recordset](https://github.com/informatics-isi-edu/chaise/blob/master/recordset/recordset.js#L318), [recordset-with-faceting](https://github.com/informatics-isi-edu/chaise/blob/master/common/table.js#L209), [recordset-re-read](https://github.com/informatics-isi-edu/chaise/blob/master/common/table.js#L161), [record](https://github.com/informatics-isi-edu/chaise/blob/master/record/record.app.js#L156), [recordedit](https://github.com/informatics-isi-edu/chaise/blob/master/recordedit/recordEdit.app.js#L232)

Also the timestamp issue is in a PR and test cases should pass. You can [see](https://travis-ci.org/informatics-isi-edu/chaise/builds/341126248#L3607) that the build in this branch is failing on new test cases that you have added.

.travis.yml
```diff
@@ -77,6 +77,7 @@ before_install:
   - cd ..
   - git clone https://github.com/informatics-isi-edu/ermrestjs.git
   - cd ermrestjs
+  - git checkout -qf facetingErrors
```
jrchudy: please remove this change.

common/table.js
```diff
@@ -218,7 +221,14 @@
                     return defer.resolve(true);
                 }).catch(function(err) {
                     if (vm.reference.uri !== uri) {
-                        return defer.resolve(false);
+                        defer.resolve(false);
+                    } else {
```
jrchudy: Your else case here ignores the code that is supposed to run outside of it. This was previously refactored to behave like the following:

```
if vm.reference.uri !== uri) {
  return defer.resolve(false)
}
// continue code execution if above condition is not met
vm.hasLoaded = true;
vm.initialized = true;
...
return defer.reject(err);
```

common/table.js
```diff
@@ -218,7 +221,14 @@
...
+                        vm.hasLoaded = true;
+                        vm.initialized = true;
+                        if(err.errorData && err.errorData.redirectPath && err.errorData.redirectPath != ''){
```
jrchudy: What you are doing in this `if` condition should replace `...` that I typed in the above code block.

Issue #1265, opened 2017-09-13
Histogram support in Range picker


jrchudy:
The range picker needs to show a histogram for searching for each of integer, float, date, and timestamp.

hongsudt:
For integer and date, we want the bin width to be a whole integer so when we label the bin, it is easy to understand... 

Let's assume that 
```
min: the current min
max: the current max
#bin: number of bins
Then 
bin width = round up ((max - min/#bin))
requested max = min + (bid width * #bin)
The new (min, requested max, #bin) will create a whole number bin size 
```

hongsudt:
For auto-zoom, we will have to make that parametric too.. How about 60% zoom: 
```
median = floor(#bin/2)
#assume that we will zoom in 60%
range = #bin * 0.3
new min = min of bin (median - range) 
new max = max of bin (median + range)
```

PR #593, opened 2016-09-13
Added support to specify chaise-config to be used for a spec

chiragsanghvi:
This **PR** addresses the issue #566 **Be able to test chaise with different chaise configs**.

To specify which config you want to use for your spec, set `chaiseConfigFilePath` in your **protractor.conf.js** file as follows.

```js

var pConfig = require('./../../../../utils/protractor.configuration.js');

var config = pConfig.getConfig({
  
    // Change this to your desired filed name
    configFileName: 'recordedit.dev.json',

    page: 'recordedit',
    setBaseUrl: function(browser, data) {
      browser.params.url = process.env.CHAISE_BASE_URL + "/recordedit" + "/#" + data.catalogId  + "/" + data.schema.name;;
      return browser.params.url;
    },

    // Specify chaiseConfigPath
    chaiseConfigFilePath: 'test/e2e/specs/recordedit/data-independent/edit/chaise-config.js'
});

exports.config = config;
```

You also need to run the [ssh-agent](http://mah.everybody.org/docs/ssh) in background as well as export a new environment variable that specifies your remote **dev.isrd** sshpath.

For example
```sh
# run ssh-agent in background
$ eval ssh-agent

# add your key to ssh-agent
$ ssh-add PATH/TO/KEY

# export REMOTE_CHAISE_DIR_PATH=USERNAME@HOST:public_html/chaise
$ export REMOTE_CHAISE_DIR_PATH=chirag@dev.isrd.isi.edu:public_html/chaise
```

PR #1437, opened 2018-02-05
Histogram facet

jrchudy:
This PR includes the functionality for the histogram feature in `chaise` outlined by issue #1265. 

Changes were made to the range picker directive to allow for the histogram to be conditionally shown in facets for columns with types: `integer`, `float`, `date`, and `timestamp`. 

A few functional decision were made:
 - range inputs will always be filled in with the min/max values represented by the histogram
   - this data is always filtered based on the rest of the facet filters
 - Histogram buttons
   - Info button on hover describes the histogram interactions
   - zoom button zooms into the middle 50% of the displayed data
   - unzoom button zooms out to the previous zoomed data which could have been from interacting with the plot or using the zoom button (think of it like an undo button)
   - reset button resets the display to the absolute min and max for that set based on the display facet filters

This PR corresponds with `ermrestJS` PR [#593](https://github.com/informatics-isi-edu/ermrestjs/pull/593). That PR needs to be merged first.


jrchudy:
This PR is updated with changes made to the corresponding ermrestJS PR and merging master into both. @RFSH it is ready for review.

common/faceting.js
```diff
@@ -588,8 +621,102 @@
                         }
                     };
 
+                    // range is defined by the x values of the bar graph because layout.xaxis.type is `linear` or 'category'
+                    function setHistogramRange() {
+                        if (isColumnOfType("timestamp")) {
+                            var minMoment = moment(scope.rangeOptions.absMin.date + scope.rangeOptions.absMin.time, 'YYYY-MM-DDHH:mm:ss');
```
RFSH: The same way that you defined the formats in a constant object, we can do the same in chaise.

common/faceting.js
```diff
@@ -608,31 +735,23 @@
...
-                                        scope.rangeOptions.absMin = response[0];
-                                        scope.rangeOptions.absMax = response[1];
+                                    // initiailize the min/max values
+                                    setRangeMinMax(response[0], response[1]);
+
+                                    // if - the max/min are null
+                                    //    - bar_plot in annotation is 'false'
+                                    //    - histogram not supported for column type
+                                    if (!scope.showHistogram()) {
+                                        return;
```
RFSH: Have you tried this? I think you have to return a promise. Instead of `return;` you can just `defer.resolve(true)` here which will resolve the outer promise and end the execution.

common/faceting.js
```diff
@@ -608,31 +735,23 @@
...
                                     }
 
-                                    defer.resolve(true);
+                                    scope.histogramDataStack = [];
+
+                                    // get initial histogram data
+                                    return histogramData(numBuckets, false);
                                 }
+                            }).then(function () {
+                                defer.resolve(true);
```
RFSH: histogramData can be resolved with true or false. You should change this to use the returned value.

common/faceting.js
```diff
@@ -641,16 +760,158 @@
                         return defer.promise;
                     };
 
+                    // Zoom the set into the middle 50% of the buckets
+                    scope.zoomInPlot = function () {
+                        try {
+                            // we don't want the user to keep zooming in infinitely to save memory space
+                            if (scope.histogramDataStack.length >= 20) { throw new Error("Maximum data stack size reached"); }
```
RFSH: What will happen if they exceed the maximum? Shouldn't we remove the `scope. histogramDataStack` in that case so they can continue using the zoom in? Or we want the zoom in to not be responsive?

common/faceting.js
```diff
@@ -641,16 +760,158 @@
...
+
+                            // If the range between min/max is less than the number of buckets, no reason to zoom more
+                            // the bars will only get bigger without any more precision being offered
+                            if (isColumnOfType("int")) {
+                                if (scope.rangeOptions.absMax - scope.rangeOptions.absMin <= numBuckets) {
+                                    throw new Error("Can't zoom anymore");
```
RFSH: Should we show these errors as alert to the users?

common/faceting.js
```diff
@@ -588,8 +621,102 @@
...
+                    }
+
+                    function histogramData(bucketCount, relayout) {
+                        var defer = $q.defer();
+
+                        (function (uri) {
+                            var requestMin = isColumnOfType("timestamp") ? moment(scope.rangeOptions.absMin.date + scope.rangeOptions.absMin.time, 'YYYY-MM-DDHH:mm:ss').format('YYYY-MM-DDTHH:mm:ss.SSSZ') : scope.rangeOptions.absMin,
+                                requestMax = isColumnOfType("timestamp") ? moment(scope.rangeOptions.absMax.date + scope.rangeOptions.absMax.time, 'YYYY-MM-DDHH:mm:ss').format('YYYY-MM-DDTHH:mm:ss.SSSZ') : scope.rangeOptions.absMax;
+
+                            scope.facetColumn.column.groupAggregate.histogram(bucketCount, requestMin, requestMax).read().then(function (response) {
```
RFSH: Please add `catch` for the promise.

RFSH:
Can you please make sure that this is working properly? I pulled your ermrestjs and chaise branches and it's throwing error as soon as I open a range facet. It might be because of your recent changes in ermrestjs.

Also We need to test the scenario that `bar_plot: false` is in the annotation.

common/faceting.js
jrchudy: zoom will be unresponsive, a soft alert at top of page maybe?

jrchudy: I added a `disableZoomIn()` function that disables the button that when clicked calls this function. Because I guard around the same condition, this case _should_ never be reached.

common/faceting.js
jrchudy: Same comment as above, this case _should_ never be reached.

common/faceting.js
jrchudy: Should I remove these error alerts instead?

PR #1439, opened 2018-02-07
Fix Facet Based on Secondary Key Bug

RFSH:
The assumption in the previous implementation was that the `tuple.uniqueId` and `filter.uniqueId` are the same, but that might not be the case. If the facet definition is not based on the shortestkey these two will be different values. I changed the choicePicker implementation to  store the whole tuple object, so we can extract the `tuple.uniqueId` which is needed for modal picker (it's used in the logic for `isSelected`).
I also added a test case to the faceting spec for this one, and fixed the condition of one of the test cases that was failing locally for me.

### DIFF
--- a/test/e2e/specs/all-features/record/related-table-link.spec.js
+++ b/test/e2e/specs/all-features/record/related-table-link.spec.js
@@ -407,17 +407,12 @@ describe('View existing record,', function() {
                     expect(addRelatedRecordLink.getText()).toBe("Add", "The Add button is not displayed as Add");
 
                     addRelatedRecordLink.click().then(function(){
-                    }).then(function(){
-                        browser.wait(function () {
-                               return chaisePage.recordsetPage.getRows().count().then(function (ct) {
-                                   return (ct > 0);
-                               });
-                           });
-                        rows = chaisePage.recordsetPage.getRows();
-                    }).then(function(ct){
+                        chaisePage.waitForElementInverse(element.all(by.id("spinner")).first());
+
                         return browser.executeScript("return $('.modal-body tr input[type=checkbox]').get(2);");
                     }).then(function (selectButtons){
-                        selectButtons.click();
+                        return selectButtons.click();
+                    }).then(function () {
                         return browser.executeScript("return $('#multi-select-submit-btn').click();");
                     }).then(function () {
                         return browser.wait(EC.presenceOf(element(by.id('page-title'))), browser.params.defaultTimeout);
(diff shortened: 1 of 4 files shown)


test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -407,17 +407,12 @@ describe('View existing record,', function() {
...
                     addRelatedRecordLink.click().then(function(){
-                    }).then(function(){
-                        browser.wait(function () {
-                               return chaisePage.recordsetPage.getRows().count().then(function (ct) {
-                                   return (ct > 0);
-                               });
-                           });
-                        rows = chaisePage.recordsetPage.getRows();
-                    }).then(function(ct){
+                        chaisePage.waitForElementInverse(element.all(by.id("spinner")).first());
```
jrchudy: I had the same failure in my branch for histograms. Instead you should remove this line and mae the other changes I suggested.

test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -407,17 +407,12 @@ describe('View existing record,', function() {
                     expect(addRelatedRecordLink.getText()).toBe("Add", "The Add button is not displayed as Add");
 
                     addRelatedRecordLink.click().then(function(){
```
jrchudy: Add `chaisePage.waitForElement(chaisePage.recordEditPage.getModalTitle());` after this line.

test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -407,17 +407,12 @@ describe('View existing record,', function() {
...
-                    }).then(function(){
```
jrchudy: this is unnecessary. So it should stay removed

test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -407,17 +407,12 @@ describe('View existing record,', function() {
...
-                        browser.wait(function () {
-                               return chaisePage.recordsetPage.getRows().count().then(function (ct) {
```
jrchudy: change this to `chaisePage.recordsetPage.getModalRows()`.

test/e2e/specs/all-features/record/related-table-link.spec.js
```diff
@@ -407,17 +407,12 @@ describe('View existing record,', function() {
...
-                                   return (ct > 0);
```
jrchudy: change this to `ct == 4`.

Issue #1431, opened 2018-01-27
Inline display should not bulletize a list of one


robes:
When showing nested entities in the columns list (via the 'visible columns') annotation, when there is only ONE entity, the display should not use the html itemized list.

PR #1441, opened 2018-02-10
Single Bullet issue#1431

amitjha21:
This PR addresses #1431 .
If there is only one bullet then it should be removed and shown as normal content. 

Some styling changes also added to avoid indentation and spacing for multiple bullet items. These indentation and spacing are only for bullets appearing in inline section.

Demo: https://dev.isrd.isi.edu/~amit/chaise1/record/#41152/product-record:accommodation/id=2002

Test cases are added in [`record-helpers.js`](https://github.com/informatics-isi-edu/chaise/compare/bulletIssue%231431?expand=1#diff-0be163363798bf5bbe5b94a286747995) with new inline table entry[`booking_new_inline`](https://github.com/informatics-isi-edu/chaise/compare/bulletIssue%231431?expand=1#diff-3f5df00f4ac8f6975843cda06e685dc3)


amitjha21:
Synced with `master` and tested.

RFSH:
I think I said this before, please assign the PRs to yourself.

I'm against using javascript for something that we can do with CSS. You could do the same by just using [`only-child`](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-child) pseudo-class.

You can replace that javascript function with only the following:
```css
record-display .markdown-container ul > li:only-child,
record-display .markdown-container ol > li:only-child {
    list-style: none;
    padding: 0;
    margin-left: 0;
}
```
And if you want to get rid of the padding, you need to make more changes. There's no parent selector so one thing you can do is removing the padding from the ul and move it to li.

```css
.markdown-container ol, .markdown-container ul {
    padding: 0; /*existing rule is padding=2, which should be changed for the other changes*/
}

.markdown-container ol > li, .markdown-container ul > li {
    margin-left: 1em;
    padding-left: 0;
}
```

If you found a better solution use that. But try to avoid javascript as much as you can for these sort of changes.

amitjha21:
Added CSS entries for the bullets.

PR #1442, opened 2018-02-13
Make sure recordset works with filters

RFSH:
We might end up in recordset page with filters. These set of changes add filter support to the recordset app by showing the filter and applying it to the facets.

If we end up in a page with filters (if ermrsetjs could translate them to facet, it would and then chaise will only see facets. This is the scenario that ermerstjs couldn't change it to facet), 
  - We should show the filters with `Custom Filter` label.
  - Filters should be effective and recordset should show the right set of data for facets and main panel.
  - Users should be able to just clear the custom filters.
  - "Clear All' should remove custom filters as well.

This requires [ermrestjs changes](https://github.com/informatics-isi-edu/ermrestjs/pull/598). We should run the travis test cases after we merged that PR. For now, I canceled the build.


test/e2e/specs/delete-prohibited/recordset/ind-facet.spec.js
```diff
@@ -1375,6 +1386,94 @@ describe("Viewing Recordset with Faceting,", function() {
...
+                browser.ignoreSynchronization=true;
+                browser.get(uri);
+                chaisePage.waitForElementInverse(element(by.id("spinner")));
+            });
+
+            it ("should show the applied filter and clear all button.", function (done) {
+                chaisePage.recordsetPage.getFilters().then(function (filters) {
+                    expect(filters.length).toEqual(1, "filter is missing");
+
+                    expect(filters[0].getText()).toEqual("Custom Filter: " + customFilterParams.ermrestFilter);
```
jrchudy: Missing error text.

test/e2e/specs/delete-prohibited/recordset/ind-facet.spec.js
```diff
@@ -1375,6 +1386,94 @@ describe("Viewing Recordset with Faceting,", function() {
...
+                    console.log(err);
+                    done.fail();
+                });
+            });
+
+            it ("main and faceting data should be based on the filter, and be able to apply new filters.", function (done) {
+                // main
+                expect(chaisePage.recordsetPage.getRows().count()).toEqual(customFilterParams.numRows, "total row count missmatch.");
+
+                chaisePage.recordsetPage.getFacetById(idx).click().then(function (ct) {
```
jrchudy: `ct` isn't used in this case

Issue #1446, opened 2018-02-27
Refactor promise chains to be more deterministic


jrchudy:
We noticed some inconsistencies in the way errors were being handled in `ermrstJS` and `chaise`. Some cases we are blindly throwing an error in the middle of a promise chain when instead we should be rejecting the current promise and letting our `.catch()` function handle it or pass it off to the exception handler.

jrchudy:
NOTE: An important fact is that reject() DOES NOT terminate control flow like a return statement does. In contrast throw does terminate control flow.

PR #1447, opened 2018-02-28
refactored promise structures in chaise

jrchudy:
This PR addresses issue #1446.

### DIFF
--- a/common/ellipses.js
+++ b/common/ellipses.js
@@ -53,6 +53,8 @@
 
                 }, function deleteFailure(response) {
                     scope.$root.showSpinner = false;
+                    // NOTE: can either use throw or use $q.reject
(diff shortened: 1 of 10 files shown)


common/ellipses.js
```diff
@@ -53,6 +53,8 @@
 
                 }, function deleteFailure(response) {
```
RFSH: I think it's more preferable to have just catch and not both reject callback and catch here.

common/faceting.js
```diff
@@ -584,6 +588,8 @@
                             scope.parentCtrl.updateFacetColumn(scope.index);
                         } catch (err) {
                             $log.warn(err);
+                            // TODO: I think we want to throw the error here especially because a 500 error could occur
```
RFSH: It seems like this try-catch is not needed anymore.

Issue #1449, opened 2018-03-02
Table of Contents Changes


amitjha21:
Table of contents header should be renamed and Main link should appear on the top of the list.

amitjha21:
Changes merged into `master`. Assigning to @hongsudt @jrchudy to verify and close it.

PR #1450, opened 2018-03-02
TOC header and Main link on the top

amitjha21:
This PR addresses #1449 
Added **Main** link on the top of the list and renamed the header to **Contents**.
Test case is in [`presentation.spec.js`](https://github.com/informatics-isi-edu/chaise/compare/tocHeader?expand=1#diff-0c4175e21d5fe1dffc2ed1275f8a5798)

### DIFF
--- a/test/e2e/specs/all-features-confirmation/record/presentation.spec.js
+++ b/test/e2e/specs/all-features-confirmation/record/presentation.spec.js
@@ -62,10 +62,10 @@ var testParams = {
       schemaName: "product-record",
       tableName: "accommodation_collection",
       id: "2003",
-      tocCount: 7,
+      tocCount: 8,
       tableToShow: 'Categories_5',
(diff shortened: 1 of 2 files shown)


test/e2e/specs/all-features-confirmation/record/presentation.spec.js
```diff
@@ -201,6 +201,19 @@ describe('View existing record,', function() {
                 done.fail();
             });
         });
+        //Main Button
+        it('after clicking Main button, page should move up', function(done){
+            var mainButton = element(by.id('main-to-top')),
+                topOfPage = element(by.id('row-id'));
+
+            mainButton.click().then(function(){
+              expect(topOfPage.getLocation()).toEqual(jasmine.objectContaining({x: 30}, "Main button did not scroll up the page"));
```
RFSH: I know it's valuable to test this feature. But this way of testing seems wrong to me. It essentially is testing the margin of that element from top of the page which is css-based and can easily be changed. We shouldn't test the exact position of elements as it can easily be broken. 
And for testing this feature, it seems like there's no other way, so just remove the test cases. It should be one of the manual test steps.

hongsudt:
Please merge and update people on this improvement.

Issue #1406, opened 2017-12-09
log user in after session expires


hongsudt:
In RBK, after my session expires, the app treats me as an anonymous user instead of directing me to re-login. 

This behavior might confuses users as they will see a small set of records, and it isn't obvious that it is due to session has expired and they are now seeing records as an anonymous user.

jrchudy:
2 of the scenarios that we want to handle are as follows:
 - When the user is logged in and their session times out while they are away
 - When the user has the app open in a tab and the session expires. They later open a new tab and are an anonymous user

In both cases, we want to alert the user that a previous session had expired (either their session or any other users) and give them a dialog asking if they want to proceed as an anonymous user or to login. 

In the case of opening a new tab, we will have to look at some type of local storage and have that store information about the previous user session. If there is some previous user session data stored, give the user the popup notification detailed above.

Another case can be when the user leaves the app open in browser but then puts their computer to sleep causing the app to lose network connection and stop functioning. We should look into having a function fire off with a reconnect event to check the state of the session and show the above detailed notification based on the results.

jrchudy:
Besides cookies, there are two other storage options in the browser:
 - `sessionStorage` maintains a separate storage area for each given origin (separate between tabs) available for the duration of the session (until that tab or browser is closed). As long as the browser (or tab) remains open, `sessionStorage` persists through page reloads and restores

 - `localStorage` maintains a separate storage area for each given origin, but persists even when the browser is closed and reopened (can only be cleared by clearing cache)

jrchudy:
@hongsudt, The modal popup dialog will have the following text displayed:
 - title: "A previous login was detected"
 - message: 
```
"Click _here_ to login" 
"or to continue anonymously, click _here_"
```

The '_' s represent what will be a clickable link.

jrchudy:
Specifications for this issue:
 - When the user first logs in on that machine and no local storage object is found, create a local storage object and set an expires timestamp on it.
 - Logout should clear this timestamp value and act like it was never set to begin with
 - Each time the user comes to the app, check the timestamp value and if it's been passed our default time (30 mins, 60mins, or 120 mins?), prompt the user that a previous login was found
   - this requires having a listener on requests that involve getting more data to then update this `expires` value so the user isn't prompted during normal use of the application

jrchudy:
When the user logs in, remove this previous prompt expiration token. The idea is to limit when the prompt is shown to an anonymous user.

[...]

PR #1453, opened 2018-03-05
Expired session

jrchudy:
This PR addresses issue #1406. 

The user will be prompted with a login dialog as soon as they come to the application. If they decide to not login by either: 
  - clicking "anonymous access" 
  - clicking the "X"
  - press escape 
  - click the backdrop

A expiration token is created in local storage. This token has an expiration set to one hour from when it is created. As the user continues to use the application anonymously, this token will be extended with each successful request made. Once the user logs in, this token is destroyed.

If the user logs in instead of continuing anonymously, this token is never created.

This prompt only shows up for `record` and `recordset` where users can be anonymous. `recordedit` doesn't prompt the user for this because it requires the user to be logged in to make changes to the data. This PR is related to `ermrestJS` [PR #605](https://github.com/informatics-isi-edu/ermrestjs/pull/605).


common/authen.js
```diff
@@ -200,6 +249,38 @@
                 return _session;
             },
 
+            promptUserPreviousSession: function() {
+                // if there's no stored data OR
+                // there's stored data but we are after it's expiration
+                if (!promptTokenExists() || promptTokenHasExpired()) {
```
RFSH: This means that we're going to show this prompt to the first time users (users that have never logged in before) too. We want to show this only if you have logged in before and your session is expired.

Issue #1288, opened 2017-09-14
ability to show/hide (collapse) facet panel


jrchudy:
There should be a UX control to show/hide (collapse) the facet panel.

PR #1467, opened 2018-03-20
Facet panel collapse

jrchudy:
This PR resolves issue #1288.

The facet panel can be opened and closed when in the `recordset` with faceting. This uses the same functionality as the Table of Contents in `record` app. I refactored the functionality in record app so that it would be consistent in both places.


jrchudy:
Closing this PR because the changes will be included in PR #1468 .

Issue #1262, opened 2017-09-13
Generalizing facets as pseudo columns/entites


karlcz:
Once the new faceting MVP is done, I think we should broaden the scope of the new pseudo column. We should refactor the code as needed and make sure it can work in other _visible columns_  or _visible fk_ contexts to present denormalized content.  This would eliminate the need for custom SQL views in many simple denormalized presentation scenarios.

- In the recordset results table
- In the record related entities results tables
- In the modal chooser results table
- In the record main entity section

In all of the above contexts, there would be a few presentation modes:
- A facet object whose "source" field has only *outbound* foreign keys is a generalization of the current foreign key name as a visible column.
    - In entity-mode, present row_name of the single referenced record
    - In scalar mode, present the display value of the single referenced scalar
- A facet object whose "source" field has one or more *inbound* foreign keys is a generalization of today's "inline" inbound pseudo column
   - TBD: allow DBA-controlled aggregation of resulting facet?  E.g. min/max/count/array

I think that interpretation of facet pseudo columns for data presentation should ignore any embedded filters in the facet definition. That way, the same visible columns, e.g. for `*` context, might serve both for basic data presentation and also provide default filtering for the facet sidebar in the recordset app. 

Another case that was brought up by @robes which is only relevant in the record app context.. 
- A facet object whose "source" field has one or more *inbound* foreign keys is a generalization of the related table in the visible foreign-key list. 
------------------
HT: Here is a list of priorities that we should tackle (12/20/17). 
- [x] allow facet outbound-chain source as part of visible-column. The chain should be all outbound. This column should show up in the record and recordset apps. In recordset app, assume that this column is not sortable or searchable.
- [x] allow facet inbound source (or combination of inbound/outbound) as part of the visible-columns (compact/inline) or visible-fks (compact/brief). This feature is for record app only.
- [x] allow pseudo-columns of aggregate facet inbound sources. There is some correctness, performance, and ermrest readiness issues to generally support this feature. So we will only start with a subset. After discussing with @karlcz, we should be able to do the following aggregation without correctness issue: min, max, count distinct, count (faster performance, but only work for simple fk chains)
```javascript 
{ 
  "source": [...],
  "markdown_name": "...", 
  "entity": bool,
  "aggregate": "..."    /* either array/min/max/count/count distinct */
}
```
- [x] make outbound fk pseudo-column sortable? 
- [x] allow pseudo-columns of array aggregate of facet inbound sources. This depends on new ermrest feature.

[...]

PR #1470, opened 2018-03-22
Pseudo Column Support in record

RFSH:
This PR adds support for having pseudo-columns in record page for inline and related tables.

This includes:
- Refactoring related-table spec, so that we are actually testing all the common features. For that I had to create three different helper functions in record-helpers.
- Add test case for inline pseudo-columns.
- Refactor the structure of `record.html`. It was a little bit buggy, it was expecting a specific value instead of actually checking for related tables.
- Fix some of the failing tests in record spec.
- Modify test cases based on changes in ermrsetjs. In ermrestjs, I changed the logic for generating pseudo-column names. Some of the test cases were relying on certain names for foreignkeys which I had to change.

P.S. I have changed the travis.yml to make sure we're testing on the correct branch. Before merging this PR, we should merge ermrestjs PR and change the travis.yml back to its original.

ermrestjs PR: [ermrsetjs#610](https://github.com/informatics-isi-edu/ermrestjs/pull/610)
Related issue: #1262


RFSH:
@jrchudy The PR is ready to be reviewed.

test/e2e/specs/all-features/navbar/base-config.spec.js
```diff
@@ -49,36 +49,40 @@ describe('Navbar ', function() {
...
+    it('should open the profile card on click of My Profile link', function(done) {
+        var link = element(by.css('.dropdown-toggle'));
         browser.wait(EC.elementToBeClickable(link), browser.params.defaultTimeout).then(function() {
             link.click();
             var profileLink = element(by.id('profile-link'));
-            browser.wait(EC.elementToBeClickable(profileLink), browser.params.defaultTimeout).then(function() {
-                return profileLink.click();
-            }).then(function () {
+            browser.wait(EC.elementToBeClickable(profileLink), browser.params.defaultTimeout);
+            chaisePage.clickButton(profileLink).then(function() {
```
jrchudy: Was it the case that the button (at a specific location) couldn't be clicked which was causing this test to sometimes fail locally?

RFSH: Yes, this was one of the failing test cases. Now that you commented, I realized I should chain the previous click too (I changed and pushed the code).

Issue #1423, opened 2018-01-12
Secondary Action Errors should be dismissible


robes:
For example, you are on a record page, you try to delete (unlink) an associated record (which could be as simple as a vocabulary 'tag' like resource). Upon failure, which can happen with dynamic ACLs that allowed the 'unlink' icon to be present, you can get a Forbidden error. The dialog will leave no option but to return to the home page and lose the rest of your state.

![image](https://user-images.githubusercontent.com/4202842/34856550-83fa1fa0-f6fa-11e7-95f9-5d6de84e771a.png)

amitjha21:
This is meant only for `403 Forbidden` case; no other scenario where we might get `409 Conflict` while deleting items?

RFSH:
I think it should always be dismissible, regardless of the error.

robes:
Right.

RFSH:
- I don't think we're throwing modal error pop-ups for submitting the record. They're only alerts that we show at the top of the page.
- Edit and Download buttons are just opening links. For download, it's not even chaise environment so we cannot control that. And for Edit, it will go to recordedit and then that app will throw the error.

Based on this, I think only delete and unlink are the paths that we might end up with a pop-up error that should be dismissible.

And for the implementation, what you most probably should do is adding extra logic in the `catch` function of these secondary actions. For delete/unlink of table view, you can find the functions in `ellipses.js`.
 In the `catch`, you can create a new type of error that wraps the error that is being thrown and only has extra attribute/flags that signal the "catch-all" (`handleException`) to show a dismissible error. I remember there was some logic in that function to make sure we're always showing just one error pop-up. So dismissing a error, will prevent other errors to show. Make sure to handle that case.
The only reason that I think you still need to call the handleException function is that you still need to show the appropriate error pop-up and we don't want to duplicate the logic. There might be better way of implementing this though.

jrchudy:
As @RFSH stated, the only "secondary" actions that need to have this dismissible error popup are the `delete` and `unlink` actions. 

These actions are present in `record` app as the `delete` button in the top bookmark bar. And as the `delete`/`unlink` buttons in the related tables.
The `delete` action is also present in the `recordset` app as the `delete` button in the table.

amitjha21:
This has been merged into master. Assigning to @robes to verify and close it.

PR #1473, opened 2018-03-23
Dismissible error modal for secondary actions

amitjha21:
This addresses #1423 .

Whenever there is an error code of  `403` or `409`, the error modal should have a close button.
Test case : [`errors.spec.js`](https://github.com/informatics-isi-edu/chaise/compare/dismissErrorModal?expand=1#diff-98cb82d265253c464440d33edaa85663)


common/errors.js
```diff
@@ -180,7 +180,7 @@
 
             var params = {
                 message: message,
-                errorCode: errorCode,
+                errorStatus: errorStatus,
```
amitjha21: `errorCode` has been renamed to `errorStatus` as `errorCode` (the actual error http like code) is introduced; also erstwhile variable name (`errorCode`)  was confusing since it was holding a string instead of any code.

RFSH:
I think the requirement was specific to the delete or any secondary request. Think about it this way, we want error modals that shows up because of a secondary action and that has nothing to do with other parts of the app to be dismissible. In this case, user opens the page without any problems. Now they click on delete and get the error modal. Since this error modal was for the secondary action of deleting and has nothing to do with the loading of the page it should be dismissible.

Issue #1469, opened 2018-03-22
Design issues with the session expiration popup


ljpearlman:
I find the current implementation of the session expiration popup a little annoying:

- It pops up too frequently (I believe it pops up after an hour of inactivity)
- It should be easier for the average person to dismiss (I'm assuming the average person thinks they need to click on something in the popup to dismiss it, so they'll either have to go through the login process or find the less-obvious option to continue without logging in).
- I think the wording is confusing (if a user logs in once, should they still be seeing messages about that session expiring every day until the end of time?).

Some of these things could be tweaked, but I'd prefer to replace the popup with a banner on/near the menubar. This would both a) avoid the issue of people forgetting their status, because it would be present any time it was applicable, and b) not interfere with people's work.

PR #1483, opened 2018-04-06
Anonymous login alert

jrchudy:
The anonymous login prompt is no longer a modal popup but a soft alert warning that follows a similar set of use cases. This is for issue #1469.

The alert only shows up if a user has logged in before and had their session expire. 
  - If the user logs out, it deletes the stored value that we check for when deciding whether to show the alert or not. 
  - If the user does nothing with the alert, it will continue to show on each subsequent `record` and `recordset` page if the user doesn't login. 
  - If the user clicks the `x` to dismiss the alert, it creates the expiration token and sets it to an hour from now
  - If the user continues to make successful http requests, this expiration is extended to be an hour from now

### DIFF
--- a/common/alerts.js
+++ b/common/alerts.js
@@ -7,25 +7,27 @@
         var alerts = [];
         var ALERT_TYPES = ['success', 'error', 'warning'];
 
-        function Alert(message, type) {
+        function Alert(message, type, cb) {
             DataUtils.verify(message, 'Message required to create an alert.');
             if (type === undefined || ALERT_TYPES.indexOf(type) === -1) {
                 type = 'error';
             }
             this.message = message;
             this.type = type;
+            this.callback = cb;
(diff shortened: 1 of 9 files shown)


common/alerts.js
```diff
@@ -48,6 +50,10 @@
                 scope.closeAlert = function(alert) {
                     AlertsService.deleteAlert(alert);
                 };
+
+                scope.login = function() {
```
jrchudy: not sure if this should be defined here or if this should be a function that is passed along and attached to `scope`.

RFSH: Since we can now pass html with any callbacks in it to the alerts, maybe we should also add another parameter to `addAlert` which is an array of callbacks that we attach to scope and therefore can be called? I'm not sure what is the better solution here.

Issue #1487, opened 2018-04-10
Panel collapse button uses too much space


jrchudy:
The facet panel collapse button should not create a whole new row of whitespace in the `recordset` directive.

PR #1490, opened 2018-04-12
Facet collapse btn redesign

jrchudy:
This PR addresses issue #1487.

### DIFF
--- a/common/styles/app.css
+++ b/common/styles/app.css
@@ -1135,6 +1135,10 @@ footer .container1 {
      flex-basis: 0px !important;
  }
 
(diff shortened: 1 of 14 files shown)


common/styles/app.css
```diff
@@ -956,33 +956,28 @@ footer .container1 {
   -ms-user-select: none;
   -webkit-user-select: none;
   user-select: none;
-  background: transparent;
+  background: #fff;
+  border: solid 1px;
+  border-top: none;
```
RFSH: add `border-bottom: none` too. I think because of the calculation that we're doing the main-container might be 1-2px off the actual size and the border-bottom is visible.

Issue #1481, opened 2018-04-06
Login link in Terminal Error


jrchudy:
When a terminal error is thrown (I found this because of a syntax error) and the user is not logged in, a login link is present in the terminal error popup. This link should only be present in the case of a "Not Found" error.

PR #1497, opened 2018-04-24
fixed login link in error popup issue

jrchudy:
This PR fixes issue #1481. Link only shows in case that when schema/table/record returns a 404. Refactored handle exception function to be more readable.


common/modal.js
```diff
@@ -45,24 +45,24 @@
         vm.showReloadBtn = false;
         var reloadMessage = ' <p>  </p>';
 
-        if(vm.params.errorStatus == 'Multiple Records Found'){
+        if (vm.params.errorStatus == 'Multiple Records Found') {
```
RFSH: We can pass the actual error object to `errorPopup`, which then can be passed to the this controller. So we can use `instanceof` instead of checking for `errorStatus`.

common/errors.js
```diff
@@ -146,45 +146,41 @@
...
 
-        var reloadCb = function(){
+        var reloadCb = function() {
             window.location.reload();
         };
 
-        function errorPopup(message, errorStatus, pageName, redirectLink, subMessage, stackTrace, errorCode, isDismissible) {
-            var providedLink = true,
-                isLoggedIn = false;
+        function errorPopup(message, errorStatus, pageName, redirectLink, subMessage, stackTrace, errorCode, isDismissible, showLogin) {
```
RFSH: Based on my other comment on the `modal.js`, if we pass the actual object then the `message` and `errorStatus` seem more like optional parameters to this functions. Because only in one case we're actually changing them. So we can move them to the end of the argument list and they can be undefined.

PR #1513, opened 2018-05-08
Issue 1511

RFSH:
For some issues that we faced in the past, I had to introduce an attribute in chaise that keeps track of applied filters. Since it was manually added in chaise, we had to make sure that it is always
aligned with the reference facets that ermrestjs is returning. In case of submitting the facets, I was changing that variable first and then was checking the reference for URL limit. So in that case
the reference and appliedFilters would be inconsistent and recordset would break.

I also noticed that we don't have any testcases for this. So I added two for the following scenarios:

1. Long search string (2000 characters)
2. Selecting a lot of facet options in the modal

The only problem with the first one is that I'm using `sendKeys` for now, which is trying to mimic human typing, so it is taking more than 30 seconds for it to finish writing the 2000 characters. I will try to find an alternative for this, otherwise I will remove the test case.

### DIFF
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -460,7 +460,8 @@
                             scope.facetColumn.column.groupAggregate.histogram(numBuckets, requestMin, requestMax).read().then(function (response) {
                                 if (scope.facetColumn.sourceReference.uri !== uri) {
                                     // return breaks out of the current callback function
-                                    return defer.resolve(false);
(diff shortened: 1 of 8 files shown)


jrchudy:
The current failing test case randomly failed for me earlier in Travis. It seems to be fine after rerunning the build. Going to document this in our wiki as a test case to possibly fix.

PR #1526, opened 2018-05-16
Histogram zoom buttons

jrchudy:
This includes a new font file that has 2 icons in it, `chaise-zoom-in` and `chaise-zoom-out`. 

Other changes in this PR include changes to the range inputs in the range picker:
  - reducing the max-width of inputs
  - move validation messages to the end of the inputs HTML so that they don't interfere with the inputs moving around
  - text in input boxes flows before the `x` for clearing the data
  - Integer min and max enum created


common/styles/app.css
```diff
@@ -1,3 +1,35 @@
+@font-face {
+    font-family: 'chaise-icons';
+
+    src: url('fonts/chaise-icons.eot');
```
RFSH: This line is not needed.

PR #1534, opened 2018-05-24
Filename ext filter support

jrchudy:
This PR adds validation to `recordedit` to verify if the user selected files to upload that are valid based on model annotations.

### DIFF
--- a/common/upload.js
+++ b/common/upload.js
@@ -44,6 +44,14 @@
 
                     }, 10);
 
+                    var fileExtensionFilter = scope.column.filenameExtFilter;
+                    // needs to be a comma separated list, i.e. ".jpg", ".png", ...
+                    scope.fileExtensions = "";
+                    if (fileExtensionFilter) {
+                        for (var j=0; j<fileExtensionFilter.length; j++) {
+                            scope.fileExtensions += (j !== 0 ? "," : "" ) + fileExtensionFilter[j];
+                        }
+                    }
 
                     scope.select = function() {
                         scope.fileEl.click();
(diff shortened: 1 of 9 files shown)


common/upload.js
```diff
@@ -44,6 +44,14 @@
 
                     }, 10);
 
+                    var fileExtensionFilter = scope.column.filenameExtFilter;
+                    // needs to be a comma separated list, i.e. ".jpg", ".png", ...
+                    scope.fileExtensions = "";
+                    if (fileExtensionFilter) {
+                        for (var j=0; j<fileExtensionFilter.length; j++) {
```
RFSH: I think you can simplify this logic by using the `join` function (`scope.column.filenameExtFilter.join(",")`).

Also we are doing this in two different places, in `form.controller.js` and  `upload.js`. We can modify the ermrestjs API to actually return a string of comma separated values instead (I'm not sure whether we should do that or not. That's just a suggestion).

common/validators.js
```diff
@@ -169,5 +169,34 @@
...
+                        // consider empty models to be valid
+                        // use the `required` attribute in the HTML
+                        return true;
+                    }
+                    var fileExtensionFilter = scope.column.filenameExtFilter;
+                    if (fileExtensionFilter) {
+                        var isMatch = false;
+                        for (var j=0; j<fileExtensionFilter.length; j++) {
+                            if (modelValue.slice(modelValue.length - fileExtensionFilter[j].length, modelValue.length) == fileExtensionFilter[j]) {
+                                isMatch = true
```
RFSH: You can `return true` from here and change the `return isMatch` to always `return false`.

common/upload.js
RFSH: I just realized that the ermrestjs api is allowing string value for `filenameExtFilter`. We should change that to always return array. If it's a string, we should turn it into an array of length one.

Issue #1536, opened 2018-05-25
better error message


hongsudt:
Currently when an anonymous user visit a page that requires login, this is what we throw:
```
Table Monthly_Submission_Dashboard not found in schema. Please login to continue.

Click OK to go to the Home Page.
```

Please change this to.. 
```
Table Monthly_Submission_Dashboard not found in schema. You may need to login to see the table. Please login to continue.
```

hongsudt:
@jrchudy @RFSH can one of you do this when you have a chance.. I just need one more sentence added to make the error message easier to understand..

RFSH:
All the errors regarding "schema/table/column not found" are coming from ermrestjs as 404. So chaise cannot know if it's the table that is missing or schema. The error message that you suggested is assuming that it's the table that doesn't exist. 

If you want it to be expressive, we have to change those error objects in ermrestjs first. If you want a temporary and fast solution we can use "You may need to login to see the model" or something similar.

RFSH:
The extra message of "You may need to login to see the model or data" has been added to not-found  errors. Assigning to @hongsudt to verify.

RFSH:
To test this you can go to a page with any arbitrary [schema name](https://dev.rebuildingakidney.org/chaise/recordset/#2/Antibodyaaa:Antibodies) or [table name](https://dev.rebuildingakidney.org/chaise/recordset/#2/Antibody:Antibodies12312321).

PR #1543, opened 2018-05-30
add login message for 404 errors

RFSH:
Based on the #1536 issue.

### DIFF
--- a/common/modal.js
+++ b/common/modal.js
@@ -58,28 +58,34 @@
         var exception = params.exception,
             reloadMessage = ' <p>  </p>';
 
-        if (ERMrest && exception instanceof ERMrest.NotFoundError && vm.params.showLogin) {
-           vm.params.message = vm.params.message + messageMap.maybeNeedLogin;
+        // in case of adding login button, we should add extra message
+        if (vm.params.showLogin) {
+            if (ERMrest && exception instanceof ERMrest.NotFoundError) {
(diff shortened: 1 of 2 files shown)


common/modal.js
```diff
@@ -58,6 +58,10 @@
         var exception = params.exception,
             reloadMessage = ' <p>  </p>';
 
+        if (ERMrest && exception instanceof ERMrest.NotFoundError && vm.params.showLogin) {
```
jrchudy: I think this should be moved down to the if/else statement below. The conditional can be written like line 72 and the modifying the message like line 71.
```
} else if (ERMrest && exception instanceof ERMrest.NotFoundError) {
    if (vm.params.showLogin) vm.params.message = vm.params.message + messageMap.maybeNeedLogin;
} else if ...
```

RFSH: I didn't add it as `else if` case because [the logic in the `else` case](https://github.com/informatics-isi-edu/chaise/blob/e80396ead64ea82770bb3d38a4a205f45389c293/common/modal.js#L76,L82) still applies to this case.

RFSH: I could duplicate those lines in the the `else if` that I'm adding but that seemed wrong to me.

jrchudy: Ahh, okay. That makes sense. The `if/else` is meant for conditionally setting the `clickActionMessage`. So maybe it would instead make sense to create an `if/else` statement (off of the one you added) that includes other changes to the message variable.

PR #1544, opened 2018-05-30
recordedit error check login state first

jrchudy:
When records are created or updated, and we receive an error, we weren't checking to make sure the user still had an active session. When the user doesn't have an active session (and it's a conflict error... maybe this should be generalized but the use case was for a conflict error), we are now presenting a popup modal for them to go through the login flow.

Note: This reloads the page. There's some refactoring that needs to be done for the session service so this can work without reloading the page.

### DIFF
--- a/common/authen.js
+++ b/common/authen.js
@@ -223,6 +223,10 @@
             });
         };
 
+        /**
+         * uses the reloadCb function to reload the page no matter what after a user logs in
+         * creates a race condition with the calback registered for the LoginInAModal function
+         */
         var popupLogin = function () {
             var reloadCb = function(){
                 window.location.reload();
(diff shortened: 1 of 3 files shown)


jrchudy:
@RFSH There's still another case that is not being handled by these changes. If you have a valid session and start creating a record then get logged out, the foreign key popups still work as if you are logged in. This makes sense because those requests don't rely on the user being logged in.

I believe we want to handle this case (based on what we discussed with @hongsudt). We might have to modify the http wrapper function in `http.js` in `ermrestJS` [here](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/http.js#L175). 

I think we can loop through the args, look for the context header, check the cid and if the cid is `recordedit`, we should do a session verification before calling `fn.apply`.

jrchudy:
@RFSH changes that we discussed today have been added. Let me know if I forgot something we talked about or if any of the comments require further clarification.

PR #1548, opened 2018-06-05
Schema table column classes

jrchudy:
Opening this PR to make sure test cases still pass. Will update body tomorrow morning.

Related wiki page:
https://github.com/informatics-isi-edu/chaise/wiki/Custom-CSS-Styling-tips

@RFSH  what should we do about the function that hashes the pseudocolumn names? Should that be exposed like we discussed the other day so whomever can take a path and know what the output hash will be?

### DIFF
--- a/common/table.js
+++ b/common/table.js
@@ -937,7 +937,7 @@
         };
     }])
 
-    .directive('recordTable', ['recordTableUtils', function(recordTableUtils) {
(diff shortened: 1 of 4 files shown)


RFSH:
The changes look good. The only thing that bothers me is the "General HTML structure" in the wiki page. It's not true for all the apps. I feel like we should remove that for now since it can be misleading until we actually unify the HTML structure of our three apps.

And about the hash function, I'll make it public (it just requires changing the name of the function and attaching it to module). I will update the wiki when I did that.

PR #1551, opened 2018-06-05
Test sorting in recordset

Divyata-Singh:
- Added a test case to test sorting and paging in recordset in presentation.spec.js. Sorting tested on a text column (title), numeric column(no_of_rooms), pseudo-columns(category, type_of_facilities).
- Added a foreign key pseudo-column: category to the accommodation table for testing sorting on the same
-  Added a multi-hop foreign key pseudo-column: Type of Facilities to the accommodation table for testing sorting on the same


test/e2e/data_setup/schema/recordset/product-recordset.json
```diff
@@ -825,7 +869,165 @@
           "display": "Categories"
         },
         "tag:isrd.isi.edu,2016:visible-columns": {
-          "*": ["id", "term"]
+          "*": ["id", "term", "type"]
+        },
+        "tag:isrd.isi.edu,2016:visible-foreign-keys" : {
```
jrchudy: This annotation is unnecessary. We currently only use this annotation to control the list of related tables in record app.

test/e2e/data_setup/schema/recordset/product-recordset.json
```diff
@@ -825,7 +869,165 @@
...
+        },
+        {
+          "comment": "vocabulary",
+          "annotations": {},
+          "unique_columns": [
+            "type"
+          ]
+        }
+      ],
+      "entityCount": 0,
```
jrchudy: For future reference, there's no need to include this property if it's not being used.

test/e2e/data_setup/schema/recordset/product-recordset.json
```diff
@@ -825,7 +869,165 @@
...
+      "foreign_keys": [],
+      "table_name": "facilities",
+      "schema_name": "product-recordset",
+      "column_definitions": [
+        {
+          "comment": null,
```
jrchudy: For future reference, there's no need to include this property if it's not being used.

test/e2e/data_setup/schema/recordset/product-recordset.json
```diff
@@ -825,7 +869,165 @@
...
+          "name": "id",
+          "default": null,
```
jrchudy: This won't make a difference in the tests because the data you added always has ids, but this line conflicts with the next one. Having `nullok: false` means that the value **CANNOT** be null, but this line says to set it to `null` if it isn't present. It creates a confusing template.

If we were to look at this table in `recordedit` with both being defined, the UX will always enforce the `nullok: false` as a required field so data (for this column) won't ever be able to be submitted with `id = null`.

test/e2e/utils/chaise.page.js
```diff
@@ -890,6 +890,18 @@ var recordsetPage = function() {
         return element.all(by.css(".table-column-displayname > span"));
     };
 
+    this.getColumnSortButton = function(index){
+        return element(by.id("divRecordSet")).all(by.css('table > thead > tr > th')).get(index).element(by.css('.glyphicon-sort'));
```
jrchudy: This morning I just merged some new changes to the table directive that adds classes to each `table > thead > tr > th`.  If you pull those changes into your branch, you can simplify all three of these selectors to be:
```
this.getColumnSortButton = function(rawColumnName){
  return element(by.css('.' + rawColumnName)).element(by.css('.glyphicon-sort-by-attributes-alt'));
}
```
or simplify it more into a single selector:
```
element(by.css('.' + rawColumnName + ' .glyphicon-sort-by-attributes-alt'));
```

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
...
+                browser.wait(EC.presenceOf(chaisePage.recordsetPage.getRows().get(2)), browser.params.defaultTimeout);
+                for (var j = 0; j < accommodationParams.sortedData.length; j++) {
+                    (function (k) {
+                        //Click on sort button
+                        chaisePage.recordsetPage.getColumnSortButton(accommodationParams.sortedData[k].columnPosition).click().then(function () {
+                            browser.wait(EC.textToBePresentInElement(element(by.id('rs-total-count')),"Displaying 3 of 5 Records"), browser.params.defaultTimeout);
+                            return chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                        }).then(function () {
+                            // Check if the url has @sort by column name
+                            expect(browser.getCurrentUrl()).toContain('@sort(' + accommodationParams.sortedData[k].nameInUrl + ',id)');
```
jrchudy: Please add a failure message to this expectation check.
```
expect(true).toBe(true, "some failure message");
```

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
...
+                            });
+                        }).then(function () {
+                            // Go to the next page
+                            return chaisePage.recordsetPage.getNextButton().click();
+                        }).then(function () {
+                            browser.wait(EC.textToBePresentInElement(element(by.id('rs-total-count')),"Displaying 2 of 5 Records"), browser.params.defaultTimeout);
+                            return chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                        }).then(function () {
+                            // Check if the url has @sort by column name
+                            expect(browser.getCurrentUrl()).toContain('@sort(' + accommodationParams.sortedData[k].nameInUrl + ',id)');
```
jrchudy: Same comment as above about including a failure message.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
...
+                chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                browser.wait(EC.presenceOf(chaisePage.recordsetPage.getRows().get(2)), browser.params.defaultTimeout);
+                for (var j = 0; j < accommodationParams.sortedData.length; j++) {
+                    (function (k) {
+                        //Click on sort button
+                        chaisePage.recordsetPage.getColumnSortButton(accommodationParams.sortedData[k].columnPosition).click().then(function () {
+                            browser.wait(EC.textToBePresentInElement(element(by.id('rs-total-count')),"Displaying 3 of 5 Records"), browser.params.defaultTimeout);
+                            return chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                        }).then(function () {
+                            // Check if the url has @sort by column name
```
jrchudy: After checking if the url has the right `@sort identifier`, add an expectation to verify the proper glyphicon is shown in the column header. This will be relevant below as well for the descending case.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
...
+                            });
+                        }).then(function () {
+                            // Click on sort button to sort in descending order. This redirects to the first page.
+                            return chaisePage.recordsetPage.getColumnSortDescButton(accommodationParams.sortedData[k].columnPosition).click();
+                        }).then(function () {
+                            browser.wait(EC.textToBePresentInElement(element(by.id('rs-total-count')),"Displaying 3 of 5 Records"), browser.params.defaultTimeout);
+                            return chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                        }).then(function () {
+                            // Check if the url has @sort by column name
+                            expect(browser.getCurrentUrl()).toContain('@sort(' + accommodationParams.sortedData[k].nameInUrl + '::desc::,id)');
```
jrchudy: failure message comment.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
...
+                                }
+                            });
+                        }).then(function () {
+                            // Click on sort button to sort in descending order. This redirects to the first page.
+                            return chaisePage.recordsetPage.getColumnSortDescButton(accommodationParams.sortedData[k].columnPosition).click();
+                        }).then(function () {
+                            browser.wait(EC.textToBePresentInElement(element(by.id('rs-total-count')),"Displaying 3 of 5 Records"), browser.params.defaultTimeout);
+                            return chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                        }).then(function () {
+                            // Check if the url has @sort by column name
```
jrchudy: add glyphicon expectation after the sort expectation

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
...
+                            });
+                        }).then(function(){
+                            // Go to the previous page
+                            return chaisePage.recordsetPage.getPreviousButton().click();
+                        }).then(function () {
+                            browser.wait(EC.textToBePresentInElement(element(by.id('rs-total-count')),"Displaying 3 of 5 Records"), browser.params.defaultTimeout);
+                            return chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                        }).then(function () {
+                            // Sanity check on the previous page
+                            expect(browser.getCurrentUrl()).toContain('@sort(' + accommodationParams.sortedData[k].nameInUrl + '::desc::,id)');
```
jrchudy: last comment about failure message

test/e2e/utils/chaise.page.js
jrchudy: I think `rawColumnName` is the same as the expected data property you added called `nameInUrl`.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -33,56 +37,134 @@ var testParams = {
...
                 title: "Sherathon Hotel",
-                website: "http://www.starwoodhotels.com/sheraton/index.html",
+                website: "http://www.starwoodhotels.com/sheraton/index.html", 
                 rating: "4.3000",
                 summary: "Sherathon Hotels is an international hotel company with more than 990 locations in 73 countries. The first Radisson Hotel was built in 1909 in Minneapolis, Minnesota, US. It is named after the 17th-century French explorer Pierre-Esprit Radisson.",
                 opened_on: "2008-12-09 00:00:00",
                 luxurious: "true",
                 json_col: JSON.stringify(null,undefined,2),
-                json_col_with_markdown: "Status is: “delivered”",
+                json_col_with_markdown: "Status is: “delivered”",
```
jrchudy: It looks like you added a few extra whitespaces to the end of this line. After highlighting it:
image

Is there an option in your editor to remove trailing whitespace on save? This will prevent git from flagging some lines as changed when it was only whitespace.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
...
             });
 
+            it("apply sorting on different columns.", function () {
+                browser.get(browser.params.url + "/recordset/#" + browser.params.catalogId + "/product-recordset:" + accommodationParams.table_name + "?limit=3");
+                var EC = protractor.ExpectedConditions;
+                chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));
+                browser.wait(EC.presenceOf(chaisePage.recordsetPage.getRows().get(2)), browser.params.defaultTimeout);
+                for (var j = 0; j < accommodationParams.sortedData.length; j++) {
+                    (function (k) {
+                        //Click on sort button
```
jrchudy: Before clicking the sort button, let's verify the proper glyphicon is shown when no sorting is applied.

RFSH:
When I was looking at the issue description, I didn't realize the test cases that we are adding are going to be this big and we're writing them as one promise chain inside an `it` block. Now that I looked at your implementation, we should change the `it` into a `describe` and break the promise chain into multiple `it`s. 

If you looked at the other test cases, we have these sort of promise chains but they are very fragile. The wait time for each promise will be added to the next one and it will make the test to break easily. That's why we keep trying to break these into multiple `it`s. This will also make the test cases more readable. I suggest the following structure:

```javascript
describe("testing sort and paging features," , function () {
    beforeAll(function () {
        browser.get(browser.params.url + "/recordset/#" + browser.params.catalogId + "/product-recordset:" + accommodationParams.table_name + "?limit=3");

        // wait for spinner
        chaisePage.waitForElementInverse(element.all(by.id("spinner")).get(0));

        // make sure data has been loaded
        var EC = protractor.ExpectedConditions;
        browser.wait(EC.presenceOf(chaisePage.recordsetPage.getRows().get(2)), browser.params.defaultTimeout);
    });

    accommodationParams.sortedData.forEach(function (data, k) {
        describe ("for column `" + data.columnName +  "`,", function () {
            it ("clicking on sort should change the displayed row.", function (done) {
                // ...
            });

            it ("clicking on next should display the next page.", function (done) {
                // ...
            });

            it ("clicking on sort in descending order, should go to the first page.", function (done) {
                // ..
            });

            it ("clicking on next should display the next page.", function (done) {
                // ...
            });

            it ("clicking on previous should display the previous page.", function (done) {
                // ...
            });
        });
    });
});
```

Also notice the `done` input for the `it`s. This function helps with the asynchronous requests. If you have `done` defined as a parameter and forget to call it in the `it` block, the block won't be completed. This just just to ensure the asynchronous task of the previous `it` has finished. So at the end of each promise chain inside the `it` you should call this function.

```javascript
it ("message", function (done) {
  req1().then(function (){
    return req2();
  }).then(function () {
    // extra stuff
   done();
  }).catch(function (err) {
    done.fail(err);
  });
})
``` 
Also the `done.fail()` will force the failure of the test case. We have documented these [in here](https://github.com/informatics-isi-edu/ermrestjs/wiki/Test-idioms), but haven't add a link to the wiki yet. I am going to add a link from the e2e test guid to this wiki.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -416,6 +502,100 @@ describe('View recordset,', function() {
                 });
             });
 
+            it("apply sorting on different columns.", function () {
```
jrchudy: After talking to @RFSH about this, we think this `it()` should be split into 2 separate `it()` cases. One testing sorting in ascending order and one testing sorting in descending order.

This should be done by changing this structure to something like:
```js
describe("", function () { 
  forEach(accommodationParams.sortedData, function (data) {
    (function (k) {
      it("do ascending test", function () {...});
      it("do descending test", function () {...});
    }(j));
  }
})
```

jrchudy:
I removed my comment from above. We should follow what Aref suggested for this.

Divyata-Singh:
I think formatting the test case as ascending and descending makes more sense rather than creating an array of dependent test cases. I discussed this with Aref and he suggested if the test cases are stable then it is fine. Though I have formatted them to have smaller promise chains.

RFSH:
@Divyata-Singh @jrchudy We can discuss about this tomorrow in the meeting. For now, you can finalize the PR based on what you think is the best format. If we break it as much as I suggested then the `it`s will be very dependent on each other. If one fails, all the other ones will fail too. On the other hand it might make it easier to read and follow. You can use your judgment on how much you need to break the test cases and then tomorrow we can talk about it.

PR #1553, opened 2018-06-06
handle null values for min and max in histogram

RFSH:
It used to throw terminal error since we have not covered that case. I did a bit of refactoring to make it easier.

I tried to explain the error in the issue description. To summarize, in case of timestamp and null value, we were not setting values for `absMin` and `absMax` and that was causing problems.

### DIFF
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -401,10 +401,7 @@
                     // range is defined by the x values of the bar graph because layout.xaxis.type is `linear` or 'category'
                     function setHistogramRange() {
                         if (isColumnOfType("timestamp")) {
-                            var minMoment = moment(scope.rangeOptions.absMin.date + scope.rangeOptions.absMin.time, dataFormats.date + dataFormats.time24);
-                            var maxMoment = moment(scope.rangeOptions.absMax.date + scope.rangeOptions.absMax.time, dataFormats.date + dataFormats.time24);
-
-                            scope.plot.layout.xaxis.range = [minMoment.format(dataFormats.datetime.submission), maxMoment.format(dataFormats.datetime.submission)];
+                            scope.plot.layout.xaxis.range = [dateTimeToTimestamp(scope.rangeOptions.absMin),  dateTimeToTimestamp(scope.rangeOptions.absMax)];
                         } else {
                             scope.plot.layout.xaxis.range = [scope.rangeOptions.absMin, scope.rangeOptions.absMax];
                         }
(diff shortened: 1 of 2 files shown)


common/faceting.js
```diff
@@ -452,12 +437,39 @@
...
+                    }
+
+                    /**
+                     * Given a string representing timestamp will turn it into an object with `date` and `time` attributes
+                     * @param  {string} ts timestamp
+                     * @return {object} an object with `date` and `time` attributes
+                     * NOTE might return `null`
+                     */
+                    function timestampToDateTime(ts) {
+                        if (!ts) return null;
```
jrchudy: Thinking about this and what I said to you in slack earlier, should the `null` case return the empty object?
```
obj = {
  date: null,
  time: null
}
```

common/faceting.js
```diff
@@ -452,12 +437,39 @@
...
                     }
 
+                    /**
+                     * Given an object with `date` and `time`, it will turn it into timestamp reprsentation of datetime.
+                     * @param  {object} obj The datetime object with `date` and `time` attributes
+                     * @return {string} timestamp in submission format
+                     * NOTE might return `null`
+                     */
+                    function dateTimeToTimestamp(obj) {
+                        if (!obj) return null;
```
jrchudy: Having this guard is good because it protects against calling `absMin.date` when absMin is `undefined` or `null`.

common/faceting.js
```diff
@@ -687,8 +694,10 @@
                                             scope.rangeOptions.absMin = moment(min).format(dataFormats.date);
                                         } else if(isColumnOfType("timestamp")) {
                                             var minMoment = moment(min);
-                                            scope.rangeOptions.absMin.date = minMoment.format(dataFormats.date);
-                                            scope.rangeOptions.absMin.time = minMoment.format(dataFormats.time24);
+                                            scope.rangeOptions.absMin = {
```
jrchudy: Makes sense why this could break here with how it was defined before.

test/e2e/specs/delete-prohibited/recordset/histogram-facet.spec.js
```diff
@@ -300,5 +301,23 @@ describe("Viewing Recordset with Faceting,", function() {
...
+                chaisePage.waitForElementInverse(element(by.id("spinner")));
+            });
+
+            it ("no data should be represented to the user.", function () {
+                expect(chaisePage.recordsetPage.getRows().count()).toBe(0);
+            });
+
+            it ("no histogram should be displayed.", function () {
+                for(var i = 0; i < testParams.totalNumFacets; i++) {
+                    expect(chaisePage.recordsetPage.getHistogram(i).isPresent()).not.toBeTruthy("histogram is available for facet index=" + i);
```
jrchudy: This is subjective, but why use `.not.toBeTruthy()` when you could use `.toBeFalsy()`?

Note: to me, `toBeFalsy()` looks like it's spelt wrong and not a real function. But we do use it in numerous places in the tests.

RFSH: I forgot about `toBeFalsy`. I will change it.

Issue #1530, opened 2018-05-18
Add Login Link in Session Alert


RFSH:
Currently we're showing an alert when the user's session has expired. To see the behavior you need to either modify the expiration date of your cookie in database, or wait for it to expire. Clicking on the `x` will stop chaise from showing this alert.

The login in that message used to be a link, but it wasn't working because of the following reason:
We wanted to pass the alert to a `ng-bind-html` and alert message had `ng-click` on it. But it wouldn’t register the `ng-click` since the `ng-bind-html` is just adding html value and not running the Angualrjs logic again to register the click event. 

To fix it, we might want to have a special alert for this case. Instead of passing the message to alert service it should be its own specific type of alert that has its own template and logic. There might be better or easier solution to this. We need to investigate this.

To find the places that we're throwing this error just search for `Session.createPromptExpirationToken`. We're throwing this after resolving the reference in record and recordset app. The login button should call the `loginInAPopUp()` function that is defined in `Session` directive (defined in `authen.js`).

PR #1554, opened 2018-06-07
Added Login Link in Session Alert #1530

dhanashritidke:
Following changes have been made to add login link in Session Alert : 
**alerts.js**
1. Added a new compile directive which uses the internal angular service $compiler to bind the scope with the element.
2. Added controller to fire the method : Session.loginInAPopUp();

**alerts.html**
3. Changed the tag "ng-bind-html" to "compile". This is done so that the "ng-click" on the Log in link gets registered.
4. Added the corresponding "ng-controller" div. 

**utils.js**
3. Added anchor <a> tag so the text "Log in" is now a clickable link.

### DIFF
--- a/common/alerts.js
+++ b/common/alerts.js
@@ -52,5 +52,21 @@
                 };
             }
         };
-    }]);
+    }])
+    .directive('compile', ['$compile', function ($compile) {
+      return function(scope, element, attrs) {
+        scope.$watch(
+          function(scope) {
+            return scope.$eval(attrs.compile);
+          },
+          function(value) {
+            element.html(value);
+            $compile(element.contents())(scope);
+          }
+       )};
+     }]).controller('alertController', function($scope, Session) {
+        $scope.login = function() {
+          Session.loginInAPopUp();
+        }
+    });
 })();
--- a/common/alerts.js
+++ b/common/alerts.js
@@ -50,23 +50,10 @@
                 scope.closeAlert = function(alert) {
                     AlertsService.deleteAlert(alert);
                 };
+                scope.login = function() {
+                  Session.loginInAPopUp();
+                }
             }
         };
-    }])
-    .directive('compile', ['$compile', function ($compile) {
-      return function(scope, element, attrs) {
-        scope.$watch(
-          function(scope) {
-            return scope.$eval(attrs.compile);
-          },
-          function(value) {
-            element.html(value);
-            $compile(element.contents())(scope);
-          }
-       )};
-     }]).controller('alertController', function($scope, Session) {
-        $scope.login = function() {
-          Session.loginInAPopUp();
-        }
-    });
+    }]);
 })();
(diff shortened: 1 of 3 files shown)


common/alerts.js
```diff
@@ -52,5 +52,21 @@
...
+        scope.$watch(
+          function(scope) {
+            return scope.$eval(attrs.compile);
+          },
+          function(value) {
+            element.html(value);
+            $compile(element.contents())(scope);
+          }
+       )};
+     }]).controller('alertController', function($scope, Session) {
```
RFSH: You don't need to define a controller here. You could have the `login` function definition to the `link` function of `alerts` directive.

common/alerts.js
```diff
@@ -52,5 +52,21 @@
                 };
             }
         };
-    }]);
+    }])
+    .directive('compile', ['$compile', function ($compile) {
```
RFSH: We usually define these utility components in the `common/utils.js`. I think we should move this directive to there. If you can add a few comments about how it works would be good too. Most of our directives and components lack documentation which just makes it harder to figure out what they're supposed to do. We're not using jsdoc in chaise, but you can use the same syntax.

```javascript

/**
* THE DOCUMENTATION GOES HERE
*/
.directive('compile', ['$compile', function ($compile) {
...
})
```

common/alerts.js
```diff
@@ -52,5 +52,21 @@
...
+      return function(scope, element, attrs) {
```
jrchudy: To follow the way we have defined each of our directives, you should change the `return` to return an object:
```
return {
  restrict: `A`,     // could also be `E` (element), `C` (class), `M` (comment`
  link: function(scope, element, attrs) {
    ...
  }
}
```

common/alerts.js
```diff
@@ -52,5 +52,21 @@
...
+        scope.$watch(
+          function(scope) {
+            return scope.$eval(attrs.compile);
```
jrchudy: I understand that `attrs.compile` is the html string that you are passing as a variable to the `compile` attribute. 

What is the return value of `$eval` when that value is `null` or empty? Does it make sense to guard against running compile against either of those cases?

common/alerts.js
```diff
@@ -52,5 +52,21 @@
...
+          },
+          function(value) {
+            element.html(value);
```
jrchudy: What is the evaluated value here? Is it the HTML string we passed to the attribute? If that's true, is this replacing the HTML for the element (which is fine in this case because the span tag is meant to only contain the content we are trying to compile here)?

common/alerts.js
```diff
@@ -52,5 +52,21 @@
...
+            $compile(element.contents())(scope);
```
jrchudy: Could we write this as:
```
var compileValue = $compile(value)(scope);
element.append(compiledValue); // or element.html(compiledValue)
```

dhanashritidke:
I am done with the changes mentioned above. I have also added detailed documentation for the directive. I believe it should answer all the questions mentioned above.

jrchudy:
Looks good. I wouldn't worry about the failing test case. I am aware of it's instability (doing binning on timestamp columns create different behaviors locally and in travis).

PR #1556, opened 2018-06-08
Empty hatrac files

jrchudy:
When a file with the same file content, same name, and same size are uploaded to hatrac, it creates a new version of the file that has size 0. This was happening because the code properly didn't re-upload the same content. But it didn't clean up the job after wards.

This PR solves this by verifying if the file exists before creating the upload job. If it does already exist (with the same md5, filename, and size), the creation of the job is skipped and the file object in chaise is marked as complete.

Another consideration, was not skipping the job creation step and instead removing those empty versioned namespaces after the upload has completed. Doing some cleanup afterwards. I didn't think this would be better than not creating the job at all.


RFSH:
Can you please add this new check to [the documentation](https://github.com/informatics-isi-edu/chaise/blob/6465bbbeab09cf8a71285292a14ab8e68f17697f/common/upload.js#L65,L93) of how this directive works? Now that you're checking the upload job, what's the point of `checkFileExists` that is being done after creating the upload job?

common/upload.js
```diff
@@ -376,12 +403,40 @@
...
             };
 
+            uploadFile.prototype.onPreFileExistSuccess = function() {
+
+                if (vm.erred || vm.aborted) return;
+
+                this.skipUploadJob = this.hatracObj.jobDone;
+
+                // if the job is already done, that means the file has an idntical file already in the server
+                // md5, filename, and size all match
```
RFSH: Since this check is in ermrestjs and I think we might want to change it (removing check for the filename), let's just say identical file and remove the actual details of what is the definition of identical file. (we don't need to mention we're checking for md5, filename, etc.)

common/upload.js
```diff
@@ -106,22 +106,25 @@
...
+
             vm.createUploadJobProgress = 0;
             vm.createUploadJobCompleted = 0;
 
             vm.fileExistsProgress = 0;
             vm.fileExistsCompleted = 0;
 
             vm.uploadProgress = 0;
-            vm.uploadCompleted = 0;
+            vm.numUploadsCompleted = 0;
```
jrchudy: renamed this because there was a boolean with the same name attached to the `uploadFile` object.

jrchudy:
After some discussion with @karlcz I realized I had configured my ACLs in hatrac improperly. I had `subtree-create` defined for all the sub directories (namespaces) but never changed the permissions on the individual job that was created by another user.

Adding the `update` ACL to the job allowed me to reproduce the 403 response for the HEAD request again and then to properly create a new job (same name but with a different version).

The error case that I outlines that falls out from this case is still true and will be noted in a separate issue to be addressed later.

Issue #1519, opened 2018-05-10
Improve Handling of url Limitation in Recordset


RFSH:
Currently we're checking for the url limit after users click on the submit button of the modal picker for facets. We should check for the url limitation on each select.

RFSH:
We should check the url limitation on each change of the filters. Upon reaching the limit, we should show an alert and disable the submit button. We should hide the alert and enable submit button as soon as a change in the filters made the url to be shorter than the limit.

PR #1561, opened 2018-06-12
Improve URL Limitation Handling In Recordset

RFSH:
This PR will improve the way we're showing URL limitation alert. We used to show the alert after users clicks on submit (in facet modal). But now, we're checking the url on each change on the selected rows.

Issues: #1519 #1216

### DIFF
--- a/common/alerts.js
+++ b/common/alerts.js
@@ -3,9 +3,10 @@
 
     angular.module('chaise.alerts', ['chaise.filters', 'chaise.utils'])
 
-    .factory('AlertsService', ['DataUtils', function AlertsService(DataUtils) {
+    .factory('AlertsService', ['DataUtils', 'messageMap', function AlertsService(DataUtils, messageMap) {
         var alerts = [];
         var ALERT_TYPES = ['success', 'error', 'warning'];
+        var urlLimitAlert;
 
         function Alert(message, type, cb) {
             DataUtils.verify(message, 'Message required to create an alert.');
@@ -31,10 +32,34 @@
             return alerts.splice(index, 1);
         }
 
+        /**
+         * Will create a specific type of alert for url limitation
+         * @param {string} message
+         */
+        function addURLLimitAlert(message) {
+            message = message || messageMap.URLLimitMessage;
+            if (urlLimitAlert) return;
+            urlLimitAlert = addAlert(message, ALERT_TYPES[2]);
+            return urlLimitAlert;
+        }
+
+        /**
+         * Delete the specific alert
+         * @return {Alert}
+         */
+        function deleteURLLimitAlert() {
+            if (!urlLimitAlert) return;
+            var alert = deleteAlert(urlLimitAlert);
+            urlLimitAlert = null;
+            return alert;
+        }
+
         return {
             alerts: alerts,
             addAlert: addAlert,
-            deleteAlert: deleteAlert
+            deleteAlert: deleteAlert,
+            addURLLimitAlert: addURLLimitAlert,
+            deleteURLLimitAlert: deleteURLLimitAlert
         };
     }])
 
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -919,6 +919,9 @@
             /**
              * Post process after selectedRows is defined coming from the modal.
              * If changeRef is false, then we only want to apply the URL limitation logic.
+             * the callback that is returning is accepting an array or an object with `matchNotNull` attribute.
+             * If the attribute exists, then we want to match any values apart from null. otherwise the
+             * parameter will be an array of tuples.
              * @param  {object} scope
              * @param  {boolean} changeRef whether we should change the reference or not
              */
@@ -1040,9 +1043,18 @@
                         params.facetPanelOpen = false;
 
                         // callback on each selected change (incldues the url limitation logic)
-                        params.onRowClick = modalDataChanged(scope, false);
+                        params.onSelectedRowsChanged = modalDataChanged(scope, false);
 
                         // if url limitation alert exists, remove it.
+                        // The alert on the main recordset page is behaving differently
+                        // from the alert that we are going to show on modal.
+                        // We're showing alert as a preventing measure in recordset.
+                        // if users are about to reach the limit, upon making the request
+                        // we're showing the modal and ignoring the request. So the alert
+                        // is just to tell users why they couldn't do the action and it
+                        // doesn't have to remain on the page.
+                        // While the alert on modal must stay untill they actually remove
+                        // some selections and url becomes shorter than the limit.
                         AlertsService.deleteURLLimitAlert();
 
                         // to choose the correct directive
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -731,7 +731,9 @@
             };
         }])
 
-        .directive('choicePicker', ["defaultDisplayname", 'logActions', "$log", 'modalUtils', '$q', 'tableConstants', '$timeout', function (defaultDisplayname, logActions, $log, modalUtils, $q, tableConstants, $timeout) {
+        .directive('choicePicker',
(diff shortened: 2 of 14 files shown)


common/alerts.js
```diff
@@ -31,10 +32,34 @@
             return alerts.splice(index, 1);
         }
 
+        /**
+         * Will create a specific type of alert for url limitation
+         * @param {string} message
+         */
+        function addURLLimitAlert(message) {
```
jrchudy: Is this a wrapper so that you can check if the alert already exists, and if it does don't add a new one?

common/faceting.js
```diff
@@ -966,6 +1039,12 @@
                         params.faceting = false;
                         params.facetPanelOpen = false;
 
+                        // callback on each selected change (incldues the url limitation logic)
+                        params.onRowClick = modalDataChanged(scope, false);
+
+                        // if url limitation alert exists, remove it.
+                        AlertsService.deleteURLLimitAlert();
```
jrchudy: Is this done because you can hit the length limitation on recordset with faceting, which will show the alert?

And then if you click "Show details ..." it would still show that alert, even though the multi-select in the popup can have the same issue?

common/modal.js
```diff
@@ -181,22 +181,35 @@
...
-            $uibModalInstance.close(res);
+            return res;
+        }
+
+        /**
+         * Will call the close modal with the appropriate results.
+         * If we had the matchNotNull, then we just need to pass that attribute.
+         */
+        function submitMultiSelection() {
+            $uibModalInstance.close(getMultiSelectionResult());
```
jrchudy: How is the return value of `getMultiSelectionResult()` used int he modal close callback (which is `modalDataChanged(scope, true)`, on line 1092 of `faceting.js`)?

common/styles/app.css
```diff
@@ -1299,7 +1299,7 @@ footer .container1 {
      display: none !important;
  }
 
- .modal-body .recordset-container alerts {
+ #record .modal-body .recordset-container alerts {
```
jrchudy: Should this also apply to `recordedit`?

common/table.js
```diff
@@ -494,34 +494,53 @@
...
                 if (refUri.length > tableConstants.MAX_URL_LENGTH) {
-                    $timeout(function () {
-                        $window.scrollTo(0, 0);
-                    }, 0);
-                    AlertsService.addAlert('Maximum URL length reached. Cannot perform the requested action.', 'warning');
+                    AlertsService.addURLLimitAlert();
+                    scrollToTop();
                     return false;
                 }
+                AlertsService.deleteURLLimitAlert();
```
jrchudy: Meaning the user could have removed one of the selections so the above conditional will return false and we no longer want the user to see that alert?

common/table.js
```diff
@@ -792,17 +812,28 @@
...
+                    return;
+                }
+
+                var tuple = scope.vm.selectedRows.splice(index, 1)[0];
+                _callOnRowClick(scope, tuple, false);
             };
 
             // function for removing all pills regardless of what page they are on, clears the whole selectedRows array
-            scope.removeAllPills = function() {
+            scope.removeAllPills = function($event) {
```
jrchudy: not sure what the point is of passing `$event` here. Looks to be unnecessary unless it's a common angular idiom that we should have been following before..

common/alerts.js
RFSH: yeah. At first I was attaching the alert object to the root and checking it that way. This looks better and cleaner I think.

common/faceting.js
RFSH: If you hit the limit in the recordset, we will not allow you to perform that action. Assume the url length is 1950 and selecting one of the facets will increase the length to more than the limit. When you click on the check box, we will show the alert and uncheck the checkbox. So you cannot perform that action at all. But in the modal, you can choose as much filters as you want. We won't prevent you from selecting filters. Only the submit button will be disabled.

Therefore, if there's an alert on the main page, that was because you wanted to do something and we prevented you from doing it. We don't need to keep showing you the alert. So when you open the modal, we can remove that alert.

common/modal.js
RFSH: If you look at the recordsetSelectFaceting, it has an extra checkbox for matching not-null values. So apart from `selectedRows` the value of `matchNotNull` is important too. The `getMultiSelectionResult` is taking care of that. And then the `modalDataChanged` is looking for those two variables.

common/table.js
RFSH: correct. I will comment these questions that you asked.

common/styles/app.css
RFSH: Yes, I completely forgot about that. I did this change at the start of doing the changes and didn't care much about it. Now that I'm looking at it, I'm going to add extra class to the one that we want to show (recordsetSelectFaceting) and do the `display:none` for alerts that don't have that class name.

common/table.js
RFSH: At first, I was using the value that  `_callOnRowClick` is returning. If you look at the `modalDataChanged` which is eventually the function that is being called for clicking on row, it's returning `false` when we hit to the URL limit. In that case, I was actually calling `$event.preventDefault()` which would prevent the action that we were doing.

But later we decided to not call the `preventDefault`. I could remove the `$even` object, but later we might actually want to change this to do the `preventDefault` and I feel like there's no harm in passing the value.

common/table.js
```diff
@@ -494,34 +494,61 @@
...
+            } else if (scope.onSelectedRowsChanged) {
+                return scope.onSelectedRowsChanged()(tuples, isSelected);
+            }
+        }
+
+        /**
+         * Scrolls the current container to top. The container
+         * can be the modal-body or the main-container
+         */
+        function scrollToTop() {
```
jrchudy: This shouldn't necessarily be refactored here, but now we have multiple places with a `scrollToTop` function. We could change this to be a utility function that takes in the element you want to scroll to the top of.

test/e2e/specs/delete-prohibited/recordset/misc-facet.spec.js
```diff
@@ -407,40 +408,100 @@ describe("Other facet features, ", function() {
...
-                done();
-            }).catch(chaisePage.catchTestError(done));
+            it ('after opening the modal, the existing url limit alert should be removed.', function () {
+                expect(alert.isPresent()).toBeFalsy();
+            });
+
+            it ("alert should be displayed upon reaching the URL limit and submit button should be disabled.", function (done) {
+                chaisePage.clickButton(chaisePage.recordsetPage.getSelectAllBtn()).then(function () {
+                    checkAlert();
+                    expect(submitBtn.getAttribute('disabled')).toBe('true', "submit is not disabled.");
```
jrchudy: nitpick: `.toBeTruthy()` although it doesn't matter

test/e2e/specs/delete-prohibited/recordset/misc-facet.spec.js
```diff
@@ -407,40 +408,100 @@ describe("Other facet features, ", function() {
...
+                    done();
+                }).catch(chaisePage.catchTestError(done));
+            });
+
+            it ("changing filters and going below the URL limit should hide the alert and enable the submit button.", function (done) {
+                chaisePage.clickButton(chaisePage.recordsetPage.getModalRecordsetTableOptionByIndex(1)).then(function () {
+                    chaisePage.waitForElementInverse(alert);
+                    expect(submitBtn.getAttribute('disabled')).not.toBe('true', "submit is disabled.");
```
jrchudy: Same comment about `toBeTruthy()` except can be written as `toBeFalsy()`.

common/table.js
RFSH: I could leave this in the place that it is being called. The ` $window.scrollTo(0, 0);` wasn't doing anything since it's not the window that is scrollable but the container inside the window. I think I created the function because I thought it will be called multiple times.

PR #1571, opened 2018-06-20
change the url limitation test case

RFSH:
The purpose of this PR is just to make sure the test case is passing on travis before merging.

The current test case was not deterministic since for the url limitation we're looking at the whole url (including the hostname) which can be different. That was the main reason that locally the test cases were passing but in travis they didn't. I changed the url check to just check the path and then changed the data to be aligned with this new change.


test/e2e/specs/delete-prohibited/recordset/misc-facet.spec.js
```diff
@@ -73,7 +73,7 @@ var testParams = {
 
 
 describe("Other facet features, ", function() {
-
+/*
```
jrchudy: I don't think you meant to commit this.

test/e2e/specs/delete-prohibited/recordset/misc-facet.spec.js
```diff
@@ -503,7 +511,7 @@ describe("Other facet features, ", function() {
         });
 
     });
-
+/*
```
jrchudy: or this

PR #1576, opened 2018-06-21
moved Filetype validation

jrchudy:
filetype validation was triggering during edit mode. Moved this validation to the bind event for when the file input changes instead of the text input.

Made a change in `recordCreate` where an error without a message (I saw it as an error with code 0) would throw a terminal error.


jrchudy:
Last commit included a change that fixed a test that only fails locally.

PR #1582, opened 2018-06-27
Batch unlink

jrchudy:
This PR adds the batch unlink functionality to the `record` app. In the related table action bar, a new link, `Remove`, is available if the related table is an association table.

This PR requires changes that were made in the corresponding `ermrestJS` [PR](https://github.com/informatics-isi-edu/ermrestjs/pull/668).


common/ellipses.js
```diff
@@ -22,39 +22,65 @@
...
+                // user accepted prompt to delete
+                reference.delete(tuples, logObject).then(function deleteSuccess() {
+                    scope.$root.showSpinner = false;
+                    // tell parent controller data updated
+                    scope.$emit('record-deleted');
+                }).catch(onError());
+            }
+
+            var dataForDelete = {};
+            // if assocation, we have to grab the key information for the association table from
```
RFSH: I know when we say association we always mean pure and binary. But since the implementation of this is actually assuming pure and binary, it's better to emphasis that. Just add a note that you assumed this will be used only for pure and binary association and it will only work for that case (having two foreignkeys).

common/ellipses.js
```diff
@@ -22,39 +22,65 @@
...
                     scope.$root.showSpinner = false;
                     // if response is string, the modal has been dismissed
-                    if (typeof response !== "string") {
+                    if (!hasConfirm || typeof response !== "string") {
                         ErrorService.handleException(response, true);  // throw exception for dismissible pop- up (error, isDismissible = true)
                     }
                 }
+            }
+
+            var deleteTuple = function (hasConfirmation) {
```
RFSH: The following comments are not very important, but are things that I think can help the readability of code. They are subjective and you might disagree with me about them. In that case you can just ignore the comment.

- It's better to move all the variable declaration top of the function. For example the tuples is defined at the bottom of the function, but used in `deleteTuple`. It's a bit hard to follow where that is coming from. Also the name `tuples` is a bit misleading, you can just pass `[dataForDelete]`.

- I think the `forEach` syntax is easier to read. It also makes it easier since you don't have to repeat a lot. We should try using that as much as we can. If we want to move to ES6 at some point, it makes that transition easier too.
Instead of:
  ```javascript
    for (var k=0; k<fk.colset.columns.length; k++) {
      var col = fk.colset.columns[k];
      ...
    }
   ```
    It can be

    ```javascript
      fk.colset.columns.forEach(function (col) { 
          ...
     });
    ```


Again these are not that important and are somethings that came to my mind by looking at the code. You can change them or leave them the way they are, it's up to you.

record/record.controller.js
```diff
@@ -195,6 +200,16 @@
             return ref.canCreate;
         };
 
+        vm.canDeleteRelated = function(relatedRef) {
```
RFSH: We talk about this and you might have changed it already, but I'm going to comment for the future reference. I think we should use the same terminology at least in the code. If it's called `unlink` somewhere else, we should keep using that name. And if you want to change it, we should be consistent and change the other parts of the code.

record/record.controller.js
```diff
@@ -255,6 +270,70 @@
             $window.open(appLink, '_blank');
         };
 
+        vm.removeRelatedRecord = function(ref) {
```
RFSH: same comment about unlink/remove/delete name.

PR #1584, opened 2018-06-28
Updated Angular-ui-bootstrap from v1.3.2 to v2.5.0

Divyata-Singh:
- Updated Angular-ui-bootstrap from v1.3.2 to v2.5.0
- Updated the selectors used for getting facets and facet-titles


RFSH:
I know the old version was not minified, but I think now that we're making change to that dependency we should use the minified version. We're not going to change any part of that code, so I don't think it's valuable to have the original version (vs. minified).

PR #1585, opened 2018-06-28
Error spec random failures resolved

dhanashritidke:
Following modifications have been done to fix the issue : 
1. btn.click() was not getting registered in some cases so changed it to browser.executeScript("arguments[0].click();",btn);
2. Fixed chaining issues in some of the it blocks.
3. waitForElement(element) added in some of the failing test cases.
4. browser.ignoreSynchronization added for all describe blocks.
5. done() added in it blocks with chaining.
6. browser.refresh() added in the beforeAll() blocks of all the describe
blocks following a redirect to the recordset page. [This needs to be
investigated further]. This is the case where (sometimes) there is
timeout because on redirecting to the recordset page, it stays on the
app for too long causing the next describe block to fail.
7. Reordered the describe blocks for some fixing some test cases.

### DIFF
--- a/test/e2e/specs/all-features-confirmation/errors/errors.spec.js
+++ b/test/e2e/specs/all-features-confirmation/errors/errors.spec.js
@@ -63,12 +63,10 @@ describe('Error related test cases,', function() {
                 expect(showDetails.getText()).toBe(testParams.hideErrors, "The Show/Hide message in modal pop is not correct");
                 expect(errorDetails.getText()).toContain("Error", "error missmatch.");
                 done();
-            }).catch(function (err) {
-                done.fail(err);
-            });
+            }).catch(catchError(done));
         });
 
-        it('On click of OK button the page should redirect to recordset/search page', function(done){
+        it('On click of OK button the page should redirect to recordset page', function(done){
             chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
                 return btn.click();
             }).then (function (){
@@ -77,12 +75,9 @@ describe('Error related test cases,', function() {
               var newapplink = url.replace("record", "recordset"),
                   lastSlash = newapplink.lastIndexOf("/"),
                   recordsetUrl = newapplink.slice(0, lastSlash);
-                expect(currentUrl).toBe(recordsetUrl, "The redirection from record page to recordset/search in case of multiple records failed");
+                expect(currentUrl).toBe(recordsetUrl, "The redirection from record page to recordset in case of multiple records failed");
                 done();
-            }).catch( function(err) {
-                console.log(err);
-                expect('Something went wrong with this promise chain.').toBe('Please see error message.');
-            });
+            }).catch(catchError(done));
         });
     });
 
@@ -108,7 +103,7 @@ describe('Error related test cases,', function() {
 
         it('On click of OK button the page should redirect Home page', function(done){
             chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
-                return browser.executeScript("arguments[0].click();",btn);
+                return chaisePage.clickButton(btn);
             }).then (function (){
                 return browser.driver.getCurrentUrl();
             }).then (function(currentUrl) {
@@ -120,11 +115,7 @@ describe('Error related test cases,', function() {
                   }
                 expect(currentUrl).toBe(homePage, "The redirection from record page to Home page failed");
                 done();
-            }).catch( function(err) {
-                console.log(err);
-                done.fail(err);
-                expect('Something went wrong with this promise chain.').toBe('Please see error message.');
-            });
+            }).catch(catchError(done));
         });
     });
 
@@ -156,12 +147,10 @@ describe('Error related test cases,', function() {
                 expect(showDetails.getText()).toBe(testParams.hideErrors, "The Show/Hide message in modal pop is not correct");
                 expect(errorDetails.getText()).toContain("Error", "error missmatch.");
                 done();
-            }).catch(function (err) {
-                done.fail(err);
-            });
+            }).catch(catchError(done));
         });
 
-        it('On click of OK button the page should redirect to recordset/search page', function(done){
+        it('On click of OK button the page should redirect to recordset page', function(done){
             chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
                 return btn.click();
             }).then (function (){
@@ -170,11 +159,9 @@ describe('Error related test cases,', function() {
                 var newapplink = url.replace("record", "recordset"),
                     lastSlash = newapplink.lastIndexOf("/"),
                     recordsetUrl = newapplink.slice(0, lastSlash);
-                expect(currentUrl).toBe(recordsetUrl, "The redirection from record page to recordset/search in case of multiple records failed");
+                expect(currentUrl).toBe(recordsetUrl, "The redirection from record page to recordset in case of multiple records failed");
                 done();
-            }).catch( function(err) {
-                done.fail(err);
-            });
+            }).catch(catchError(done));
         });
     });
 
@@ -212,10 +199,7 @@ describe('Error related test cases,', function() {
                   }
                  expect(currentUrl).toBe(homePage, "The redirection from recordset page to Home page failed");
                  done();
-            }).catch( function(err) {
-                console.log(err);
-                expect('Something went wrong with this promise chain.').toBe('Please see error message.');
-            });
+            }).catch(catchError(done));
         });
     });
 
@@ -245,11 +229,7 @@ describe('Error related test cases,', function() {
                 // Added OR case to avoid discrepancy in error message when table is deleted
                 expect(errorText == testParams.conflictRecordEditErrorBooking || errorText == testParams.conflictRecordEditErrorAccommodationImg ).toBe(true, "409 Conflict could not be matched! Check conflict during deletion in RecordEdit.");
                 done();
-            }).catch(function(error) {
-                console.log(error);
-                done.fail(err);
-                expect('Something went wrong with this promise chain.').toBe('Please see error message.');
-            });
+            }).catch(catchError(done));
         });
 
         it('On click of Reload button the page should reload itself in Recordedit app', function(done){
@@ -260,11 +240,7 @@ describe('Error related test cases,', function() {
             }).then (function(currentUrl) {
                 expect(currentUrl).toBe(currentUrl, "Reload button could not refresh the recordedit page.");
                 done();
-            }).catch( function(err) {
-                console.log(err);
-                done.fail(err);
-                expect('Something went wrong with this promise chain.').toBe('Please see error message.');
-            });
+            }).catch(catchError(done));
         });
     }).pend("no button on Recordedit for delete");
 
@@ -289,25 +265,19 @@ describe('Error related test cases,', function() {
               return modalTitle.getText();
           }).then(function (text) {
               expect(text).toBe("Confirm Delete", "Deleteion confirmation pop-up could not be opened!");
-              return chaisePage.recordPage.getConfirmDeleteButton().click().then(function() {
-                errModalClass =  chaisePage.recordPage.getModalText();
-                return chaisePage.waitForElement(errModalClass).then(function() {
-                  return errModalClass.getText();
-                });
-              });
+              return chaisePage.recordPage.getConfirmDeleteButton().click();
+          }).then(function() {
+            errModalClass =  chaisePage.recordPage.getModalText();
+            return chaisePage.waitForElement(errModalClass);
+          }).then(function() {
+            return errModalClass.getText();
           }).then(function (errorText) {
               // Added OR case to avoid discrepancy in error message when table is deleted
               expect(errorText == testParams.deletionErrTextBooking || errorText == testParams.deletionErrTextAccommodationImg).toBe(true, "409 Conflict could not be matched! Check conflict during deletion.");
               done();
-          }).catch(function(error) {
-              console.log(error);
-              done.fail(err);
-              expect('Something went wrong with this promise chain.').toBe('Please see error message.');
-          });
+          }).catch(catchError(done));
       });
-
     });
-
 
     describe("History for errorneous Url", function(){
 
@@ -328,9 +298,7 @@ describe('Error related test cases,', function() {
             }).then (function(currentUrl) {
                 expect(currentUrl).toContain('id=269111', "The back button failed to go back to previous page.");
                 done();
-            }).catch(function(error) {
-                done.fail(error);
-            });
+            }).catch(catchError(done));
         });
 
     });
@@ -354,9 +322,7 @@ describe('Error related test cases,', function() {
                 chaisePage.waitForElement(closeModal);
                 expect(closeModal.isDisplayed()).toBeTruthy('Close modal option is not available for conflict/forbiddden errors');
                 done();
-            }).catch(function(error) {
-                done.fail(error);
-            });
+            }).catch(catchError(done));
         });
     });
 
@@ -378,10 +344,7 @@ describe('Error related test cases,', function() {
               chaisePage.waitForElement(closeModal);
               expect(closeModal.isDisplayed()).toBeTruthy('Close modal option is not available for conflict/forbiddden errors');
               done();
-          }).catch(function(error) {
-              console.log(error);
-              done.fail();
-          });
+          }).catch(catchError(done));
       });
 
     });
@@ -405,10 +368,7 @@ describe('Error related test cases,', function() {
           }).then(function (errorText) {
               expect(errorText).toBe(testParams.facetErrorstext.invalidPageCriteriaBody, "Error action text did not match");
               done();
-          }).catch(function(error) {
-              console.log(error);
-              done.fail();
-          });
+          }).catch(catchError(done));
       });
 
       it('On click of OK button the page should reload the page without paging condition but with invalid filter conditions', function(done){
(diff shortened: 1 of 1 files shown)


test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -223,36 +195,41 @@ describe('Error related test cases,', function() {
...
             }).then (function(currentUrl) {
-              var newapplink = url.replace("recordedit", "recordset"),
-                  lastSlash = newapplink.lastIndexOf("/"),
-                  recordsetUrl = newapplink.slice(0, lastSlash);
-                expect(currentUrl).toBe(recordsetUrl, "The redirection from recordedit page to recordset failed");
+              var homeAppUrl = browser.params.url,
+                  homePage =   homeAppUrl.slice(0, homeAppUrl.slice(0, homeAppUrl.lastIndexOf("/")).lastIndexOf("/") + 1);
+                  //Travis local URL has different structure
+                  if (process.env.TRAVIS) {
+                      homePage = currentUrl;
```
jrchudy: This is comparing `currentUrl` to `currentUrl`. Shouldn't this be `homePage = browser.params.url`?

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -223,36 +195,41 @@ describe('Error related test cases,', function() {
...
                 return btn.click();
             }).then (function (){
                 return browser.driver.getCurrentUrl();
             }).then (function(currentUrl) {
-              var newapplink = url.replace("recordedit", "recordset"),
-                  lastSlash = newapplink.lastIndexOf("/"),
-                  recordsetUrl = newapplink.slice(0, lastSlash);
-                expect(currentUrl).toBe(recordsetUrl, "The redirection from recordedit page to recordset failed");
+              var homeAppUrl = browser.params.url,
+                  homePage =   homeAppUrl.slice(0, homeAppUrl.slice(0, homeAppUrl.lastIndexOf("/")).lastIndexOf("/") + 1);
```
jrchudy: This is a little confusing as to what you are doing here. Please add a comment as to what the url you are working with looks like and what it's being changed to for the comparison.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -223,36 +195,41 @@ describe('Error related test cases,', function() {
...
+                  //Travis local URL has different structure
+                  if (process.env.TRAVIS) {
+                      homePage = currentUrl;
+                  }
+                 expect(currentUrl).toBe(homePage, "The redirection from recordset page to Home page failed");
+                 done();
             }).catch( function(err) {
```
jrchudy: Make sure to add `done.fail()` here and remove the `expect()`.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -337,107 +289,107 @@ describe('Error related test cases,', function() {
               return modalTitle.getText();
           }).then(function (text) {
               expect(text).toBe("Confirm Delete", "Deleteion confirmation pop-up could not be opened!");
-              chaisePage.recordPage.getConfirmDeleteButton().click();
-              errModalClass =  chaisePage.recordPage.getModalText();
-              chaisePage.waitForElement(errModalClass);
-              return errModalClass.getText();
+              return chaisePage.recordPage.getConfirmDeleteButton().click().then(function() {
```
jrchudy: It looks like you are intending to write this as a promise chain by returning the promise each time, but you are acting on the promise right away rather than part of the outer chain. This should be changed to something like:
```js
  return chaisePage.recordPage.getConfirmDeleteButton().click();
}).then(function() {
  errModalClass =  chaisePage.recordPage.getModalText();
  return chaisePage.waitForElement(errModalClass);
}).then(function() {
  return errModalClass.getText();
```

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -337,107 +289,107 @@ describe('Error related test cases,', function() {
...
+            deleteBtn  = chaisePage.recordPage.getDeleteRecordButton();
+        });
 
+        it('Error modal is dismissible in case of conflict/forbidden error', function(done){
+            chaisePage.waitForElement(deleteBtn).then(function(){
+                return deleteBtn.click();
+            }).then(function () {
+                return chaisePage.recordPage.getConfirmDeleteButton().click();
+            }).then (function() {
+                closeModal = chaisePage.recordEditPage.getModalCloseBtn();
```
jrchudy: Before we close the error modal that pops up after the confirm delete modal, we should check that the title and error message are as we expect in this case.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -337,107 +289,107 @@ describe('Error related test cases,', function() {
...
-          }).then(function (errorText) {
-              expect(errorText).toBe(testParams.facetErrorstext.invalidPageCriteriaBody, "Error action text did not match");
-              done();
-          }).catch(function(error) {
-              console.log(error);
-              done.fail();
-          });
-      });
+        it('After clicking back button initial page should appear', function(done){
+            chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
```
jrchudy: Before clicking on the ok button, we should add verification of the error title and message being present in the modal.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -337,107 +289,107 @@ describe('Error related test cases,', function() {
...
-          }).then(function (text) {
-              expect(text).toBe(testParams.facetErrorstext.invalidPageCriteriaTitle, "Invalid Page Criteria error pop-up could not be opened!");
-              return modalActionBody.getText();
-          }).then(function (errorText) {
-              expect(errorText).toBe(testParams.facetErrorstext.invalidPageCriteriaBody, "Error action text did not match");
+        it('Error modal is dismissible in case of conflict/forbidden error while deleting from ellipses', function(done){
+            deleteBtnEllipses.click().then(function(){;
+              return chaisePage.recordsetPage.getConfirmDeleteButton().click();
+            }).then (function() {
+              closeModal = chaisePage.recordEditPage.getModalCloseBtn();
```
jrchudy: Same comment as above, we should verify title and message here as well.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -493,118 +445,170 @@ describe('Error related test cases,', function() {
...
+                  done();
+              }).catch(function(error) {
+                  console.log(error);
+                  done.fail();
+              });
+          });
+
+          it('On click of OK button the page should reload the page without paging conditions', function(done){
+              chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
+                  browser.executeScript("arguments[0].click();",btn);
```
jrchudy: Use `chaisePage.clickButton(btn)` here.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -493,118 +445,170 @@ describe('Error related test cases,', function() {
...
+            });
 
-        beforeAll(function() {
-            var url = browser.params.url + "/record/#" + browser.params.catalogId + "/" + testParams.schemaName + ":" + testParams.table_name +  "/id=2002";
-            browser.get(url);
-            deleteBtn  = chaisePage.recordPage.getDeleteRecordButton();
-        });
+            it('An error modal window should appear with Record Not Found title', function(done){
+              chaisePage.waitForElement(element(by.css('.modal-dialog '))).then(function() {
+                chaisePage.recordPage.getErrorModalTitle().then(function(modalTitle) {
```
jrchudy: Let's use promise chaining here as well and add a catch handler with `done.fail()`:
```
  return chaisePage.recordPage.getErrorModalTitle()
}).then(function(modalTitle) {
  expect(modalTitle).toBe("Record Not Found", "The title of no record error pop is not correct");
  done();
}).catch(function(err) {
  console.log(err);
  done.fail();
});
```

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -161,57 +119,71 @@ describe('Error related test cases,', function() {
                       homePage = currentUrl;
                   }
                 expect(currentUrl).toBe(homePage, "The redirection from record page to Home page failed");
+                done();
             }).catch( function(err) {
                 console.log(err);
+                done.fail(err);
                 expect('Something went wrong with this promise chain.').toBe('Please see error message.');
```
jrchudy: This line can be removed because of `done.fail()`.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -161,57 +119,71 @@ describe('Error related test cases,', function() {
...
+            var showDetails = chaisePage.errorModal.getToggleDetailsLink();
+            var errorDetails = chaisePage.errorModal.getErrorDetails();
+            chaisePage.waitForElement(showDetails);
+            showDetails.click().then(function(){
+                chaisePage.waitForElement(errorDetails);
+                expect(showDetails.getText()).toBe(testParams.hideErrors, "The Show/Hide message in modal pop is not correct");
+                expect(errorDetails.getText()).toContain("Error", "error missmatch.");
+                done();
+            }).catch(function (err) {
+                done.fail(err);
```
jrchudy: We should log the error before calling `done.fail()`.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -161,57 +119,71 @@ describe('Error related test cases,', function() {
...
-                 expect(currentUrl).toBe(homePage, "The redirection from recordset page to Home page failed");
+                var newapplink = url.replace("record", "recordset"),
+                    lastSlash = newapplink.lastIndexOf("/"),
+                    recordsetUrl = newapplink.slice(0, lastSlash);
+                expect(currentUrl).toBe(recordsetUrl, "The redirection from record page to recordset/search in case of multiple records failed");
+                done();
             }).catch( function(err) {
-                console.log(err);
-                expect('Something went wrong with this promise chain.').toBe('Please see error message.');
+                done.fail(err);
```
jrchudy: We should still log the error before here.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -161,57 +119,71 @@ describe('Error related test cases,', function() {
...
-                  homePage =   homeAppUrl.slice(0, homeAppUrl.slice(0, homeAppUrl.lastIndexOf("/")).lastIndexOf("/") + 1);
-                  //Travis local URL has different structure
-                  if (process.env.TRAVIS) {
-                      homePage = currentUrl;
-                  }
-                 expect(currentUrl).toBe(homePage, "The redirection from recordset page to Home page failed");
+                var newapplink = url.replace("record", "recordset"),
+                    lastSlash = newapplink.lastIndexOf("/"),
+                    recordsetUrl = newapplink.slice(0, lastSlash);
+                expect(currentUrl).toBe(recordsetUrl, "The redirection from record page to recordset/search in case of multiple records failed");
```
jrchudy: `/search` can also be removed form this message too.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -267,65 +244,40 @@ describe('Error related test cases,', function() {
             }).then(function (errorText) {
                 // Added OR case to avoid discrepancy in error message when table is deleted
                 expect(errorText == testParams.conflictRecordEditErrorBooking || errorText == testParams.conflictRecordEditErrorAccommodationImg ).toBe(true, "409 Conflict could not be matched! Check conflict during deletion in RecordEdit.");
+                done();
             }).catch(function(error) {
                 console.log(error);
+                done.fail(err);
                 expect('Something went wrong with this promise chain.').toBe('Please see error message.');
```
jrchudy: This `expect()` can be removed.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -267,65 +244,40 @@ describe('Error related test cases,', function() {
...
                 return btn.click();
             }).then (function (){
                 return browser.driver.getCurrentUrl();
             }).then (function(currentUrl) {
                 expect(currentUrl).toBe(currentUrl, "Reload button could not refresh the recordedit page.");
+                done();
             }).catch( function(err) {
                 console.log(err);
+                done.fail(err);
                 expect('Something went wrong with this promise chain.').toBe('Please see error message.');
```
jrchudy: This `expect()` can be removed.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -337,107 +289,107 @@ describe('Error related test cases,', function() {
...
+                });
+              });
           }).then(function (errorText) {
               // Added OR case to avoid discrepancy in error message when table is deleted
               expect(errorText == testParams.deletionErrTextBooking || errorText == testParams.deletionErrTextAccommodationImg).toBe(true, "409 Conflict could not be matched! Check conflict during deletion.");
+              done();
           }).catch(function(error) {
               console.log(error);
+              done.fail(err);
               expect('Something went wrong with this promise chain.').toBe('Please see error message.');
```
jrchudy: This `expect()` can be removed.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -337,107 +289,107 @@ describe('Error related test cases,', function() {
...
+                return btn.click();
+            }).then (function (){
+                return browser.navigate().back();
+            }).then(function (){
+                return browser.driver.getCurrentUrl();
+            }).then (function(currentUrl) {
+                expect(currentUrl).toContain('id=269111', "The back button failed to go back to previous page.");
+                done();
+            }).catch(function(error) {
+                done.fail(error);
```
jrchudy: Make sure to log the error still.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -337,107 +289,107 @@ describe('Error related test cases,', function() {
...
+                return deleteBtn.click();
+            }).then(function () {
+                return chaisePage.recordPage.getConfirmDeleteButton().click();
+            }).then (function() {
+                closeModal = chaisePage.recordEditPage.getModalCloseBtn();
+                chaisePage.waitForElement(closeModal);
+                expect(closeModal.isDisplayed()).toBeTruthy('Close modal option is not available for conflict/forbiddden errors');
+                done();
+            }).catch(function(error) {
+                done.fail(error);
```
jrchudy: Make sure to log the error still.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -493,118 +445,170 @@ describe('Error related test cases,', function() {
...
+                }).then (function(currentUrl) {
+                    var newapplink = url.replace("recordedit", "recordset"),
+                        lastSlash = newapplink.lastIndexOf("/"),
+                        recordsetUrl = newapplink.slice(0, lastSlash);
+                    expect(currentUrl).toContain(recordsetUrl, "The redirection from record page to recordset/search in case of multiple records failed");
+                    done();
+                }).catch( function(err) {
+                    console.log(err);
+                    done.fail(err);
+                    expect('Something went wrong with this promise chain.').toBe('Please see error message.');
```
jrchudy: This `expect()` can be removed.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -493,118 +445,170 @@ describe('Error related test cases,', function() {
...
+                }).then (function(currentUrl) {
+                  var newapplink = url.replace("recordedit", "recordset"),
+                      lastSlash = newapplink.lastIndexOf("/"),
+                      recordsetUrl = newapplink.slice(0, lastSlash);
+                    expect(currentUrl).toBe(recordsetUrl, "The redirection from recordedit page to recordset failed");
+                    done();
+                }).catch( function(err) {
+                    console.log(err);
+                    done.fail(err);
+                    expect('Something went wrong with this promise chain.').toBe('Please see error message.');
```
jrchudy: This `expect()` can be removed.

jrchudy:
There are a good number of inconsistencies between some of the it cases that should be cleaned up. For our catch blocks, we always want to log the error first then call `done.fail()` if done is being used in the `it`. Having the `expect()` that forces a test failure is a bad idiom that we shouldn't have included in our tests in the first place. Using `done.fail()` is a much better practice.

A couple of the tests could benefit from verifying that the proper error dialog was being shown instead of just assuming that was the error thrown/displayed.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -75,65 +78,20 @@ describe('Error related test cases,', function() {
                   lastSlash = newapplink.lastIndexOf("/"),
                   recordsetUrl = newapplink.slice(0, lastSlash);
                 expect(currentUrl).toBe(recordsetUrl, "The redirection from record page to recordset/search in case of multiple records failed");
+                done();
             }).catch( function(err) {
```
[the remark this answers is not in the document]
RFSH: Since all these catch function will be the same you can just create a common function.

```javascript
var catchError =function (done) {
  return function (err) { 
   done.fail(err);
 }
}

it ("message", function (done) {
  promise.then(function () {

  }).catch(catchError(done))
});
```

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
RFSH: It seems like we're testing this type of error in the other `describe`s. This is supposed to just test the "back" button in that case. I feel like we could just move this `it` to that `describe` that is testing that type of error.

dhanashritidke:
I have completed the above mentioned changes in review comments.

PR #1586, opened 2018-06-29
Manual test for checking placement of tooltips

Divyata-Singh:
- Created a manual test for checking the placement of tooltips over column headers in the recordset app
- Added a command "make testmanually", in Makefile, to run manual tests 
- Updated  test/e2e/utils/protractor.configuration.js, to reuse its functions for specs in the manual test folder
- Updated documentation for the manual tests at https://github.com/informatics-isi-edu/chaise/wiki/Manual-UX-and-E2E-Testing#recordset---tooltip-on-column-headers

### DIFF
--- a/test/manual/specs/recordset.spec.js
+++ b/test/manual/specs/recordset.spec.js
@@ -1,0 +1,21 @@
+var chaisePage = require('./../../e2e/utils/chaise.page.js');
+
+describe('View recordset,', function () {
+    var waitTime = {
+        small: 15000,
+        medium: 30000,
+        large: 120000
+    }
+    beforeAll(function () {
+        browser.ignoreSynchronization = true;
+        browser.get(browser.params.url + "/recordset/#" + browser.params.catalogId + "/product-recordset:accommodation");
+
+        chaisePage.waitForElement(element(by.id("divRecordSet")));
+        chaisePage.recordsetPage.waitForAggregates();
+    });
+
+    // test to setup recordset table with long tooltip text and check the placement of tooltips manually
+    it("should have correct placement of longer tooltip texts on column headers", function () {
+        browser.sleep(waitTime.large);
+    });
+});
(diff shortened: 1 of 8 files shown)


test/manual/specs/recordset.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        browser.ignoreSynchronization = true;
+        browser.get(browser.params.url + "/recordset/#" + browser.params.catalogId + "/product-recordset:accommodation");
+
+        chaisePage.waitForElement(element(by.id("divRecordSet")));
+        chaisePage.recordsetPage.waitForAggregates();
+    });
+
+    // test to setup recordset table with long tooltip text and check the placement of tooltips manually
+    it("should have correct placement of longer tooltip texts on column headers", function () {
+        browser.sleep(waitTime.large);
```
jrchudy: I think it's better to use `browser.pause()` for this so we aren't constrained to a certain time limit.

Divyata-Singh: Browser.pause() is used if we have an action ( like click or sendkeys) in the test script following this statement (http://www.protractortest.org/#/api?view=ProtractorBrowser.prototype.pause). Since in this script, there is no further action, browser.pause() doesn't stop the execution and the test goes through without waiting.

jrchudy: oh, interesting. I hadn't thought of that. I still think pause is better, so maybe we could add `expect(true).toBeTruthy()` and it will behave how we expect it should.

jrchudy: My concern with sleep is that you can finish the verification rather quickly and have to wait for that timer to finish before the next `it()` get's called (when we add another `it()` to this spec for other manual tests.

Divyata-Singh:
Replaced browser.sleep() with browser.pause() in the test spec. Tested it with node v6.14.3

PR #1595, opened 2018-07-13
Chaise config

dhanashritidke:
Files in the following folders that use "chaiseConfig" have been modified:
1. common
2. record
3. recordedit
4. recordset
5. viewer


common/utils.js
```diff
@@ -1173,6 +1228,15 @@
         }
     }])
 
+    .factory("ConfigUtils", [function() {
+        function getConfigJSON() {
```
RFSH: I thought this function would take care of checking whether chaiseConfig is defined or not. It's easier to have that in here, instead of always returning the default here and then delegating the task of checking the existence of chaiseConfig to the caller.

Also I thought we want one singular place to attach the default variables for each individual attributes. Assume `chaiseConfig` is defined but `chaiseConfig.attribute` is not. Then this function should actually attach the default value of `attribute` to the `chaiseConfig`(Although this is only valid for attributes that their default value is a truthy value like a non-empty string, `true`, or a number). 

If you do these then we don't need to check for existence of chaiseConfig in different places.

dhanashritidke: Moved the check for 'chaiseConfig != "undefined"' to the function getConfigJSON(), and removed the redundant checks from all files.
A few places require the check (e.g. Cases where we do not have a default value) so I have kept them the same.

common/authen.js
```diff
@@ -3,7 +3,9 @@
 
     angular.module('chaise.authen', ['chaise.utils', 'chaise.storage'])
 
-    .factory('Session', ['messageMap', 'modalUtils', 'StorageService', '$cookies', '$http', '$interval', '$log', '$q', 'UriUtils', '$window', function (messageMap, modalUtils, StorageService, $cookies, $http, $interval, $log, $q, UriUtils, $window) {
+    .factory('Session', ['messageMap', 'modalUtils', 'StorageService', '$cookies', '$http', '$interval', '$log', '$q', 'UriUtils', 'ConfigUtils', '$window', function (messageMap, modalUtils, StorageService, $cookies, $http, $interval, $log, $q, UriUtils, ConfigUtils, $window) {
+
+        var chaiseConfig = ConfigUtils.getConfigJSON();
```
jrchudy: This can be removed and instead, `$rootScope` should be included as a dependency. Then each place that chaiseConfig is being used, it should be called like: `$rootScope.chaiseConfig` without any conditional checks. This should be done everywhere that this was defined like this.

common/errors.js
```diff
@@ -328,7 +330,7 @@ window.onerror = function() {
 
     var canClose = false;
 
-    if (chaiseConfig && chaiseConfig.allowErrorDismissal) {
+    if (typeof chaiseConfig != 'undefined' && chaiseConfig.allowErrorDismissal) {
```
jrchudy: Because this will run without angular running, this has to be done this way.

common/errors.js
```diff
@@ -339,7 +341,7 @@ window.onerror = function() {
         arguments[3]
     ].join(':');
 
-    var redirectLink = (chaiseConfig.dataBrowser ? chaiseConfig.dataBrowser : window.location.origin);
+    var redirectLink = (typeof chaiseConfig != 'undefined' && chaiseConfig.dataBrowser ? chaiseConfig.dataBrowser : window.location.origin);
```
jrchudy: Because this will run without angular running, this has to be done this way.

common/footer.js
```diff
@@ -8,14 +8,16 @@
...
+
+          var chaiseConfig = ConfigUtils.getConfigJSON();
 
             return {
                 restrict: 'E',
                 scope: {},
                 templateUrl: '../common/templates/footer.html',
                 link: function(scope, ele) {
-                    var footerText = chaiseConfig.footerMarkdown;
+                    var footerText = chaiseConfig.footerMarkdown ? chaiseConfig.footerMarkdown : "**Please check** [Privacy Policy](/privacy-policy/){target='_blank'}";
```
jrchudy: This should be `var footerText = $rootScope.chaiseConfig.footerMarkdown`. The property should be defined because of `setConfigJSON()`.

common/navbar.js
```diff
@@ -6,10 +6,17 @@
...
         'ui.bootstrap'
     ])
-    .directive('navbar', ['$window', '$rootScope', 'Session', 'modalUtils', function($window, $rootScope, Session, modalUtils) {
+    .directive('navbar', ['$window', '$rootScope', 'Session', 'modalUtils', 'ConfigUtils', function($window, $rootScope, Session, modalUtils, ConfigUtils) {
+
+        var chaiseConfig = ConfigUtils.getConfigJSON();
 
     // One-time transformation of chaiseConfig.navbarMenu to set the appropriate newTab setting at each node
-        var root = chaiseConfig.navbarMenu = chaiseConfig.navbarMenu || {};
+        var root;
```
jrchudy: There really wasn't a need to change this case. Setting the navBar menu to it's proper default means we can simplify this to:
```js
var root = $rootScope.chaiseConfig.navbarMenu
```

common/navbar.js
```diff
@@ -6,10 +6,17 @@
...
+        if(chaiseConfig.navbarMenu)
+          root = chaiseConfig.navbarMenu;
+        else {
+          root={};
+        }
         if (root) {
```
jrchudy: We can even remove this because `root` will always be defined, either as the chaiseConfig property set by the user or to the default value:
```json
{
  "newTab": true,
  "children": []
}
```

common/navbar.js
```diff
@@ -6,10 +6,17 @@
...
             // Set default newTab property at root node
             if (!root.hasOwnProperty('newTab')) {
```
jrchudy: This can also be removed because it should be set by default properly.

common/navbar.js
```diff
@@ -44,13 +51,36 @@
             scope: {},
             templateUrl: '../common/templates/navbar.html',
             link: function(scope) {
+              if(chaiseConfig.navbarBrand)
```
jrchudy: The rest of the checks below can be removed because of `setConfigJSON()`. Each property should be set like the following:
```js
scope.<property> = scope.$root.chaiseConfig.<property>
```

common/utils.js
```diff
@@ -3,6 +3,61 @@
 
     angular.module('chaise.utils', ['chaise.errors'])
 
+    .constant("defaultChaiseConfig", {
```
jrchudy: I realize I had suggested we should include the default value for things that don't necessarily have a default defined. But there are some cases where we don't want a default to bet set. 

This default object won't be populated in the case of search (you didn't make a change to call `setConfigJSON()` in search app (which is fine, we don't want to change search functionality). This means any property that is specific to search shouldn't be included in here.

I don't think we need to define the following properties:
- catalog (deprecated config option. Catalog is not actually used anywhere in the code. duplicate of defaultCatalog that replaces it. defaultCatalog should be undefined by default)
- feedbackUrl (I don't see it used in any of the apps, this might be deprecated at this point)
- helpurl (only used as a link in the tour code which is only in search. `''` is the same as `undefined` for that code)
- recordResource (this shouldn't even be configurable, deprecated)
- showBadgeCounts (only used in search app)
- tableThreshold (unused)
- showAllAttributes (only used in search app)
- dataBrowser (this is used to redirect for error messages. By default, if not set, we redirect to the root or `window.location.origin`. We don't want to force deployments to redirect to search instead of root... the default for this is wrong in configuration.md)
- hideSearchTextFacet (only used in search)
- maxColumns (only used in search)
- showUnfilteredResults (only used in search)
- sidebarPosition (only used in search)
- attributesSidebarHeading (the default is different from what is defined in the search code. I'm not sure what is the right solution, so we should not set this property. also it's only in search)
- searchPageSize (only used in search)
- customCSS (should not have a default value. if this is undefined by the user, it's assumed there is no custom CSS document)
- footerMarkdown (there is no default for this and if it's undefined, we don't show a footer. This change won't allow the footer to be NOT shown)
- maxRelatedTablesOpen (the place in the code that this is used, we want this to be "infinity" if undefined, not 15. just remove this property)
- plotViewEnabled (only used in search)
- recordUiGridEnabled, recordUiGridExportCSVEnabled, recordUiGridExportPDFEnabled (unused, not even defined in configuration.md)
- tour (only used in search. search e2e tests are no longer run)


Some of the default values are also set wrong:
- navbarBrand (should be `""` if not defined, not root)
- editRecord (default is false)
- deleteRecord (default is false)
- allowErrorDismissal (default is false)
- showFaceting (default is false)

@RFSH  can you verify these 2 lists?

jrchudy:
I found 2 other issues in the Pull Request when running the tests. 

First, the login app uses `scripts/app.js` which we didn't add `setConfigJson()` to. This was causing issues in the test suite when loading the login app (which is the only way we can login for the e2e tests). This uncovered a minor flaw in our logic which was when the `navbarMenu` property was not defined. We were relying on `navbarMenu.children` in the `link` function of the navbar directive, but if `navbarMenu` is undefined, this would error out. I updated the default config constant to reflect this.

Second, `chaiseConfig.editRecord` has the default listed as `false` in `configuration.md` which conflicts with the description of that property in the same document. I have changed this to say `N/A`. The issue is that `chaise` uses that property to control whether the `recordedit` app is available (for both create and edit). In `recordset.app.js` we explicitly check if it's **NOT** set to false before showing the create and edit links. If it is undefined, we _default_ to `true` which was lost based on the changes I had asked to be made before to include `editRecord: false` in the default config constant.

PR #1602, opened 2018-07-19
Warn on leaving recordedit page with unsaved changes

Divyata-Singh:
- Warn the user when he's about to leave the recordEdit page with unsaved changes.
- Checked on Chrome, Firefox, Safari, and Edge. All 4 of them display their own messages and have stopped the support for displaying custom messages. (https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload#Browser_compatibility)
- Stopped this event from triggering on form-submit event

### DIFF
--- a/recordedit/form.controller.js
+++ b/recordedit/form.controller.js
@@ -229,11 +229,10 @@
                 editOrCopy = true,
                 form = vm.formContainer,
                 model = vm.recordEditModel;
-
+            form.$setSubmitted()
             if (form.$invalid) {
                 vm.readyToSubmit = false;
                 AlertsService.addAlert('Sorry, the data could not be submitted because there are errors on the form. Please check all fields and try again.', 'error');
-                form.$setSubmitted();
                 return;
             }
 
@@ -800,6 +799,13 @@
             $timeout(function() {
                   resizeColumns();
             }, TIMER_INTERVAL, false);
+        });
+
+        $window.addEventListener("beforeunload", function(e) {
+            if(vm.formContainer.$submitted){
+                return undefined;
+            }
+            e.returnValue = "Do you want to leave this page? Changes you have made will not be saved.";
         });
     }]);
 })();


jrchudy:
We should have a test case for this that makes sure the notification shows up as we expect. Ideally it would be an e2e test (you should be able to seeing as `$window` is an angular component) but if not it can be added to the manual tests guide.

test/e2e/utils/recordedit-helpers.js
```diff
@@ -193,7 +193,26 @@ exports.testPresentationAndBasicValidation = function(tableParams, isEditMode) {
...
-
+    
+    if (tableParams.table_name === "accommodation") {
+        it("should give a warning when leaving the edit page with unsaved changes", function (done) {
+            var EC = protractor.ExpectedConditions;
+            var ratingInput = element(by.name("rating"));
+            browser.wait(EC.visibilityOf(ratingInput, browser.params.defaultTimeout));
+            ratingInput.sendKeys("5");
+            browser.refresh();
+            browser.switchTo().alert().then(function (alert) {
```
jrchudy: Only thing missing from this test case is an `expect()` clause. Not sure what will happen without the expectation.. 🤔 but to the best of my knowledge, the test will pass without actually running (unless done changes that fact).

I think we want to do something like:
```js
expect(alert.getText()).toBe("Do you want to leave this page? Changes you have made will not be saved.", "Alert message is not what we expected");
```

Divyata-Singh: So the alert behaves differently on different browsers. But on most browsers, it doesn't return the custom text that I have passed while calling the alert. I think alert().accept() is sufficient in this case because if the alert doesn't pop up at that point in the test case we'll get an error saying: "Failed: no such alert", i.e. we are trying to accept an alert that is not even present on the screen.

test/e2e/specs/all-features-confirmation/errors/errors.spec.js
```diff
@@ -529,8 +529,9 @@ describe('Error related test cases,', function() {
 
             it('On click of OK button the page should redirect to RecordSet', function(done){
                 chaisePage.recordPage.getErrorModalOkButton().then(function(btn){
-                    return btn.click();
-                }).then (function (){
+                    btn.click();
```
jrchudy: I think we want to act on this with a promise.
```
  return btn.click();
)}.then(function () {
  return browser.switchTo().alert().accept();
}).then(function () {
  ...
```

PR #1606, opened 2018-07-21
View header tooltip

Divyata-Singh:
- Added a tooltip to the view header in recordset table based on whether the header is View or Action
- Added the tooltip text to the recordset model

### DIFF
--- a/recordset/recordset.app.js
+++ b/recordset/recordset.app.js
@@ -64,7 +64,11 @@
         selectedRows: [],   // array of selected rows
         search: null,       // search term
         pageLimit: 25,      // number of rows per page
-        config: {}
+        config: {},
+        tooltip: {
+            actionCol: "Click on the action buttons to view, edit, or delete each record",
+            viewCol: "Click on the eye button to view the detailed page associated with each record"
+        }
     })
 
     // Register work to be performed after loading all modules
--- a/recordset/recordset.app.js
+++ b/recordset/recordset.app.js
@@ -64,11 +64,7 @@
         selectedRows: [],   // array of selected rows
         search: null,       // search term
         pageLimit: 25,      // number of rows per page
-        config: {},
-        tooltip: {
-            actionCol: "Click on the action buttons to view, edit, or delete each record",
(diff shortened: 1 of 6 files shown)


recordset/recordset.app.js
```diff
@@ -64,7 +64,11 @@
         selectedRows: [],   // array of selected rows
         search: null,       // search term
         pageLimit: 25,      // number of rows per page
-        config: {}
+        config: {},
+        tooltip: {
+            actionCol: "Click on the action buttons to view, edit, or delete each record",
+            viewCol: "Click on the eye button to view the detailed page associated with each record"
```
jrchudy: Both of these tooltips should be on the message map object. The "column" that these tooltips are being added to are part of the table directive which can appear in all 3 apps, so it doesn't make sense to add it here to the `recordsetModel`.

jrchudy:
Just need a test case for this now. We are testing the column comments (tooltips) in the recordset presentation spec, [here specifically](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js#L285-L303).

If you look at the [columns array of objects](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js#L12-L28) that is being used, we don't have a case for verifying the "Action" or "View" column comment.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -300,6 +301,13 @@ describe('View recordset,', function() {
                         });
                     });
                 });
+
+                //Check tooltip of Action column
+                var actionCol = element(by.cssContainingText('table > thead > tr > th > span', 'Actions'));
```
jrchudy: This will work for now, but if we change the structure of the HTML this test case will need to be updated. I think it's better if you add a class to the DOM like the rest of the `th` tags. We can call the class `actions-header` (no reason so complicate it and use actions in one case and view in another.

PR #1611, opened 2018-07-28
Rs table performance

jrchudy:
Changes to page data into the DOM rather than trying to show it all at once. The limit is set to 500 (cells) right now after some discussion with @hongsudt .


common/table.js
```diff
@@ -239,7 +239,34 @@
                     return vm.getDisabledTuples ? vm.getDisabledTuples(page, vm.pageLimit) : '';
                 }).then(function (rows) {
                     if (rows) vm.disabledRows = rows;
-                    vm.rowValues = DataUtils.getRowValuesFromPage(vm.page);
+                    var rowValues = DataUtils.getRowValuesFromPage(vm.page);
+                    var cellLimit = 500;
```
RFSH: This is not really important, but can you please define this in the `tableConstants` that is at the top of this file?

jrchudy: Oh right, I meant to move this to the constants object, especially because it shouldn't ever change.

PR #1617, opened 2018-08-09
Made all template urls absolute

Divyata-Singh:
- Added a function in chaise.utils to find the chaise path (UriUtils.chaiseDeploymentPath())
- Converted template URLs of all directives from relative to absolute
- Used the above function in logInHelper in authen.js and removed getDeploymentPathName()

### DIFF
--- a/common/alerts.js
+++ b/common/alerts.js
@@ -63,11 +63,11 @@
         };
     }])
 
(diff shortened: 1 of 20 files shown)


hongsudt:
can't we deal with the trailing / ourselves? If it is there, use it.. If it
is not there, add it..

PR #1618, opened 2018-08-11
Related entities tooltip

Divyata-Singh:
- Added tooltips to all action buttons on related and inline-related entities in record app
- In app.css added a class to remove the bottom border of these buttons with tooltips
- Added test cases to check these tooltips


record/record.app.js
```diff
@@ -125,6 +125,7 @@
                     }
 
                     model.column = col;
+                    model.baseTableName = $rootScope.reference.displayname.value;
```
jrchudy: For table name, use `reference.table.name`. It's true that the reference displayname is often this value, but this may not be true in the future.

hongsudt: I don't know the detail of the code, but want to state the objective which is to use the display value of the table that is shown on the page, so it makes sense to the user.

Divyata-Singh: I checked the different datatypes and found that for the base table the displayName and table.name are same but for related tables, the displayName and table.name might be different. So for the related entities, I have kept the displayName in the tooltip as that is what is displayed as the title as well.

RFSH:
I know that you just changed it to `table.name` but please change it to `table.displayname`. `name` attribute is supposed to be used internally. We should show the `displayname` to users.

As for why `reference.displayname` and `table.displayname` could be different, the heuristic for getting the displayname of related entities is as follows:

For pure and binary association,
- Use the `to_name` defined on the fk from association table to related table, if defined.
- Otherwise use the related table displayname.

For other related references,
- Use the `from_name` that is defined on the fk from main table to related table, if defined.
- Otherwise use the related table displayname.

Although in this case we're using `$rootScope.reference.displayname` which is the displayname of the main reference and not the related table. Based on the current heuristics this should always be the same as `table.displayname`. But I agree with @jrchudy. If we want to show the table displayname and don't care about the reference, it's better to use the `table.displayname`. We might later change the `reference.displayname` to return something else.

RFSH:
Based on the discussion that we have, we decided to always use the `uib-tooltip-html` and if each of the displaynames are not html (`isHTML == false`) we should just escape html character to make sure we're not encoding unwanted HTML tags. You can use [this](https://stackoverflow.com/a/6234804) function for escaping HTML.

Divyata-Singh:
I have changed the buttons to have uib-tooltip-html and if isHTML is false then I have sanitized the string.

PR #1620, opened 2018-08-15
e2e tests for export feature

jrchudy:
This PR adds some e2e tests in chaise for the export feature. After evaluating some more, it doesn't make sense to try to test the negative cases yet. 

Once the `500` error issue is addressed, writing some error cases would make sense, but the retry logic makes this pointless (for now).

For these tests I added a new variable to my global variables called `DOWNLOADS_PATH`, which when specified will allow the script to cleanup the files that it downloaded.


test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -271,6 +284,56 @@ describe('View recordset,', function() {
                 });
             });
 
+            it("should display the Export dropdown button with proper tooltip.", function() {
```
RFSH: Please add `done` and catch block to these test cases.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -271,6 +284,56 @@ describe('View recordset,', function() {
...
+                });
+            });
+
+            it("should have 'BDBag' as a download option and download the file.", function() {
+                chaisePage.recordsetPage.getExportDropdown().click().then(function () {
+                    var bagOption = chaisePage.recordsetPage.getExportOption("BDBag");
+                    expect(bagOption.getText()).toBe("BDBag");
+                    return bagOption.click();
+                }).then(function () {
+                    return chaisePage.waitForElement(element(by.css(".export-progress")));
```
RFSH: I think we should move this `element(by.css(".export-progress")` to the chaisePage.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.js
```diff
@@ -271,6 +284,56 @@ describe('View recordset,', function() {
...
+
+            it("should have 'CSV' as a download option and download the file.", function() {
+                chaisePage.recordsetPage.getExportDropdown().click().then(function () {
+                    var csvOption = chaisePage.recordsetPage.getExportOption("CSV");
+                    expect(csvOption.getText()).toBe("CSV");
+                    return csvOption.click();
+                }).then(function () {
+                    return browser.wait(function() {
+                        return fs.existsSync(process.env.DOWNLOADS_PATH + '/Accommodations.csv');
+                    }, 10000);
```
RFSH: Use the default wait time that we have defined. Is there any reason behind using `10000`?

jrchudy:
@RFSH I cleaned up the test cases, added `done` to each of them, and verified that they pass much more consistently. 

Testing file download in travis isn't going to work because we don't have access to reading the file in the file system (even to verify it exists).

PR #1626, opened 2018-08-23
Facet order

RFSH:
The [recent change in ermrsetjs](https://github.com/informatics-isi-edu/ermrestjs/pull/694) will break the existing test cases because the `jsonb` column types are not sortable in ermrestjs. I had to defined the `order` attribute for that facet.

Also I had to change the API call for getting the scalar values to pass the two new APIs that are added to faceting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,6 @@ before_install:
   - cd ..
   - git clone https://github.com/informatics-isi-edu/ermrestjs.git
   - cd ermrestjs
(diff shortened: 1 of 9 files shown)


jrchudy:
@RFSH This PR can also be merged. Make sure to change `travis.yml` in this one after merging the [ermrestJS PR](https://github.com/informatics-isi-edu/ermrestjs/pull/694).

PR #1643, opened 2018-09-12
Download bibtex

jrchudy:
This PR adds support to download the citation (if present) in the `BibTex` format.

Fixed the citation annotation to be more inline with what is defined in the [citation issue](https://github.com/informatics-isi-edu/ermrestjs/issues/678) and shown [here on RBK](https://dev.rebuildingakidney.org/about/usage.html).

Other changes include:
 - collating config setup for record app
 - refactoring some of the share dialog tests to be simpler
 - prepending `chaise-` so the share dialog can more accurately be identified

### DIFF
--- a/record/record.controller.js
+++ b/record/record.controller.js
@@ -90,6 +90,7 @@
         };
 
         vm.sharePopup = function() {
+            var tuple = $rootScope.tuple;
             modalUtils.showModal({
                 templateUrl: UriUtils.chaiseDeploymentPath() + "common/templates/shareCitation.modal.html",
                 controller: "ShareCitationController",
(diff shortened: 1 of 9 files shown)


record/record.controller.js
```diff
@@ -90,15 +90,17 @@
...
-                windowClass: "share-citation",
+                windowClass: "chaise-share-citation",
                 controllerAs: "ctrl",
                 resolve: {
                     params: {
-                        citation: $rootScope.tuple.citation,
-                        permalink: $window.location.href
+                        citation: tuple.citation,
+                        permalink: $window.location.href,
+                        displayname: $rootScope.reference.table.name+'_'+tuple.data.RID
```
RFSH: Shouldn't we use `tuple.uniqueId` instead of  `tuple.data.RID` ? They mean the same thing for the tables that have `RID`. Besides I feel like we should avoid making assumption about name of columns in chaise.

RFSH: Also why not use the `tuple.displayname` if we want uniqueness? Unless we actually want to show the RID to the users, which then the filename format makes sense.

jrchudy: We want an unformatted unique identifier for the row. We don't care what the identifier is as long as it's unique. So using `tuple.uniqueId` will be sufficient. We are using `table.name` because it doesn't include any special formatting from markdown or HTML.

Issue #1640, opened 2018-09-07
Faceting Performance Issue


RFSH:
We're seeing very bad performance in one of the tables in `rbk-dev`. You can see the behavior by doing the following (Please don't try this multiple time. It will slow the whole server down):
- Go to the following link: https://dev.rebuildingakidney.org/chaise/recordset/#2/Vocabulary:Anatomy_terms
- select "All records with value" for "Specimen Expression"
- Open "Anatomical Source of Specimen" and it will hang there. It will eventually throw the timeout error.

RFSH:
These are the set of improvements that we should apply:

- Use `array_d` instead of `array` for foreign-key and main entity projection list.
- Don't ask for foreign-key data when we're only getting a list of row-names (Use `entity` api instead of `attributegroup`). 
- Switch the foreign-key paths to inner join instead of left outer join.

The first two improvements won't have any visible effects in the client. They will improve the performance on the main page, but still the "show more" page for facet will throw the timeout. So we need to do the third improvement too. 

Changing the joins to inner will remove `null` from the list of possible values for the facet. To fix this issue, for now, we decided to add the `No Value` option to all the facets. It will be the second option that will always be visible for the facets. In the "show more" modal, it will be a checkbox besides the "All Records With Value". Although, when a facet has "null" filter, we must use left outer join for that path for getting the result on the main page.

Other possible improvements that we talked about for later:
- Ask server if "null' is possible and show/hide the "No Value" option.
- Instead of always showing the "No Value" on top, we can turn the request in ermrestjs into two different requests. One asking for values that exist (inner), and another for the null. The problem with this approach is the complexity of merging the "null"  row with the other rows. In entity-mode it should be simple. It's either at the bottom of the list or first. But in scalar, since we might not be sorting based on the value of that column, it could be anywhere in the list of results (This requires more investigation).

PR #1645, opened 2018-09-13
Null facet

RFSH:
1. Hide them based on the two newly added ermrestjs APIs.
2. Show null on top of the other options.
3. Show null option in the modal beside the not-null

I just need to add one extra test case for testing the "null" option in the modal.

#1640

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,6 @@ before_install:
   - cd ..
   - git clone https://github.com/informatics-isi-edu/ermrestjs.git
   - cd ermrestjs
-  - git checkout performance-fix
   - sudo npm install
   - bower install
   - sudo make build
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,6 +82,7 @@ before_install:
   - cd ..
   - git clone https://github.com/informatics-isi-edu/ermrestjs.git
   - cd ermrestjs
+  - git checkout performance-fix
   - sudo npm install
   - bower install
   - sudo make build
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -4,11 +4,12 @@
     angular.module('chaise.faceting', ['plotly', 'chaise.utils'])
 
         .factory('facetingUtils', ['defaultDisplayname', function (defaultDisplayname) {
-            function getNullFilter(selected) {
+            function getNullFilter(selected, disabled) {
                 return {
                     selected: (typeof selected == 'boolean') ? selected: false,
+                    disabled: (typeof disabled == 'boolean') ? disabled: false,
                     uniqueId: null,
-                    displayname: {"value": defaultDisplayname.null, "isHTML": true}
+                    displayname: {"value": null, "isHTML": false}
                 };
             }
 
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -82,6 +82,7 @@ before_install:
   - cd ..
   - git clone https://github.com/informatics-isi-edu/ermrestjs.git
   - cd ermrestjs
+  - git checkout performance-fix
```
jrchudy: Don't forget to change this back.

common/faceting.js
```diff
@@ -851,12 +855,16 @@
                     defer.resolve();
                 }
                 else {
+                    if (scope.facetColumn.hasNullFilter) {
+                        scope.facetModel.appliedFilters.push(facetingUtils.getNullFilter(true));
```
jrchudy: Does this mean it should be selected by default when the facetColumn has it? I thought if it was defined for the facetColumn that just means it should be shown..

common/faceting.js
```diff
@@ -924,8 +945,15 @@
 
 
                 var appliedLen = scope.facetModel.appliedFilters.length;
+                var pageSize = tableConstants.PAGE_SIZE;
+                if (!scope.facetColumn.hideNullChoice) {
```
jrchudy: Could you add a comment as to why the page size should be increased because of this filter. I was thinking the presence of this filter shouldn't change the page size used in the read function..

common/modal.js
```diff
@@ -215,12 +216,11 @@
...
         function getMultiSelectionResult() {
-            var res = vm.tableModel.selectedRows;
-            if (!Array.isArray(res)) res = [];
             if (vm.tableModel.matchNotNull) {
-                res = {matchNotNull: true};
+                return {matchNotNull: true};
             }
-            return res;
+            var res = vm.tableModel.selectedRows;
+            return Array.isArray(res) ? res : [];
```
jrchudy: Returning an array here, but above you return an object..

common/templates/recordsetSelectFaceting.html
```diff
@@ -17,8 +17,8 @@
                         <a class="remove-link" ng-click="removePill(row.uniqueId, $event)"><i class="glyphicon glyphicon-remove"></i></a>
                         <span ng-if="row.displayname.value && row.displayname.isHTML" ng-bind-html="row.displayname.value"></span>
                         <span ng-if="row.displayname.value && !row.displayname.isHTML" ng-bind="row.displayname.value"></span>
-                        <span ng-if="row.displayname.value == '' && row.uniqueId != null" ng-bind-html="defaultDisplayname.empty"></span>
-                        <span ng-if="row.displayname.value == null || row.uniqueId === null" ng-bind-html="defaultDisplayname.null"></span>
+                        <span ng-if="row.displayname.value == ''" ng-bind-html="defaultDisplayname.empty"></span>
```
jrchudy: why don't we need to check uniqueId anymore?

common/faceting.js
RFSH: This should be refactored.

common/modal.js
RFSH: This can be refactored.

PR #1647, opened 2018-09-18
added error cases for 400 query timeout

jrchudy:
_body pending_

### DIFF
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -48,6 +48,8 @@
                         ctrl.facetingCount++;
 
                         $scope.vm.facetModels[index] = {
+                            facetName: facetColumn.displayname.unformatted,
+                            hasError: false,
                             initialized: false,
                             isOpen: false,
                             isLoading: false,
--- a/common/table.js
+++ b/common/table.js
@@ -86,8 +86,8 @@
      * modified. ellipses will fire this event and recordset directive will use it.
      */
     .factory('recordTableUtils',
-            ['AlertsService', '$document', 'messageMap', 'modalBox', 'DataUtils', '$timeout','Session', '$q', 'tableConstants', '$rootScope', '$log', '$window', '$cookies', 'defaultDisplayname', 'MathUtils', 'UriUtils', 'logActions',
-            function(AlertsService, $document, messageMap, modalBox, DataUtils, $timeout, Session, $q, tableConstants, $rootScope, $log, $window, $cookies, defaultDisplayname, MathUtils, UriUtils, logActions) {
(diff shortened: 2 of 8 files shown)


jrchudy:
@RFSH  Some changes that I made to accommodate the simplified facet mode:
 - `retryQuery` and `resetFacet` functions for options in facet.
    - should `hasError` be set to `false` here? it is only used for the QTE case and avoids throwing the error
    - I'm thinking it might need to be done "only" in success callbacks (regardless of it's previous value)
 - updateCount bug
    - if [this](https://github.com/informatics-isi-edu/chaise/blob/master/common/table.js#L377-L385) catch clause is reached and it make it to the "fail silently" comment, no value is returned to the calling function and therefore `_afterUpdateCount` is not called. This causes the flow control `occupiedSlots` count to be incorrect and will eventually cause all requests to "hang"
 - Should the changes apply to range facets as well?

common/faceting.js
```diff
@@ -72,6 +73,10 @@
                         });
                     };
 
+                    /**
+                     * @param {int} index index of facetColumn
+                     * @param {boolean} noConstraints set property to run the query without constraints
```
RFSH: It seems like you've removed this parameter

common/table.js
```diff
@@ -380,8 +402,15 @@
                         return defer.promise;
                     }
 
+                    if (err instanceof ERMrest.QueryTimeoutError) {
+                        vm.countError = true;
```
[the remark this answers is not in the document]
jrchudy: Right, this should be done in the success callbacks. I'm going to change the `hasError` properties so that the errors are `facetError`, `tableError`, `countError`, `tableError`.

jrchudy:
400 Query Timeout in facet:
   - Simplify button tooltip -> "Retry the query that timed out without any other facet constraints applied."
image

After clicking "Simplify" from above picture:
  - Reset link tooltip -> "Reset"
image

jrchudy:
@RFSH  I updated this issue with the changes to the text that @hongsudt suggested. I also added a class to each ellipses cell to change the background color of it if the current aggregate column has an error.

common/templates/ellipses.html
```diff
@@ -21,7 +21,7 @@
         <label />
     </div>
 </td>
-<td ng-if="config.showNull || tuple.uniqueId" ng-repeat="val in rowValues track by $index">
+<td ng-if="config.showNull || tuple.uniqueId" ng-repeat="val in rowValues track by $index" ng-class="{'disabled-cell': columnModels[$index].columnError}">
```
jrchudy: This is why the ellipses directive (our row directive) has all the columnModels attached to it.

common/faceting.js
```diff
@@ -59,6 +59,7 @@
                         ctrl.facetingCount++;
 
                         $scope.vm.facetModels[index] = {
+                            facetError: false,
```
RFSH: Just for the sake of bookkeeping, can you please add the `noConstraints:false` to here too. So we would know that the facetModel can have such attribute.

PR #1652, opened 2018-09-20
Refactor null changes

RFSH:
We rushed the #1645 changes. This PR will add some comments to the code and also refactors some part of it.


jrchudy:
Merged master into the branch and was able to run the tests locally without any issues.

PR #1657, opened 2018-09-24
fix check presence bug

RFSH:
The list of filters that we're showing to the users is different from the actual list of filters that are used for faceting. It's chaise responsibility to keep them in sync. For check presence, I was always showing a `not-null` filter (breadcrumb) which this PR fixes it.


RFSH:
I'm closing this PR since we are going to do this change in the other PR.

PR #1660, opened 2018-09-27
Default export

RFSH:
It's a small change based on [ermrestjs#718](https://github.com/informatics-isi-edu/ermrestjs/pull/718) changes. I've added the `exportTemplates` to reference object, which chaise should call.

I am also printing the default export template so data-modelers can just grab and change it.

I'm not sure whether I need to update the test cases or not. If we're testing the scenario of not having any export annotation, they I should.


RFSH:
@jrchudy I updated this PR to reflect the changes that I made on the ermrestjs PR. Most of the changes are just updating the attribute names.

RFSH:
@carlkesselman Yes, I will explain on the issue what needs to be changed and mention you.

.travis.yml
```diff
@@ -82,6 +82,7 @@ before_install:
   - cd ..
   - git clone https://github.com/informatics-isi-edu/ermrestjs.git
   - cd ermrestjs
+  - git checkout default-export
```
jrchudy: Don't forget to remove this.

Issue #715, opened 2016-11-01
record2: successful deletion should direct to the search page


hongsudt:
The current successful deletion is very abrupt. When an experiment is deleted, it directs me to the landing page instead of a search page.

jrchudy:
This should be redirecting you to `chaiseConfig.dataBrowser` if that is set. If it isn't it goes to the landing page.

jrchudy:
This has been fixed.

PR #1661, opened 2018-09-28
Catalog resolver

jrchudy:
This PR adds the functionality mentioned in issue [#715](https://github.com/informatics-isi-edu/ermrestjs/issues/715) in `ermrestJS`. Added tests for 3 of the 4 defined in the issue.

I left a question in the code because it's not clear if we should _always_ use RID or if we should use the `uniqueId` property.

The case not tested is when the `id` provided in the `chaiseConfig` matches the id of the current record page. We don't have a way of injecting the current catalog id (an incremental 5 digit integer) into the `chaiseConfig` document. 

I could write a test that only runs locally that points to `dev.isrd.isi.edu/chaise/#1/isa:dataset` but this wouldn't have the appropriate RID in the url (that catalog on dev.isrd has not been updated with system columns).

Edit:
About that 4th test case, testing if the `chaiseConfig` property is the same as the current page's `catalogId`, this could actually be tested in travis without concern because travis will always create 1 catalog for each parallel configuration with catalog numbers 1-4.


hongsudt:
The resolver is designed to primarily resolve ermrest RID. So RID should always be used in the shared link.

jrchudy:
@RFSH  this PR is ready for review.

record/record.controller.js
```diff
@@ -94,7 +96,21 @@
...
             var tuple = $rootScope.tuple;
+
+            if (chaiseConfig.resolverImplicitCatalog !== undefined && chaiseConfig.resolverImplicitCatalog !== null) {
+                var catalogFilter = "";
+                // catalogId in chaiseConfig property does not match current catalogId
+                if (chaiseConfig.resolverImplicitCatalog != $rootScope.reference.location.catalog) catalogFilter = chaiseConfig.resolverImplicitCatalog + "/";
+                // catalogId in chaiseConfig property is 'false'
+                if (chaiseConfig.resolverImplicitCatalog == false) catalogFilter = $rootScope.reference.location.catalog + "/";
+                // TODO should this be RID or uniqueId?
+                permalink = $window.location.origin + "/id/" + catalogFilter + tuple.data.RID;
```
RFSH: I know RID will always be available with the current version of ermrest, but we should actually check for it first and if it doesn't exist, we should not offer resolver.

PR #1664, opened 2018-10-10
Batch copy

jrchudy:
This PR includes the functionality for batch copy in the `recordedit` app. 

Users will have the ability to click `Select All` in the row header (where the column name is). This expands a new table row below the current one for that column with the appropriate input type based on the type of the column that should match the input from the current `recordedit` form. After setting a value, the user can click apply which will set the value in all the rows to be created/edited. 

If no value was selected, the "apply" button will show up as "clear all" instead which will remove the value from all `recordedit` form inputs for that column. There is also a cancel button next to the "apply all" button that collapses the all input row and "resets" that input.

When you toggle any of the select all row to be visible, that row's inputs are changed to a "disabled" input which has it's own "type" in the `recordedit` app. This change required some other minor refactoring changes to be made to the `RE` app.


jrchudy:
@RFSH this is ready for remove. I can go over the workflow with you if you want.

recordedit/form.controller.js
```diff
@@ -392,6 +384,73 @@
...
+                controller: "RecordeditInputPopupController",
+                controllerAs: "ctrl",
+                resolve: {
+                    params: params
+                },
+                templateUrl: UriUtils.chaiseDeploymentPath() + "common/templates/inputPopup.modal.html",
+                windowClass: "modal-select-all"
+            }, function (model) {
+                if (params.displayType === "popup-select") {
+                    // set data in view model (model.rows) and submission model (model.submissionRows)
```
RFSH: Add a little explanation why you're doing this. You can briefly explain the issue that you faced when you were just setting the value without actually copying it.

common/inputs.js
```diff
@@ -172,5 +172,126 @@
...
+                        } else if (scope.column.isAsset) {
+                            return "No file Selected";
+                        }
+                    } catch (e) {
+                        $log.info(e);
+                    }
+                }
+
+                // Assigns the current date or timestamp to inputValue
+                scope.applyCurrentDatetime = function() {
```
RFSH: Are these functions that you have defined in this directive, the same functions that are currently in the recordedit page? Since at some point we might refactor recordedit to use this, would it help to add some TODOs to these functions in `form.controller.js`? 
If we want to change any of these functions, then we have to make sure that we're going to change the two identical places. So it might be easier to have a reminder that these functions are the same.

jrchudy:
@RFSH  I updated the PR body above with the latest information for this PR. This is ready for review again.

PR #1683, opened 2018-11-16
Recordedit remove form (edit mode)

jrchudy:
This PR adds the remove form functionality to edit mode as well in the `recordedit` app.

A new test case was added to test and verify this feature. It's in a new spec file but will run like it's part of the multi-edit spec.


recordedit/form.controller.js
```diff
@@ -425,15 +425,36 @@
...
             vm.recordEditModel.submissionRows.splice(index, 1);
+            vm.recordEditModel.foreignKeyData.splice(index, 1);
+            if (vm.editMode) $rootScope.tuples.splice(index, 1);
             $timeout(function() {
                 onResize();
                 $rootScope.showSpinner = false;
             }, 10);
         }
 
+        var CONFIRM_DELETE =  (chaiseConfig.confirmDelete === undefined || chaiseConfig.confirmDelete) ? true : false;
```
RFSH: This is not really important, but since we refactored the `chaiseConfig`, you don't need to check for it anymore and you should be able to use the `chaiseConfig.confirmDelete` directly.

PR #1685, opened 2018-11-26
Reload button for No Connection Error

jrchudy:
Added reload button for when a No Connection Error occurs during the loading of the application.

### DIFF
--- a/common/modal.js
+++ b/common/modal.js
@@ -42,11 +42,12 @@
             $uibModalInstance.dismiss('cancel');
         }
     }])
-    .controller('ErrorModalController', ['Errors', 'messageMap', 'params', 'Session', '$uibModalInstance', '$window', function ErrorModalController(Errors, messageMap, params, Session, $uibModalInstance, $window) {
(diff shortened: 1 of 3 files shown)


jrchudy:
@RFSH  the changes to show the reload button for all retriable error codes have been added to this PR.

common/modal.js
```diff
@@ -60,6 +54,28 @@
...
+            }
+        }
+
+        var vm = this;
+        vm.params = params;
+        vm.displayDetails = false;
+        vm.linkText = messageMap.showErrDetails;
+        vm.showReloadBtn = false;
+        if (ERMrest && isRetryError(exception)) {
+            vm.showReloadBtn = !$rootScope.displayReady;
```
RFSH: I think this means that we're only showing reload button on the requests that are being created before the page completely loads? Can you please add a comment as to why it's looking for `displayReady`?

RFSH: I might be wrong, but it seems like recordset is not setting this variable.

jrchudy: That's exactly the point, we are only showing this for the 4 types of retriable errors while the page is loading. Last week we discussed that it doesn't make sense to "retry" other requests that may fail after the data has loaded.

I made a change to table.js to set `$rootScope.displayReady` after the main entity request succeeds or fails.

RFSH: My bad, I didn't see the `table.js` changes. Makes sense. I think we need to comment the reasoning in the code. It's not really obvious what the code is trying to do.

PR #1686, opened 2018-11-27
Tooltips for null and not-null facet options

jrchudy:
Changed the tooltips for null, empty, and not null facet options to include more than just the same information in the tooltip message.


common/faceting.js
```diff
@@ -28,7 +28,11 @@
                 return {
                     selected: (typeof selected == 'boolean') ? selected: false,
                     isNotNull: true,
-                    displayname: {"value": defaultDisplayname.notNull, "isHTML": true}
+                    displayname: {"value": defaultDisplayname.notNull, "isHTML": true},
+                    tooltip: {
+                        value: messageMap.tooltip.notNull
```
RFSH: @jrchudy Is there a reason that you have defined the tooltip as an object here? It seems like you can just say `"tooltip": message.tooltip.notNull`.

RFSH: But in the template you're always using the `.value` without checking for the `isHTML`. Also if you want to be following the same pattern, in here you need to add the `"isHTML": true`. But I don't think it's worth doing so (since as I mentioned we're not checking for it in the template).

RFSH:
@jrchudy I added tooltip for `No Value` and `Empty` too. It should be ready to merge. 

@hongsudt These are the tooltips that we're showing in different scenarios. Please let me know if we need to change any of them.

- Null value: Search for any record with no value assigned
- Not null: Search for any record that has a value
- Empty string value: Search for any record with empty string value assigned

RFSH:
I updated the empty string tooltip to "Search for any record with the empty string value". The PR is ready to be reviewed and merged.

Issue #1695, opened 2018-12-14
Check Page.content In related entities


RFSH:
If it results empty (template is wrong), it should be treated as if the table-display is missing (fall back to tabular display).

We might want to do the same for inline tables. Should we?

hongsudt:
It is not just because the template is wrong, but also because the template is not applicable e.g. I can only apply treeview to mouse data only instead of all data.

RFSH:
This has been merged with master. The new behavior regarding the nested tables are as follows:

inline table:

- If page.content is null (no matching rows), show None and a button to switch to tabular mode.
- If page.content is empty string (empty string, due to invalid or empty row_markdown_pattern result), show the tabular mode. Don't provide the switch button.
Otherwise show the page.content in markdown-container. Provide the switch button.

Related table:

- If page.content is valid and there are matching rows, show page.content in a markdown-container with the switch button.
- Otherwise show the tabular mode.

What has been changed is checking for empty result of page.content and also checking for null values in the related section. Previously, as long as a related table has the `row_markdow_pattern`, we were showing the markdown display even if it didn't have any matching rows.

Assigning to @hongsudt to verify the behavior.

hongsudt:
This is now working as intended.

PR #1699, opened 2019-01-07
switch related table mode when page.contet is empty

RFSH:
This PR changes the related tables behavior based on what is described in #1695. 

### Expected Behavior
To summarize this is how we want to present the related/inline tables:

inline table:
 - If `page.content` is null (no matching rows), show `None` and a button to switch to tabular mode.
 - If `page.content` is invalid (empty string, due to invalid or empty `row_markdown_pattern` result), show the tabular mode. Don't provide the switch button.
 - Otherwise show the `page.content` in `markdown-container`. Provide the switch button.

Related table:
 - If `page.content` is valid and there are matching rows, show `page.content` in a `markdown-container` with the switch button.
- Otherwise show the tabular mode.

### Code Changes

The problem with this requirement is the fact that showing or not showing the page.content is solely based on `reference.display.type` and not the data. I had to change this to actually look for the data too. Another issue here was the fact that I had to change this for both related and inline tables. We're showing page.content in `record.html` (for inline), and also `record/index.html.in` (for related table).

I could think of doing this in two different ways that I explained in the following. The code changes currently are based on the second approach.

1. At first I was doing the same thing that we're doing with the `showRelatedTable` function in `record.controller.js`. Since the checks are very minimal I don't think having a function that would run multiple times would be an issue here. The problem that I had with this approach was the fact that I had to write the same function in `record.js` too. So if we want to change this behavior later, we have to make sure to update these two functions and the logic is not in just one place.

2. I ended up with adding a new parameter to the `tableModel` object in `table.js` which will signal the UI that you can use the value of `page.content`. This will avoid the computation of it in different places. We can just use this value in the two different places. The only problem that I can see with this approach is adding a parameter to the `tableModel` object and polluting the object (which is not that much of an issue in my opinion).

### Test Changes

1. It seems like we had accidentally removed inline table test cases from record page, which I added back.
2. Added a new inline table to the product schema. Since the indentation in that file was messy I formatted the whole file, but the actual changes are just adding a new table.
2. Added a new related table to related-table test spec which is identical to the table that I added for inline test case.
3. Since the chaise-config that the related-table spec is using, has a `maxRelatedTablesOpen`, I had to increase its value. I also had to add a new table to the product-max-RT schema which is testing this chaise config.


RFSH:
I decided to change the code and go with the first approach. These are the main reasons behind it:


1. I realized adding an attribute to tableModel will impact other paths of the code. `table.js` is used heavily in recordset and the structure of reference object could be different from the reference in record page (for example in recordset directive we could have AttributeGroupReference). 
2. We are currently showing `page.content` only in record app. So it's better to do the checking in the exact place that it's going to be used.
3. We are going on change how to get the data for the markdown display of related/inline tables in near future. When we do that, this check must be in the record page where we have the logic for getting the data.

PR #1700, opened 2019-01-10
modify test cases and code for stabilizing tests

RFSH:
This PR is an effort to stabilize the test cases. To make sure the test cases that "randomly" fail are going to pass. Most of the issues are usually not waiting for the correct elements. Or assuming an order of loading the elements without actually waiting for them.

The test cases are passing locally but we should not merge this PR until all the test cases pass in travis as well. This will be a continuous effort,  I will change the pipeline to review when all the test cases pass.


jrchudy:
oops, I didn't mean to review this PR. Meant to mark that on the changes to Ermrest Data Utils.

jrchudy:
I updated this wiki page with my findings from looking into the e2e test failures in travis/sauce:
https://github.com/informatics-isi-edu/chaise/wiki/KEDB-(Test-Cases-Fail-Count)

I collected a set of builds from the last month that failed or errored out after 30+ minutes then ran my script against it. I did the same thing for a subset of that set but only for “failed” builds (ones that finished in under 50 minutes but with test failures).

PR #1702, opened 2019-01-18
Login app

Divyata-Singh:
- login.html - is the rbk home page with the login app
- login.app.js - is the angular app for login
- login-app.md - It has information on how to use the login app


RFSH:
I don't think we should add the `login.html` to chaise. It's very specific to the deployments. It most probably should be added as part of building process in any deployments. If your intention is to have an example, you can just have a very minimal example of how it should look like in your `login-app.md`. Something like the following: 

```html
<!DOCTYPE html>
<html>
  <head>
    <!-- other deployment specific dependencies -->
    <!-- Scripts to be included for login app -->
    <script src="./chaise/chaise-config.js"></script>
    <script src="./chaise/scripts/vendor/angular.js"></script>
    <script src="./chaise/loginApp/login.app.js"></script>
  </head>
  <body>
    <div class="wrap">
      <header id="header">
        <nav id="navbar" class="navbar navbar-default navbar-fixed-top">
          <div class="container-fluid">
            <div class="row">
              <div class="col-sm-4 col-md-4">
                <div class="navbar-header">
                  <!-- navbar header (title, brand) -->
                </div>
              </div>
              <div class="col-sm-6 col-md-6">
                <div class="navbar-collapse collapse" id="navbar-collapse">
                  <!-- navbar content with links -->
                </div>
              </div>
              <div class="col-sm-2 col-md-2">
                <login></login>
              </div>
            </div>
          </div>
        </nav>
      </header>
      <!-- your main content should be here -->
    </div>
  </body>
</html>
```

RFSH:
We had a discussion about the folder name and we decided that we should create a folder called `lib` in chaise. Then you can move these new files to that location. You can do `/lib/login/<.js and .md files here>` or just `/lib/<.js and .md files here>`. I think the `/lib/login/` would be more organized.

loginApp/login.app.js
```diff
@@ -0,0 +1,99 @@
...
+            'chaise.authen',
+            'chaise.login',
+            'chaise.modal',
+            'chaise.utils',
+            'ermrestjs',
+            'ngCookies',
+            'ngAnimate',
+            'ui.bootstrap'])
+
+            .config(['$cookiesProvider', function ($cookiesProvider) {
```
jrchudy: The only nitpick I have is that there is no need to have 3 separate configuration blocks.
```js
.config(['$cookiesProvider', '$uibTooltipProvider', '$logProvider', function ($cookiesProvider, $uibTooltipProvider, $logProvider) {
   $cookiesProvider.defaults.path = '/';
   $uibTooltipProvider.options({ appendToBody: true });
   $logProvider.debugEnabled(chaiseConfig.debug === true);
}])
```

jrchudy:
I agree with @RFSH that the html page should not be included as part of the repository and should instead be included as a sample html page. This can either be a simple html file with a name like, `login-sample.html` or it can be included as part of the readme document.

Also please add some comments in the JS or HTML as to why certain decisions were made. Notably why the scripts are being injected the way they are and how you are attaching the "ng-app" to the DOM.

We should place the files at the top level of the chaise directory (i.e. `/chaise/lib/login/*.js` and so on)

Divyata-Singh:
Here are the 2 links: 
1) https://dev.rebuildingakidney.org/~divyatas/sample-loginOnRbk.html
2) https://dev.rebuildingakidney.org/~divyatas/sample-login.html

jrchudy:
Awesome! You can merge this when you are ready.

Issue #1705, opened 2019-01-26
Show inline related tables in the ToC


jrchudy:
The Table of Contents panel on the right side of `record` app only shows the related tables that show up below the main entity section, it should include the related tables that are part of the main entity section as well.

jrchudy:
The changes for this issue have been merged into master, assigning to @hongsudt for verification.

PR #1706, opened 2019-01-29
added inline related tables to ToC

jrchudy:
This PR has changes that will resolve issue #1705 . We already had all the information we needed for this functionality, just needed to add a function to `record.controller.js` to check properly. Added appropriate test as well to `record/presentation.spec.js`.


RFSH:
I approved the PR by accident. The click event on the inline tables in toc is buggy. It's currently throwing a terminal error.

jrchudy:
@RFSH  I added a new function to handle the scroll to event properly for inline tables.

PR #1708, opened 2019-02-02
Navbar submenus open on click

jrchudy:
This PR changes the behavior of the navbar dropdown menus and sub menus. Before you could hover over the dropdown menus and they would open automatically. This PR removes that functionality so that sub menus will only open on click.


jrchudy:
@RFSH I looked into what we discussed Friday about dropdown toggle for the parent and dropdown and the submenus. It looks like the preferred method is to use data-toggle for the parent and onclick events for the submenus. This is what I found from [w3schools](https://www.w3schools.com/bootstrap/bootstrap_ref_js_dropdown.asp) and other places when looking for a solution.

PR #1712, opened 2019-02-06
Support for Custom Error messages for deriva-webapps

Divyata-Singh:
- Added a new CustomError type to the "Errors" factory
- Currently, the CustomError has 4 parameters for error header, message,redirectUrl and ok button message.
- Didn't add support for loginLink as session management is not implemented in any of the deriva webapps currently. Also didn't add subMessage and stack trace.

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -732,6 +732,10 @@
         // takes pathname attribute of window.location object and returns app name
         // path should be a string literal which appears before #catalog id in URL (/chaise/recordset/)
         function appNamefromUrlPathname(path){
+            if (path.includes("deriva-webapps")) {
+                var app = path.substring(path.indexOf("deriva-webapps"));
+                return app.substring(app.indexOf('/') + 1, app.lastIndexOf('/'));
+            }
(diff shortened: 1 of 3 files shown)


common/utils.js
```diff
@@ -732,6 +732,10 @@
         // takes pathname attribute of window.location object and returns app name
         // path should be a string literal which appears before #catalog id in URL (/chaise/recordset/)
         function appNamefromUrlPathname(path){
+            if (path.includes("deriva-webapps")) {
```
jrchudy: The changes you made won't have an affect on this function, so there's no need to include this. 

The places that we call `appNamefromUrlPathname` have really specific use cases. In `modal.js`, vm.params.appname` is only used to check if we are in `recordedit` to show a reload button. In `utils.js`, this function is called as part of the `chaiseDeploymentPath` function. We would not want to match on a deriva-webapps path and incorrectlyget the chaise path.

Divyata-Singh: Removed this appName change.

PR #1717, opened 2019-02-12
Use ok button to dismiss instead of redirect

Divyata-Singh:
- In the boolean-search app there is a requirement to dismiss the error modal on clicking the ok button


Divyata-Singh:
Moved the dismiss logic to the html

PR #1720, opened 2019-02-13
Live button for versioned pages

jrchudy:
This PR adds a button in the navbar to redirect back to live data from a versioned page. This requires the navbar templating PR to be merged first because it relies on changes from that branch.


jrchudy:
@RFSH  I merged the navbarMenu templating issue (which this PR relied on) and reran the tests without any issues. It doesn't look like this is a feature we can test unless we go to a page with a fake versioned catalog. The error _could_ be dismissed and the live button can be clicked to then change the page to a working one (if we want to test this with e2e tests).

Issue #1408, opened 2017-12-13
Support array column type in recordedit


hongsudt:
The array column type's contents can only be values of scalar types.

In the `recordedit` form, need to support the following functionality:
 - adding more cells for the array
 - removing cells from the array
 - reordering the data in the array
 - edit any one of the cells

This issue is currently relying on [this PR](https://github.com/informatics-isi-edu/ermrestjs/pull/566) in `ermrestJS`. It adds more functionality for tinkering with the display of array columns.

NOTE: 
 - No array of type `markdown`.
 - This allows for a third level of types to be defined. (domain type that is of type array that has scalar types inside)

amitjha21:
As per today's discussion, 4 different cases pertaining to base type and domain type.

No. | isArray | isDomain | Typename | Base type
------- | ------- | ------- | -------| -------
1| False | False | Text|
2| False | True | Markdown| 1
3| True | False | Text [ ] |1
4| False | True | "Tex" |3

RFSH:
I created an issue for the changes that we need to have basic array support in the read-only modes.

jrchudy:
- [ ] Types that have yet to be implemented and tested fully:
 - float
 - date
 - timestamp
- [ ] Test cases are very minimal and need to be extended to each of these types.
- [ ] presentation bug when a new cell is added, it duplicates column header.
- [ ] HTML changes (needs more information)

hongsudt:
A simple array support that shouldn't take a lot of time is to provide a adjustable text box and validate to make sure that it follows the json array syntax.

RFSH:
#### Required features and How it should behave
- Chaise should detect array columns. You can do this check by looking at the `Column.type.isArray` attribute.
- When there's no existing data (in create or edit mode), we should show one input alongside two buttons. One for adding new input and one for removing the existing input (each input should have the remove button beside it). The latter should be disabled if there's just one input showing. The ordering of inputs matter too. I don't think this is in the requirements but we might want to add a way to change the ordering of inputs. There are some directives that might give you all these features.
- When there's an existing value we should put each of the array values into different inputs.
- Having just one input that is empty should be translated into `null` value instead of an array of `null` ([`null`]).
- If there are more than one inputs, we should treat the empty ones as null value. So for example if there are two inputs and the first one is empty and the second one has a value of "test". We should send a `[null, "test"]` to ermrestjs.

#### How to create array column
[...]

PR #1726, opened 2019-02-21
recordedit simple array support.

RFSH:
This PR will add simple array support to recordedit. We are going to show a texarea for arrays and it has its own set of custom validators.

I added test cases to `edit.spec.js`, `add.spec.js`, and `add-x-forms.spec.js` (batch-copy feature).

The only thing that I wasn't sure of is the timestamp format. Which formats should we allow? Should we be flexible i.e. allow only dates, allow timestamptz without specifying timezone? Currently I haven't specified any formats, but I feel like I should just allow a certain list of formats. We might want to discuss about this in the meeting.

#1408


RFSH:
It seems like moment has a pre-defined ISO 8601 format. So I just used that for validating timestamp columns.
@jrchudy the PR is ready to be reviewed.

Issue #1725, opened 2019-02-20
Improve Pre-fill When Creating New Related Entities


RFSH:
When users click on "add" button for related entities, we navigate them to a recordedit page with pre-filled value for the foreign key relationship with the main table. We are currently only doing this pre-fill based on the foreign key value. We could improve this by pre-filling other foreignkeys as well. 

To better understand this, take a look at the following example:

- User navigates to the record page for `Study` table.

- `Study` has two keys.
  
  ```
  "keys": [
    {
      "unique_columns": ["RID"], 
      "names": [["s", "key1"]]
    },
    {
      "unique_columns": ["RID", "Data_Provider", "PI", "Consortium"],
      "names": [["s", "key2"]]
    }
  ]
  ```
- `Experiment` has two foreign keys to the `Study` table based on the defined keys and all the columns are not-null. 

- We are showing `Experiment` on the related section based on the `RID` foreign key.

- User clicks on "Add" button in `Experiment` table to create new rows in relation with the displayed row of `Study` table. Here, we expect two foreign keys to be pre-filled because `Study.RID` alone is determining the one-to-one relationship of this row to the rows that we want to create for `Experiment`.

This can be generalized as following:

- `T2` has multiple foreign keys to `T1` (`FK`, `FK1`, `FK2`, ...).
- `T2` is displayed in the related entity section of `T1` based on `FK` foreign key and what user sees in the record page is based on this relationship.
- Let's call the keys that all their columns are part of `FK`, "pre-fill keys". In the example above, `["s", "key1"]` is the only pre-fill key.
- When user clicks on the add button, in the opened recordedit page we should pre-fill all the foreign keys that all their columns are at least part of one of the pre-fill keys.

karlcz:
As you said, the second foreign key column set must be a superset of the first foreign key columns in order for us to infer anything about its possible contents. Thus, the pre-fill for the first foreign key partially fills out the second foreign key.

Our conjecture is that the second foreign key must point to the same record and we can therefore copy over additional values to fill in the rest. We need to actually prove that by checking more conditions:

- The second foreign key points to a uniqueness constraint that is a superset of the uniqueness constraint used by the first foreign key.

Rationale: If the smaller set of columns is a unique key for the parent record, then that key plus additional columns MUST also be a unique key to the same parent record. There is no other possible configuration (other than NULLs) that can satisfy the schema constraints and allow the record to be written.

- The extra columns of the second foreign key have `nullok: false`.

[...]

PR #1730, opened 2019-02-27
Prefill fks

RFSH:
This PR will implement the changes required for #1725. To summarize:

- The cookie structure that we're using for sending pre-filled data has been changed to support filling multiple foreignkeys (We might be able to pre-fill more than one foreign key column). The new structure is as follows:
  ```
  {
    "constraintNames": [], // an array of foreignkey names that can be pre-filled
    "keys": {}, // key-value pair of raw values of foreignkey columns that can be pre-filled 
    "origUrl": "" // the reference url of the parent. to create a reference object and read the data
    "rowname": " // the rowname of parent record that should be shown in the foreignkey column
  }
  ```

- Changed the logic for "addRelatedRecord" to look for any extra foreign keys that can be pre-filled. This will look for foreign keys that are superset of the `origFKR`, and all of their extra columns are "not-null".

- Moved the pre-fill logic in `recordedit.app.js` to be inside the "CREATE" mode. I did this to increase the readability of the code. The pre-fill logic only applies to this mode.

- Added a new `isDisabled` attribute to `columnModel` object in recordedit. Previously we were passing the cookie object around just to check for `isDisabled`. In some places we were not even passing this value. The value now is computed once and is stored to be used in different places.

- Previously we were attaching the raw values of the pre-filled columns to `submissionRows` but no the `rows`. This would cause confusion for users since they couldn't see the value but we still were sending them to ermrest. I fixed this and also made sure to pre-fill a foreign key that all of its columns are pre-filled.

This PR relies on ermestjs changes in `default-display` branch.


RFSH:
We realized that we don't have to pass the foreign key constraint names that can be pre-filled. We can just pass the column values and recordedit can decide which foreign keys should be pre-filled. But I didn't remove the code for now. I explained this in the code why I didn't change the code for now.

I also created a new `record.md` document in user-docs. It's supposed to summarize all the features of record app in one place. For now, I just explained this current feature of "adding related entities". To better explain the record app structure, I created a minimal wireframe using [sketch app](https://www.sketchapp.com/) and pushed its source file too. I also realized that I should not continue hosting doc images on isrd-dev and we should just push them to chaise. We should follow this pattern for any new features that we are going to add. We should complete the `record.md` gradually and create the same docs for our other apps too.

Issue #1721, opened 2019-02-14
Add support for chaise-config.configRules


jrchudy:
As mentioned in #1704 , the chaise-config syntax is being extended to include another property called `configRules`. This property will be defined as follows:
```
"chaise-config": {
  other chaise config properties
  ...
  "configRules": [
    {
      "host": [], // array of host names
      "config": {
        any chaise config rules excluding "configRules"
      }
    }
  ] // end configRules
}
```

Chaise should be updated to add support for the config rules functionality.

jrchudy:
- Generally follow the existing chaise-config syntax and extend the config with rules that can overwrite some values for different server name.
- Use the same syntax in `"tag:isrd.isi.edu,2019:chaise-config":{ // chaise-config syntax }`
- Follow the order of rule-overwriting according to the section above. 
- The advantages are: 1) it is consistent with existing chaise-config; 2) it allows an explicit section to describe generic config parameters; 3) the document is self-contained and can be used for specific snapshots; 4) the navbar can be written per host for static site, and/or the same config can be moved to catalog annotation. 

```
  "configRules": [
     { 
       "host": <hostname> | [<hostname>, ...],
       "config": {
          // configuration properties here
        } 
    },
    {}
  ]
```
Example:
```
{
   // generic chaise-config
    catalog: 2,   // default catalog
    headTitle: "RBK/GUDMAP",
    navbarBrand: "/resources/",
    navbarMenu: {
   },
    "configRules": {
     { 
       // only support host as part of the configuration at the moment
       "host": ["www.rebuildingakidney.org", "staging.rebuildingakidney.org", "dev.rebuildingakidney.org"],
       "config": {
        "headTitle": "RBK/GUDMAP",
        "navbarBrand": "/resources/",
        "navbarMenu" : { 
           // rbk specific navbar
          { "name": "Protocols", "url": "/chaise/recordset/{{{$catalog.snahsot}}}/Protocol:Protocol/Consortium="RBK'"}
        } 
   },
   {
       "host": ["www.gudmap.org", "staging.gudmap.org", "dev.gudmap.org"],
       "config": {
        "headTitle": "GUDMAP/RBK",
        "navbarBrand": "/",
        "navbarMenu" : { 
          // GUDMAP specific navbar
          { "name": "Protocols", "url": "/chaise/recordset/{{{$catalog.snahsot}}}/Protocol:Protocol/Consortium="GUDMAP'"}
        } 
    }
```

[...]

PR #1734, opened 2019-03-05
changes to support configurations per hostname

jrchudy:
Changes to support chaise configuration per hostname.


common/utils.js
```diff
@@ -1283,6 +1282,20 @@
                     }
                 }
             }
+
+            if (chaiseConfig.configRules) {
```
RFSH: I feel like we should do more checking here. If any of these properties are ill-defined, this implementation will throw a terminal error. These are the changes that I suggest:

- change line 1286 check to `Array.isArray(chaiseConfig.configRules)`
- Based on the issue, `configRules` is written as `config-rules`. Either change the implementation or documentation. 
- Based on the issue, `host` can be a single string too. So you could just check for that first and then turn it into an array of one.
- Make sure `host` is array and then do the `forEach` (`Array.isArray(ruleset.host`). 
- (not important) since you just need to match with one of the hosts in the ruleset.host, you can turn the `ruleset.host.forEach` into a simple loop and break out of it when the host matches. Although since the host array is usually small, this shouldn't matter.
- Make sure `ruleset.config` is defined and is an object before looping through its properties.

RFSH:
Apart from the comment, I feel like we need a proper documentation for this, which you can include in this PR. Most probably you can just copy and paste the issue description in that document.

jrchudy:
@RFSH  I updated the code with your suggestions. I included documentation for this in `chaise-config.md`. I also included an example in `chaise-config-sample.js`. The [attached issue](https://github.com/informatics-isi-edu/chaise/issues/1721) has a comment on it further explaining the nuances of this feature

PR #1743, opened 2019-03-14
Changes to support proper live link with version when no version is present

jrchudy:
Changes for this feature have been completed. When the user navigates to a record page, regardless of it being versioned or not, the share dialog will have both a live link and a versioned link. The headings for these links have been changed too.

Other smaller features include:
 - If you load the record page to a live page then open the share dialog, a check is made to see if you are actually looking at the live content
   - a notification will be shown if you are viewing out of date content
 - Functions moved to a utility factory for re-ruse
 - other changes made in corresponding [ermrestJS PR](https://github.com/informatics-isi-edu/ermrestjs/pull/767)

### DIFF
--- a/common/modal.js
+++ b/common/modal.js
@@ -296,16 +296,21 @@
         }
     }])
 
+    /**
+     * Params object values:
+     *   - {String} displayname - used for citation content and filename for bibtex download
+     *   - {String} permalink - link to the live catalog
+     *   - {String} versionLink - link to the current version of the live catalog
+     *   - {String} versionDate - version decoded to it's datetime
+     *   - {String} versionDateRelative - version decoded to it's datetime then presented as relative to today's date
+     *   - {boolean} showVersionWarning - decides whether to show "out of date content" warning
+     *   - {Object} citation - citation object returned from ERMrest.tuple.citation
+     *
+     */
     .controller('ShareCitationController', ['$uibModalInstance', '$window', 'AlertsService', 'params', function ($uibModalInstance, $window, AlertsService, params) {
         var vm = this;
-        vm.cancel = cancel;
-        vm.citation = params.citation;
-        vm.permalink = params.permalink;
-        vm.versionLink = params.versionLink;
-        vm.versionDate = params.versionDate;
-        vm.versionDateRelative = params.versionDateRelative;
-        vm.filename = params.displayname;
-        vm.showVersionWarning = params.showVersionWarning;
+        vm.params = params;
+        vm.warningMessage = "The displayed content may be stale due to recent changes made by other users. You may wish to review the changes prior to sharing the <a ng-href='{{ctrl.params.permalink}}'>live link</a> below. Or, you may share the older content using the <a ng-href='{{ctrl.params.versionLink}}'>historical link</a>.";
 
         vm.moreThanWeek = function () {
             var weekAgo = moment().subtract(7, 'days').startOf('day');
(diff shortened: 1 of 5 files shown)


common/utils.js
```diff
@@ -767,20 +767,82 @@
...
+            createRedirectLinkFromPath: createRedirectLinkFromPath,
             fixedEncodeURIComponent: fixedEncodeURIComponent,
+            getQueryParams: getQueryParams,
+            isBrowserIE: isBrowserIE,
             parseURLFragment: parseURLFragment,
-            setOrigin: setOrigin,
             parsedFilterToERMrestFilter: parsedFilterToERMrestFilter,
+            queryStringToJSON: queryStringToJSON,
+            resolveLivePermalink: resolveLivePermalink,
+            resolveVersionedPermalink: resolveVersionedPermalink,
```
jrchudy: I have 2 concerns with these 2 functions I added:
 1. They are almost identical to each other except the extra `versionString` calculation
 2. The properties I rely on in those functions (tuple and reference) are being passed as parameters. In both places that this is done, those parameters are passed from the `$rootScope`. These parameters could be removed in favor of the functions assuming they will always work off of the properties defined on the `$rootScope`. The current implementation allows for more future flexibility, but I'm not sure if that is even necessary to worry about.

common/modal.js
```diff
@@ -305,13 +305,20 @@
...
 
         vm.moreThanWeek = function () {
             var weekAgo = moment().subtract(7, 'days').startOf('day');
             var versionMoment = moment(params.versionDate);
 
             return weekAgo.isAfter(versionMoment);
         }
+
+        if (params.showVersionWarning) {
+            vm.alerts = AlertsService.alerts;
```
RFSH: The current implementation has some issues that by creating a special alert (separated from alert service) we can just avoid.  

One issue is showing the existing alerts in the share modal. if there are some existing alerts (the session expire alert for example), those will be shown in the share modal as well. Another one is the fact that each time users click on the share button, a new alert is being created. So if you keep clicking on “Share” button, they will stack up. Besides I feel like we might prefer only showing this alert in the share modal (The alert description matches the modal and not the main record display).

Instead of reusing the alert service, you can just show an alert box with the same html structure in the `shareCitation.modal.html`. This will duplicate the html structure of alert, but I feel like it will avoid all these unnecessary logic that you have added to alert service (and most probably have to add more to fix those bugs).

common/utils.js
RFSH: 1. two functions:
If I remember correctly, the first implementation had more branching logic for "versioned" permalink. But based on this latest implementation, I don't see the point of having two separated functions. Based on what I understood the only difference is that in the versioned one, you're always  adding `versionString` to the result of the other function. 

If you want to have two functions without an optional parameter, you could call the "unversioned" one in the other function.

```javascript
function resolveVersionedPermalink(tuple, reference) {
  var version = reference.location.version;
  var versionString = "@" + (version || snaptime);
  return resolveLivePermalink(tuple, reference) + versionString;
}
```
It's up to you whether you want two functions. I don't have any strong opinion about it.

2. Since you're passing those variables, I would say leave it the way it is. Don't worry too much about it. You can change them later.

jrchudy:
@RFSH the suggested changes have been made. I went with the optional parameter approach because in one case, the version is being placed in the middle of the `url` string. In the case that the user is viewing a live catalog, the catalog number is replaced with the `<catalogId> + @<version>`.

I changed the alert to be contained to the share dialog. It makes more sense based on what you said, mostly this line:
>The alert description matches the modal and not the main record display

PR #1746, opened 2019-03-20
hack to replace # with? in the chaise url 

RFSH:
This PR will add changes required for [ermrestjs#769](https://github.com/informatics-isi-edu/ermrestjs/issues/769).

Using the Google SearchConsole, I was able to send a request for indexing last week, but for some reason it's not showing the index. I send another request to index the page today. I will update the PR with the result of indexing.

It's worth noting that the "live version" of the Google SearchConsole is not rendering the Angularjs page and it's showing an empty page. So I don' think google could be able to index this and we should find more information about indexing an Angularjs app.


RFSH:
I am going to remove Josh and I from this PR for now. The whole purpose of this PR was announcing this change. We are not going to merge this PR if these changes are not helpful in google indexing.

ljpearlman:
It works! Google indexes these pages and finds actual content.

RFSH:
@jrchudy I updated this branch with the master and made some more changes. It's ready to be reviewed.

jrchudy:
Fixed a bug that was introduced in resolve permalink function. Fixed a failing test case because of using the old unique id instead of RID. Tested this by changing all links in `record` specs to have `?` instead of `#`.

Issue #1740, opened 2019-03-12
logging issues


hongsudt:
I am going to summarize the issues I found here, so that people might be able to refer to it when they look at their logs:  (as of 2019/03/13)
**Chaise**
- There are cid = 'record', but action is 'recordset/main'. This is caused by ... 
- There are cid = 'recordset', action='recordset/main/load' without table. These requests have incorrect action. They belong to the following events: 
  - pseudo-columns that ask for count e.g. `/ermrest/catalog/2/attributegroup/M:=RNASeq:Experiment/value:=Markers;count:=cnt(*)@sort(count::desc::,value)?limit=26`
  - the count used to show the facet list e.g. 
- viewer app only has logs that ask for schema
- no dcctx in webautn requests

**Webapps**
- heatmap has cid='null' and there is no detail in the log
- treeview/boolean-search need to add logging

**Log servers**
- Extract uinit from the r_path 

**Hatrac**
- incorrect content-type (see example below) 
    - @karlcz mentioned that this is because  log entries are buggy. They duplicate the same value in the `"type"` and `"range"` fields of the raw systemevents JSON blob. This may have been fixed in webauthn on 2018-10-26.  it was a bug in the common logging code implemented there and used in ermrest, hatrac, and exporter
```
708842289 | 2018-12-03 15:21:49 | HEAD     | /hatrac/resources/rnaseq/study/14-3PXT/experiment/14-3PY4/replicate/14-3PZ8/6007ec003038b604af54ed157d7a73f7.bam                                       | */0
708842156 | 2018-12-03 15:21:31 | POST     | /hatrac/resources/rnaseq/study/14-3PXT/experiment/14-3PY4/replicate/14-3PZA/6007ec003038b604af54ed157d7a73f7.bam;upload/IUB7WUCDNZ5TTUDQNZIYJN534U     | */140
```

**Others**
- There is no cid that are non-chaise
- There are many requests without u_track e.g. r_path=/internal/, webauthn path, etc. 
- No logging for navbar loading
- No logging for navbar links (not sure if possible)
- Should template be included in dcctx header in log?

RFSH:
Chaise bugs have been fixed. Please refer to the [change log](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/logging.md#041819) for more information about the changes.

Issue #1518, opened 2018-05-10
Cleanup Log Object


RFSH:
We designed log object a while ago. We should take a step back and take a look at what we are logging, we might want to remove/add some attributes. Or change the way we are logging some of the requests.

RFSH:
These are some of the changes that we have to make:

- [ ] Add action to the `/catalog/N/schema` request that we're doing during the introspection.
- [ ] Add `catalog` alongside the `schema_table`. This will the catalog and its version ("N@snapshot").
- [ ] Add `column` to request for getting scalar facets data.

RFSH:
The changes have been done. Please refer to the [change log](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/logging.md#041819) for more information.

Issue #1516, opened 2018-05-10
Mechanism to Find Referrer App (PID, etc.)


RFSH:
Currently we navigate users between different apps. But these navigations restart the state of chaise and we don't have any way of knowing how users end up in a chaise app.
We should modify the URLs that we are generating in chaise to include the necessary attributes (most probably as query parameter) to track the referrer app. This can be achieved by just passing the PID, considering if PID is actually universally unique. If it's not we should pass cid, and wid too (considering this combination is always unique).

We mainly need this for log purposes, but this might be useful for other features.

RFSH:
We can simply change the `applink` function that is defined in chaise to include the `ref_pid` as a query parameter. So when user navigates to other parts of the app via links and buttons that we provide, the url will include this query parameter and the second page has knowledge of `ref_pid`.

Now the question is that in the second page, to which request should we attach this `ref_pid`? Should it be attached to all the requests (same as `cid`, `wid`, and `pid`)? Or just the first request on that page? What is the definition of first request on the page? the request for getting the schema? Or the main read? In case of recordedit in create mode, there's no main read.

RFSH:
new name: `ppid`, `pcid`

example.com/asdasdasdas?pcid=whatever&

example.com/asdasdasdas?pcid=whatever&


attach to the main request of each app
- recordedit: submit
- recordset and record page: main read

RFSH:
The changes have been done. Please refer to the [change log](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/logging.md#041819) for more information.

PR #1748, opened 2019-03-25
Fix Log Issues

RFSH:
This PR will fix the log issues that was discovered recently. I spent a little more time and found some other issues. All of these are documented in the "change log" section of `logging.md` (the following is the exact copy of the change log).

**Changed**
- Changed the action for removing related entity rows from `delete/recordset/related` to `delete/record/related`.
- Fixed a bug that would cause record load requests to be reported as update.
- Fixed the issue of reporting both load and update requests of record page as `record/main`. Now they are being logged with proper action.
  - Impact: We cannot determine the number of record page loads prior to this fix. This only affects deployments that allow their users to create new records since the update request would only fire if the user updates any of the related entities.
- `recordset/export` was used as action for export from both record and recordset which is changed to `export`.
- Changed the `template` object that is reported with `export` action to only include the `displayname` and `type`.
- Instead of logging the whole template document in the `export` action, we're just going to include `displayname` and `type`.
  ```
  {
    "action": "export",
    "template": {
        "displayname": "sample template",
        "type": "BAG"
    }
  }
  ```
- Fixed the action of scalar column requests.
- Changed the `recordedit/default` action to `default`.

**Added**
- Added proper `dcctx` logging to viewer app.
- `ppid` and `pcid` has been added to the main request of each app which shows the parent `pid` and `cid`.
- New actions have been attached to schema (`"model/schema"`) and catalog (`"model/catalog"`) requests.
- `catalog` has been added to all the requests.
- `column` has been added to requests for getting a scalar facet data.
-  If there are preselected facets, in entity mode we have to get the row-name corresponding to the selected value. Customized action (`recordset/facet/init`) has been added for this request.
- Added proper action `recordset/facet/histogram` to the request for getting buckets of a range picker facet.
- Added `cfacet`, `cfacet_str`, and `cfacet_path`.

ermrestjs PR: [ermerstjs#773](https://github.com/informatics-isi-edu/ermrestjs/pull/773)
related issues: #1740 #1516 #1518 [ermrestjs#768](https://github.com/informatics-isi-edu/ermrestjs/issues/768)


RFSH:
@jrchudy I updated the code and documentation to reflect the changes after the discussion that I had with @hongsudt. Both PRs are ready to be reviewed and merged.

RFSH:
@jrchudy I changed the test cases that you mentioned. Can you please try running the test cases again?

jrchudy:
The test cases are passing for me now.

PR #1752, opened 2019-04-03
fix hatrac completion bug

RFSH:
This PR will fix the recent issue that we found in hatrac. All the code in hatrac to check whether we should start the next step was relying on counters. I changed all of those to check the booleans that we are already keeping track of. 

I also changed the job completion step to be serialized. Previously we were firing all these request at the same time, but now they will be fired after the previous one is completed.

The other change is how we're handling the existing files in the server. Previously we were firing a complete upload request immediately and we would skip the job creation/upload steps. But still when we would get to the last step, we were firing duplicated callbacks in ermrestjs to complete the job again. Now they will be treated the same in chaise, and in ermrestjs if the file is already in the server and therefore marked as `jobDone`, I am just returning the `chunkUrl` instead of actually creating a job and uploading.

related ermrestjs issue: [ermrestjs#774](https://github.com/informatics-isi-edu/ermrestjs/pull/774)

### DIFF
--- a/common/upload.js
+++ b/common/upload.js
@@ -105,7 +105,7 @@
                 If the file was marked to be skipped, the upload job is marked as complete (and never created)
 
             5.  StartUpload functions calls the start function in hatrac.js for the hatrac hatracObj for all files.
-                It keeps track of the upload progress for each file and once all are done it calls completeUpload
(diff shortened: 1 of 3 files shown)


jrchudy:
Changes look good. I don't have any other comments to add outside of what we talked about earlier this morning. All e2e tests passed for me locally.

PR #1758, opened 2019-04-09
Catalog annotation

jrchudy:
This PR adds the catalog annotation support. This should be merged at the same time as the corresponding [ermrestJS PR](https://github.com/informatics-isi-edu/ermrestjs/pull/770).

## Code
The bulk of changes in this PR were made to support setting up the `chaiseConfig` object as the first step of our application before the configuration apps bootstrap the actual app. A new module was created called `chaise.config` which will read the chaiseConfig document from the server and attach it to the window object. We now wait for `ermrestJS` to finish loading before fetching the catalog object and seeing if a configuration annotation was defined. This is where the `setConfigJSON()` is called where the defaults are applied and any further properties defined in the catalog annotation.
  - `config.js` added as its own module with a run block to execute on inclusion.
     1. does the $window configuration for each app now )set window name and other client context variables)
     2. sets the chaise configuration (`setConfigJson()`) based on `chaise-config.js` on the server so other modules can use the server configuration if necessary
     3. waits for `ermrestJS` to finish loading so the server object can be fetched with client context variables for logging
     4. checks for a catalog annotation and sets the configuration (`setConfigJson()`) again if available so it updates previously set values
       - added a conditional to this function which checks if the catalog annotation object is indeed an object before updating the configuration properties
     5. once configuration is complete, `configuration-done` event is emitted. the configuration app that is attached to the DOM will be listening for that event so it can attach the chaise `app` the user is trying to access
 - refactored some functions in `UriUtils` so that `chaiseURItoErmrestURI` returns an object with the parsed tokens from the url. This was done to avoid having multiple functions parse the url for individual pieces of it
    - added `getCatalogId()` which wraps the above mentioned parse function in case no catalog id is present in the url. there are 2 cases where we want the catalog id
       - `navbar.js` needs it for templating the urls
       - `config.js` needs the id to fetch the catalog from the server. If no catalog, don't try to fetch the server object
    - possibly change `chaiseURItoErmrestURI` to store the parsed `object` so if it's called again (and the url didn't change since before), the object is returned instead of reparsing
    - another change could be to instead store the parsed url object (chaise location object) with the dcctx
 - Added `getContextJson()` to `ConfigUtils` that fetches the `dcctx` from the `$window` object
     - each app will no longer define its own `context` constant and pass it around. Instead the $window context object will be used
     - instead each configuration app has an `appName` constant that is passed to the config module to use as the `cid`
 - Added `$sce.trustAsHtml` to a few strings that we were trying to present in the DOM that weren't being rendered.
 - `login.app.js` and `navbar.app.js` include the configuration module. Changed a lot of spacing in each file
 - `scripts/app.js` is the entry point for the old search app. The configuration module needed to be included here because the navbar is injected into this old app
 - `ermLoginController` is the entry point for the old login app. This also includes a navbar so the configuration module needed to be included here too

## Tests
A new test spec was added to test another state of the navbar for when a `catalog-annotation` is available by overriding values set by `chaise-config.js`
 
Other major changes in this PR were made to stabilize the tests based on this change to the configuration. The configuration now causes the app to load a little bit slower than before (because we are waiting on `ermrestJS` to be available first). Some of the UI elements show up in different orders than they did before. The changes to the wait conditions are more appropriate to the test case we are trying to test rather than being a "generic" wait condition.


record/record.app.js
```diff
@@ -10,10 +11,15 @@
         'ui.bootstrap'
     ])
 
-    .run(['ERMrest', function (ERMrest) {
+    .run(['ERMrest', 'ConfigUtils', 'UriUtils', '$window', function (ERMrest, ConfigUtils, UriUtils, $window) {
         ERMrest.onload().then(function () {
-            angular.element(document).ready(function(){
-                angular.bootstrap(document.getElementById("record"), ["chaise.record"]);
+            var urlParts = UriUtils.extractParts($window.location);
```
RFSH: I'm a bit confused. If we're doing this logic here and not in the `config.js`, then what's the whole point of `config.js`?

By the way calling `catalogs.get` will introspect the whole catalog and since we just want the annotation, we might want to create another specific api that only calls `catalog/N` and returns the annotation. Although since we're always going to introspect the schemas that shouldn't matter.

RFSH: We have to pass all the `cid`, `pid`, and `wid` to the getServer function.

jrchudy:
@RFSH this Pr has been updated based on our discussion and is ready for review.

common/config.js
```diff
@@ -0,0 +1,32 @@
...
+        // initialize dcctx object
+        $window.dcctx = {}
+        // set chaise configuration based on what is in `chaise-config.js` first
+        ConfigUtils.setConfigJSON();
+
+        ERMrest.onload().then(function () {
+            var cc = ConfigUtils.getConfigJSON(),
+                service = cc.ermrestLocation,
+                // Get the catalog id and strip off the version
+                catalogId = UriUtils.getCatalogId().split('@')[0];
```
RFSH: Shouldn't we get the config from the versioned catalog?

common/config.js
```diff
@@ -0,0 +1,32 @@
...
+
+            if (catalogId) {
+                ERMrest.ermrestFactory.getServer(service).catalogs.get(catalogId).then(function (response) {
```
RFSH: You need to pass `cid`, `pid`, and `wid` to the `getServer` function.

common/utils.js
```diff
@@ -172,18 +172,42 @@
...
+        function getCatalogId() {
+            var chaiseConfig = ConfigUtils.getConfigJSON();
+
+            if ($rootScope.context && $rootScope.context.catalogID) {
+                return "" + $rootScope.context.catalogID;
+            }
+
+            // NOTE: what if it's '?' instead of '#'
+            if ($window.location.hash != "") {
+                return getCatalogIDFromLocation();
```
RFSH: Change `getCatalogIDFromLocation` to return false or null if the hash doesn't exist at all (instead of throwing the error, you should just catch and return false/null). Then in here, you can check for the string and use it.

common/utils.js
```diff
@@ -172,18 +172,42 @@
...
          */
         function getCatalogIDFromLocation() {
-            var hash = getLocationHash($window.location);
-            return hash.split('/')[0].slice(1);
+            try {
+                var hash = getLocationHash($window.location);
+                return hash.split('/')[0].slice(1);
+            } catch (exception) {
+                // pass to error handler
+                throw exception;
```
RFSH: This should fail silently as I explained in the other comment.

lib/login/login.app.js
```diff
@@ -2,6 +2,26 @@ function loadModule() {
     (function () {
         'use strict';
 
+        angular.module('chaise.configure-login', [
```
RFSH: Can you please list all the dependencies in here? Including the ones that `modal.js` and `utils.js` might depend on. So we have a list of modules that should not use `chaiseConfig` during the load.

RFSH: I commented here because this is the first place that I saw this. You might want to do it for all the places that we're using this module (all the apps)

lib/login/login.app.js
```diff
@@ -12,38 +32,38 @@ function loadModule() {
...
-             */
-            angular.element(document).ready(function(){
-                angular.bootstrap(document.getElementsByTagName("login")[0],["chaise.loginapp"]);
-            });
+        /**
+        * Manually initialize the ng-app
+        * Otherwise angular throws an error if ng-app is used in the HTML template as angular tries to load the module before all the dependencies have finished loading
+        */
+        angular.element(document).ready(function(){
+            angular.bootstrap(document.getElementsByTagName("head")[0],["chaise.configure-login"]);
```
RFSH: Was there a reason that you changes this to head? This might cause the issue that Cris found. We don't want to turn the whole HTML page to Angularjs and just the login/navbar apps.

jrchudy: I added a `configure-login` and `configure-navbar` angular-app that can't be on the same element or child element as another angular app (login app is attached tot he `login` tag element and navabr app is attached to the `navbar` tag element).

I didn't want to introduce another element that would force cris to have to change html templates for the static pages.

jrchudy:
@RFSH This PR is ready for review again after addressing the concerns we discussed this week and refactoring the configuration to be more consistent for each app (including login and navbar). The other major change was to remove some other functions in `UriUtils` to enforce the idea that we should try to use our `chaise` url parser (`chaiseURLtoErmrestURL`) instead of a handful of functions that all do parts of what that parse function does.

common/config.js
```diff
@@ -0,0 +1,36 @@
...
+        // set chaise configuration based on what is in `chaise-config.js` first
+        ConfigUtils.setConfigJSON();
+
+        ERMrest.onload().then(function () {
+            var cc = ConfigUtils.getConfigJSON(),
+                service = cc.ermrestLocation,
+                catalogId = UriUtils.getCatalogId();
+
+            if (catalogId) {
+                ERMrest.ermrestFactory.getServer(service, { cid: dcctx.cid, pid: dcctx.pid, wid: dcctx.wid }).catalogs.get(catalogId).then(function (response) {
```
RFSH: dcctx is not defined

common/config.js
```diff
@@ -0,0 +1,36 @@
+(function() {
+    'use strict';
+
+    angular.module('chaise.config', ['ermrestjs'])
```
RFSH: technically now this module relies on `utils.js`, so maybe we should list it here.

jrchudy: That's true. I'll add it to the dependency list

common/config.js
jrchudy: It technically is. Anything that is defined on `$window` (or in js, `window`), is able to be accessed as if it were defined in the current scope. I can change it so it's clear that we are fetching it from the `$window` object by adding a variable declaration in here.

RFSH: oh right. I didn't pay attention to that. I think it's better if you explicitly use `$window.dcctx` to be aligned with the file lines above it that you're attaching the value to `$window` object.

PR #1766, opened 2019-04-27
File input improvements

jrchudy:
Made changes to show the filesize in the file input tooltip in both create and edit mode. The metadata is also fetched and attached to the submission object in the copy case when the metadata is available.

Not sure if we should bother adding testing for a tooltip feature or test metadata is copied correctly when it's never displayed.


recordedit/recordEdit.app.js
```diff
@@ -335,13 +335,28 @@
...
                                             }
+
+                                            // if in copy mode and copying an asset column with metadata available, attach that to the submission model
+                                            if (column.isAsset && context.mode == context.modes.COPY) {
+                                                // may not have been set or fetched above because of disabled case
+                                                // we still want to copy the metadata
+                                                var metadata = column.getMetadata(tuple.data);
+
+                                                // I don't think this should be done brute force like this
+                                                if (metadata.filename) recordEditModel.submissionRows[j][column.filenameColumn.name] = metadata.filename;
```
RFSH: I could add a new API to the `getMetadata` because the `filename` that I'm returning is not necessarily the filename but rather the name that we can use to refer to the file. I can change its name to something else.

Also `column.filenameColum` and the rest might be undefined.

RFSH: The previous comment is not relevant anymore. After discussing about this last week, I separated the displayed value from the filename. So 
- `filename` will return the filename that is stored in database.
- `caption` will return the value that should be displayed for the asset

So as long as we make sure that we're storing the `metadata.filename` and showing `metadata.caption`, it should be fine.

jrchudy:
@RFSH changes have been made to use the metadata object and to default to the caption, if no filename was present.

common/upload.js
```diff
@@ -84,13 +85,16 @@
...
-                            }
+                            // value.filename will always be either the stored filename or the "caption" (text stripped of the hatrac path)
+                            return value.filename;
                         }
                     }
+
+                    scope.fileTooltip = function () {
+                        var value = scope.value;
+                        // value.filename will always be either the stored filename or the "caption" (text stripped of the hatrac path)
+                        return (value.filesize ? "- " + value.filename + "<br>- " + value.filesize + " bytes" : value.filename);
```
RFSH: You should call `humanFileSize` function that we have in utils.js. Instead of showing the raw byte count.

Also I'm not sure whether we want to handle this or even how we should handle this. But the `value.filename` won't be available in case of default values in create mode. We're calling into `formatFile` function which only is attaching `url` to the value. If we want to attach the `filename` by getting the caption from ermrestjs then we have to create a key-value pair of all the default values of the table. Or maybe just passing the url value itself is enough and ermrestjs will return the last part of url for you.

jrchudy: > Or maybe just passing the url value itself is enough and ermrestjs will return the last part of url for you.

That's what I was thinking. The generate caption function could be public on the `ermrestJS` object in `chaise` so it could be used in a similar way to `renderTemplates`. i.e. `ErmrestJS.getFileCaption(defaultUrl)`.

RFSH: We don't need any extra apis. the `getMetadata` can be used here too. You just have to pass an object with just the url column default value. ermrestjs should handle it. 

You can ignore the "we have to create a key-value pair of all the default values of the table". I was thinking about a case that shouldn't even happen. The case that I was thinking was when you have defined meta columns in your annotation. What I assumed is that in this case for caption we should use the default value of `filename` column. But I don't think that's the behavior that we want. If we want to do this, then we should also send the default value of filename to database (if users don't change the value of that file).

jrchudy: Oh right. I forgot `getMetadata` was written to accept an argument.

And yes, if we have a case where the asset column and filename column both have default values, we should honor them as is. The data modeler and user would want to see what the `asset` path is alongside the filename if both had defaults and were visable but diabled during create.

Issue #1741, opened 2019-03-14
Change $http to ERMrest http module


RFSH:
We should avoid using $http module directly and instead we should rely on the http module attached to the `Server` object in ermrestjs. This will gives us the ability to pass the `cid`, `pid`, and `wid` consistently throughout chaise.

PR #1772, opened 2019-05-03
Use ermrestjs http service instead of $http

RFSH:
The following are the three different places in chaise we're directly generating http requests. 

- Search app: 
  It's relying on the old ermrestjs that resides in chaise. We cannot do anything about this.

- viewer app: 
  There's a `getGoauth` function that is being called upon 401/404 error response while getting the session. Since it's using the `authen.js` this shouldn't even happen and the `authen.js` should handle this case. I'm not sure why this code path even exists in viewer app. Most probably we can just remove this code path (While I was fixing the viewer app I saw this config block and added a TODO with more details on why I think we should just remove it).  We should discuss about this

- `authen.js`: 
  The login/logout requests. In this PR, I modified the $http in this file, to use the new `window.dcctx.server.http`.

P.S. Instead of directly using `window.dcctx.server.http`, I added a `getHTTPService` for backward compatibility. We could remove it if you think it's not necessary. 

#1741

### DIFF
--- a/common/authen.js
+++ b/common/authen.js
@@ -3,8 +3,8 @@
 
     angular.module('chaise.authen', ['chaise.utils', 'chaise.storage'])
 
-    .factory('Session', ['ConfigUtils', 'messageMap', 'modalUtils', 'StorageService', 'UriUtils', '$cookies', '$http', '$interval', '$log', '$q', '$rootScope', '$sce', '$uibModalStack', '$window',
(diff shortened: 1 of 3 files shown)


RFSH:
After looking at viewer app, I decided to do the changes that I mentioned in the description. I removed the `getGoauth` since it's not used at all.

PR #1775, opened 2019-05-08
Sugest login in error

jrchudy:
Refactored the error flow a little bit so it's more consistent.

I moved all `$sce.trustAsHtml` calls for error messages to one place. Moved all message manipulation to the `handleException` function. Always add the `maybeShowLogin` message when the user is not logged in unless it's the `noRecordError` case, in which we show a different message.

@RFSH you mentioned in the issue that we should differentiate between the following 2 errors cases in `ermrestJS`:
 - when the facet string is completely invalid (and therefore cannot be decoded) 
 - when the facet string is correct but the generated path is invalid (possibly because user isn't authorized for parts of that path

Should this change still be made or do you think that distinction is not as important if we are _always_ going to show the login link when the user has no session?

### DIFF
--- a/common/errors.js
+++ b/common/errors.js
@@ -194,8 +194,8 @@
     }])
 
     // Factory for each error type
-    .factory('ErrorService', ['AlertsService', 'ConfigUtils', 'DataUtils', 'errorMessages', 'errorNames', 'Errors', 'modalUtils', 'Session', 'UriUtils', '$document', '$log', '$rootScope', '$sce', '$window',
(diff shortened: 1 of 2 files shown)


RFSH:
I don't know why I didn't see this PR until now. The changes are simple and based on what I understood are just refactoring, you could merge them without creating the PR.

And about the ermrestjs change, I'm not sure. I don't think that would be a priority for now. Eventually we should look back at ermrestjs error messages and improve them. I could change all the error messages to be more verbose and informative but that might be more confusing to the users.

PR #1781, opened 2019-05-11
File input improvements

jrchudy:
This PR has changes to further improve the consistency of file inputs in recordedit. Previously, a tooltip  was added to the file input that shows the filename and, if available, the filesize. To get this working properly in cases that the input is disabled (marked as generated or immutable), I had  remove the use of the `disabled` input when the column has an asset annotation.

The div that wraps the input has the tooltip because when the input is disabled, `<input disabled="true" \>`, hover events and other mouse events won't occur. 

Also added the use of `humanFileSize()` so the text is more readable for the users.

Added a test case for when there's a default set for a column with an asset annotation. As well as testing normal file input cases with tooltips.


test/e2e/specs/default-config/recordedit/multi-col-types.spec.js
```diff
@@ -118,9 +118,16 @@ describe('When editing a record', function() {
     // Tests that check the values for regular, non-disabled input fields are in 01-recordedit.edit.spec.js
     it('should display the correct values in disabled input fields', function(done) {
         testParams.table_w_generated_columns.row.forEach(function checkInput(col) {
-            var input = recordEditPage.getInputById(0, col.name);
-            expect(input.isEnabled()).toBeFalsy("col " + col.name + " was not disabled.");
-            expect(input.getAttribute('value')).toBe(col.value, "col " + col.name + " value missmatch.");
+            if (col.name == "asset_col_gen") {
```
RFSH: This is not really important and is just an observation. Since in the code the displayType of  disabled assets is not "disabled" anymore. You could modify the `displayType` of this column to be `upload`. And then in here instead of checking for the `col.name`, you could check its `col.displayType`.

Issue #1711, opened 2019-02-05
Error message on file upload in RecordEdit


Divyata-Singh:
When the network is disconnected while uploading a file in the recordedit app, the alert gives the angular stack trace instead of a proper error message.
![UploadError.PNG](https://images.zenhubusercontent.com/5b0458274b5806bc2bd42f17/f4f793d0-6fb0-46c3-a76c-ee1e8137449b)

jrchudy:
To reproduce this error:
>1. navigate to `recordedit`
>2. add a file to be uploaded
>3. before clicking submit, disable the internet connection
>4. click submit

The error occurs because the browser can't fetch the `uploadProgress.html` template. This error will occur anywhere in any of the apps if the internet is disconnected while the app is trying to fetch templates. The error is only exposed in `recordedit` though. In record and recordset, this error is swallowed up (never shown to the user).

This can be reproduced in `recordset` by waiting for the app to load, then disconnecting the network, and finally try to open a "show more" dialog.

This can be reproduced in `record` by waiting for the app to load, then disconnecting the network, and finally try to open an "add" dialog for a pure and binary related table.

jrchudy:
In both `record` and `recordset`, this error case recovers without incident if the network connection returns. The retry logic continues to run trying to fetch templates. When or if they resolve, those errors are swallowed up and no template is displayed to the user. If the user retries the action after the connection is restored, everything behaves as expected.

jrchudy:
@hongsudt and I discussed this issue on Friday and the plan is to show a dismissible error popup when the user encounters this error in any of the chaise apps. This error being when we have a network outage that doesn't allow the user to communicate with the server and fetch templates after the page has loaded (usually when a secondary action is taken that shows a modal dialog).

One reasoning behind this is to not distract the user (move their attention by scrolling or simply adding another colored element to the top of the page). This is going to be handled the same everywhere in the application as a "blocking" error that's dismissible because the user should take an action to verify that it's not a loss of network connectivity on their end.

PR #1782, opened 2019-05-13
No internet no templates

jrchudy:
Opening this PR so we can have a discussion about the changes. 

The error case described in issue #1711 occurs when the user is fetching a template but there is no network connection. In the `chaise` apps, this can only occur _after_ a page has loaded, so the user had a network connection at one point but then lost it. The only time that we try to "fetch" templates after a page has loaded, is when a user clicks on an action that opens a modal instance.

I made changes to `modalUtils` to handle all of these cases. I'm using the navigator object API, [described here](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events), to check the `onLine` property to determine what type of error case we are in. If the app is in an offline state, it will call into the added function that generates an HTML modal with JS instead of trying to fetch another template.

NOTE: I wasn't sure if there's a way to "pre-fetch" templates so they are always available (pre-fetch the error template instead so this isn't an issue with the error dialog). Something like a pseudo-cache.

### DIFF
--- a/common/errors.js
+++ b/common/errors.js
@@ -307,9 +307,8 @@
 
             // if network is offline, use offline dialog workflow
             if (exception.online === false) {
-                console.log("In offline exception condition");
                 offlineErrorTemplate(exception);
-                return null;
+                return;
             }
             $rootScope.error = true;    // used to hide spinner in conjunction with a css property
 
--- a/common/errors.js
+++ b/common/errors.js
@@ -306,12 +306,10 @@
                 message = exception.message,
                 errorStatus;
 
-            // if network is offline, use offline dialog workflow
-            if (exception.online === false) {
-                offlineErrorTemplate(exception);
-                return;
-            }
             $rootScope.error = true;    // used to hide spinner in conjunction with a css property
+
+            // if network is offline, use offline dialog workflow
+            if (!$window.navigator.onLine) return offlineModalTemplate(exception, "Please check that you are still connected to your network. ", null, true);
 
             // don't throw an angular error if in search/viewer or one has already been thrown
             if (exceptionFlag || window.location.pathname.indexOf('/search/') != -1 || window.location.pathname.indexOf('/viewer/') != -1) return;
@@ -372,6 +370,7 @@
 
 })();
 
+/** Functions to execute outside of the scope of angular **/
 
 /**
  * Log the error in ermrest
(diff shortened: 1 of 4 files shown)


common/modal.js
```diff
@@ -16,6 +16,11 @@
                 if (rejectCB) {
                     rejectCB(response);
                 } else if (typeof response !== "string") {
+                    // there is no internet connection, so templates couldn't be fetched
+                    // set a flag on the response (error) object
+                    if (response.message.includes("Error: [$compile:tpload]")) {
```
RFSH: This could be part of the "catchAll" function. That way we're handling this error type the same way for all the apps. I know the issue was mainly about recordedit, but since by just changing where we're doing this logic we can handle it for all the other apps too, I think we should.

common/recordCreate.js
```diff
@@ -122,7 +122,10 @@
                         // happens with an error with code 0 (Timeout Error)
                         $log.warn(exception);
                         var message = exception.message || messageMap.errorMessageMissing;
-                        AlertsService.addAlert(message, 'error');
+                        // if online, we don't know how to handle the error
```
RFSH: What's the purpose of this?

common/errors.js
```diff
@@ -436,5 +440,44 @@ window.onerror = function() {
...
+
+var errorVisible = false;
+var offlineErrorTemplate = function (error) {
+    var errorVisible = document.getElementById('divErrorModal');
+    if (!document || !document.body || errorVisible) return;
+
+    var errName = error.constructor.name;
+    errName = (errName.toLowerCase() !== 'error') ? errName : "Terminal Error";
+
+    var html  = '<div modal-render="true" tabindex="-1" role="dialog" class="modal fade in" index="0" animate="animate" modal-animation="true" style="z-index: 1050; display: block;">'
```
RFSH: This is almost identical to the one above it. We could create a function that would generate this for us given a set of parameters (to avoid duplicating it).

common/modal.js
jrchudy: This is in `modal.js` as part of the function we use for showing all modals. I figured we only wanted to change the error flow for this case. I'm not sure what other cases would cause `$compile:tpload` error to be thrown, so I didn't want to fully generalize it.

common/recordCreate.js
jrchudy: I made changes that look at the connection state of the browser to decide what the current error condition is. This  was the Alert that was being shown for the current error. Hongsuda wanted to handle this the same in all apps, so we have a modal dialog popup being shown (and this Alert is being suppressed).

common/modal.js
RFSH: Makes sense.

common/modal.js
```diff
@@ -16,6 +16,11 @@
...
+                        response.online = $window.navigator.onLine;
```
RFSH: You're only adding this attribute when we cannot fetch the template. So instead of using `online` (which someone might think that it's always going to be true), we should use the inverse of it (`offline`) and set it to true in here.

jrchudy:
@RFSH  I refactored the changes as you suggested and generalized the error cases. This is ready for review again.

PR #1783, opened 2019-05-21
Conflict link

jrchudy:
When a unique constraint violation is triggered on create or edit, the alert notification will include a link to the record that already exists with the unique key information that was violated.

This can be tested from [this link](https://dev.rebuildingakidney.org/~jchudy/chaise/record/#2/Antibody:Antibodies/RID=16-W24J) by clicking Copy, then submit once `recordedit` loads.


jrchudy:
This Pr is ready for review again after making changes to `ermrestJS` and merging master into this branch and resolving merge conflicts.

test/e2e/specs/default-config/recordedit/submission-disabled.spec.js
```diff
@@ -5,15 +5,16 @@ var testParams = {
         name: "id",
```
RFSH: In the `errors.spec.js`, it seems like we had planed to test this duplicate case and there's a `uniqueConstraint` message defined for it, which is not used at all.  We also could completely move this spec to the errors spec so that all the error-related test cases are in one place.

These are just observations. I don't think you should do these changes now and these can be done as part of refactoring for other features.

jrchudy: I noticed the same thing but didn't want to include it in the error spec because it was a different "type" of error. This is an alert notification rather than a popup error, so I wanted to keep those tests together for `recordedit` but separate from general error popup errors.

PR #1785, opened 2019-05-24
attach onclick handler to download buttons

jrchudy:
Checks the head request first for the asset then fetches the asset/file if user has the proper permissions to. The changes for this are done, I'm not sure what I can do to test this. I'll have to think about it some more.

Links I used during development:
 - [Download button in main entity section (record)](https://dev.rebuildingakidney.org/~jchudy/chaise/record/#2/Gene_Expression:Image/RID=16-2QEC)
 - [Download button in related table section (record)](https://dev.rebuildingakidney.org/~jchudy/chaise/record/#2/Gene_Expression:Specimen/RID=16-2QE8)
 - [Download button in main entity section (record) that returns 401 if not logged in](https://dev.rebuildingakidney.org/~jchudy/chaise/record/#2/RNASeq:File/RID=16-1YDJ)
 - [Download button in main table (recordset)](https://dev.rebuildingakidney.org/~jchudy/chaise/recordset/#2/Gene_Expression:Image@sort(RID))
 - [Download button in popup modal table (recordedit)](https://dev.rebuildingakidney.org/~jchudy/chaise/recordedit/#2/Gene_Expression:Specimen)
    - click on the `Representative Image` foreign key input

I also made sure this feature works in `recordset` when the user changes the search criteria and new values are fetched from the database and presented to the user. This also applies to modal dialogs with the recordset directive.


common/table.js
```diff
@@ -806,6 +806,36 @@
                 _callonSelectedRowsChanged(scope, [tuple], isSelected);
             };
 
+            // NOTE: For some reason this is being registered/triggered twice when the download widget is part of the record table (main body of record)
```
RFSH: This code should be somewhere else. I know we're including this file in all our apps but it's supposed to be for recordset and recordset table.

common/table.js
```diff
@@ -806,6 +806,36 @@
...
+                        e.target.innerHTML += spinnerHTML;
+
+                        var dcctx = ConfigUtils.getContextJSON();
+                        // make a HEAD request to check if the user can fetch the file
+                        dcctx.server.http.head(e.target.href).then(function (response) {
+                            // fetch the file for the user
+                            $window.open(e.target.href, '_blank');
+                        }).catch(function (exception) {
+                            // If an error occurs while a user is trying to download the file, allow them to dismiss the dialog
+                            ErrorService.handleException(exception, true);
```
RFSH: You should call `ERMrest.responseToError` here.

RFSH: Also please make sure to test the 404 case too. I'm not sure whether the error message that we're showing makes sense in this case or not.

jrchudy:
@RFSH  Changes have been made to make this feature more complete and uses the class `deriva-url-validate` instead of `download`. I also updated the main body text above with more info about how it was tested.

Where do you think I should add info to the docs about the use of this new class?

common/utils.js
```diff
@@ -1339,6 +1339,33 @@
...
+
+                    var spinnerHTML = ' <span class="glyphicon glyphicon-refresh glyphicon-refresh-animate"></span>';
+                    //show spinner
+                    e.target.innerHTML += spinnerHTML;
+
+                    var dcctx = ConfigUtils.getContextJSON();
+                    // make a HEAD request to check if the user can fetch the file
+                    dcctx.server.http.head(e.target.href).then(function (response) {
+                        // fetch the file for the user
+                        $window.open(e.target.href, '_blank');
```
RFSH: This will open the file in a new tab which is not the same as the default behavior. You could do a simple `$window.location = e.target.href`.

Although this means that we're not going to trigger the browser's default download behavior. I don't know if we can or whether this can cause issues.

RFSH: We might want to look into whether we can manually trigger the same event that the download does.

common/utils.js
```diff
@@ -1339,6 +1339,33 @@
...
+                    }).catch(function (exception) {
+                        // If an error occurs while a user is trying to download the file, allow them to dismiss the dialog
+                        ErrorService.handleException(ERMrest.responseToError(exception), true);
+                    }).finally(function () {
+                        // remove the spinner
+                        e.target.innerHTML = e.target.innerHTML.slice(0, e.target.innerHTML.indexOf(spinnerHTML));
```
RFSH: It seems like in case of 401 this "finally" won't even run. We might want to remove this spinner in the catch block above. Was there a reason for doing this here and not in the catch-block before calling the `handleException`?

common/utils.js
jrchudy: The way this change behaves is much more similar to the browser's default behavior.

common/utils.js
jrchudy: This finally is triggering when I test this with an asset/file that returns a 404. Do you know of any assets that have a permission on them requiring a user to be logged in?

RFSH: https://dev.rebuildingakidney.org/chaise/record/#2/RNASeq:File/RID=16-1YDJ

RFSH: But even if this didn't have any issues, why finally? Why shouldn't we just hide the spinner in the catch block?

common/utils.js
```diff
@@ -1339,6 +1339,33 @@
             }
         }
 
+        function overrideDownloadClickBehavior(containerElem) {
+            var downloadElements = containerElem[0].querySelectorAll(".deriva-url-validate");
+
+            downloadElements.forEach(function (el) {
```
RFSH: I tested this on IE11. It's throwing an error here. It seems like `forEach` is not defined for the returned value of `querySelectorAll` in IE implementation according to this document:
https://developer.mozilla.org/en-US/docs/Web/API/NodeList

You could either change it to simple for loop, or use the suggested method in that document which is the following:

```javascript
Array.prototype.forEach.call(downloadElements, function (el) {
  // the function goes here
});
```

common/utils.js
RFSH: much more as compared to which other solution? The browser's default behavior is not opening a new tab. I noticed this because after a few downloads my browser disallowed opening new tabs and asked for my permission.

common/utils.js
RFSH: You should also test this on Firefox. It's not registering the click handler in the recordset app. It seems like it's calling this function before rendering the table rows. It works fine in record page.

common/utils.js
jrchudy: If the request is successful (the user is authorized) we still need to hide the spinner, I didn't want to duplicate logic.

common/utils.js
RFSH: About the `forEach`, you actually don't need it. you can just attach the event handler on the result of querySelectorAll.

```javascript
angular.element(downloadElements).on('click', function (e) {

});
```

RFSH:
By using [event delegation](http://learn.jquery.com/events/event-delegation/) we can do this without adding new watchers. You don't need to call this function in the watchers that are specific to record and recordset. Instead, you can call this in the setup step that we have in our apps. I tested this by calling this function in the `setupHead` function that we have (I think that's the most universal DOM manipulation function that we have).  So it would look like the following:

```javascript
function setupHead() {
    addCustomCSS();
    addTitle();
    setWindowName();
    overrideDownloadClickBehavior();
}

function overrideDownloadClickBehavior() {
    angular.element(document).on('click', '.deriva-url-validate', function (e) {
     // the function that you have
    });
}
```

common/utils.js
jrchudy: I meant to say that your suggested change is more similar to what the browser does by default (if not that same), so I'm going to update that.

Issue #1461, opened 2018-03-13
Unauthorized Assets


RFSH:
In ermrestjs, if the asset column has a value, we will turn that into a download button, so when users click on it, the browser download feature gets triggered (instead of just returning a link).

This can cause problems in the case that the hatrac url is not valid or user does not have access to it. Different browsers are handling this differently.  For instance Firefox downloads the page but the downloaded HTML just says "not found". We might be able to improve the UX in this case.

A simple solution is avoiding the `download` attribute and just returning a link. But the better solution might be sending a request to the given url and find out if it's invalid/not found/unauthorized we can show different type of buttons to signal that to the users.

RFSH:
We had a quick discussion about this. We could modify all the `download` links in chaise to do a`HEAD` request first to see whether user has access to the file or not. So the UX would be

1. User clicks on the download button.
2. Chaise shows a spinner inside the button.
3. Chaise sends a HEAD request to the download url.
  3.1. If the request was successful, Chaise will redirect the page to the url so it can be downloaded.
  3.2. Otherwise we will show a message to the users about the error.
4. Chaise will hide the spinner.

I tested `HEAD` request to assets in hatrac. It will throw 401 if user doesn't have access to the asset. I'm not sure about the other types of assets that are hosted in other websites.

hongsudt:
https://www.rebuildingakidney.org/chaise/recordset/#2/RNASeq:File/*::facets::N4IghgdgJiBcDaoDOB7ArgJwMYFM6JHQBcAjdafEAJQDkBBAZRwEcQAaEAMQEsAbHAPo9+AgCoBPAA6CAZgGsc4kAF0AvhyoBJACIqOWABYpuuJJQCKAWgCaAVmsAGdiCt2AGgFkV65Omx4EUGIyNAoEanomVg5hQQBRAA9pDG4AWxwIIgEtbQF5RW8NHT0QQ2NTSgBGADZLbQ8AdVFnGrrGhpba+obrTra3Jw5W+rcAJj6RgBYJjzdqmbcADgW6BYBhBbiFgHFC30xcOBAAGRwAczAscSE+QRznCBQsiDReXjgiDDQcVTUgA@sort(Replicate_RID::desc::,File_Type,RID)

Here is the list of files that should required login that are not on hatrac. Please check whether HEAD command will throw 401 or 403.

RFSH:
[...]

PR #1787, opened 2019-06-04
skip login dialog for 401 on asset

jrchudy:
This PR is an extension of issue #1461 . This changes the error behavior to show the error dialog instead of the login dialog. This is an extension PR #1785 .

The following is what it looks like when a 401 error is thrown while trying to download an asset:
image

@RFSH  The functionality changes in this PR can be reviewed. I still need to make sure with @hongsudt that the language is what we want.

Once this PR is ready to be merged, I will add the documentation for the new `chaiseConfig` property.

### DIFF
--- a/common/errors.js
+++ b/common/errors.js
@@ -23,6 +23,7 @@
         multipleDataErrorCode : "Multiple Records Found",
         multipleDataMessage : "There are more than 1 record found for the filters provided.",
         facetFilterMissing : "No filtering criteria was specified to identify a specific record.",
+        unauthorizedAssetRetrieval : "You must be an authorized user to download this asset.",
         systemAdminMessage : "An unexpected error has occurred. Try clearing your cache. If you continue to face this issue, please contact the system administrator."
     })
 
@@ -134,6 +135,23 @@
         InvalidInputError.prototype = Object.create(Error.prototype);
         InvalidInputError.prototype.constructor = InvalidInputError;
 
+        function UnauthorizedAssetAccess() {
+            /**
+            * @type {string}
+            * @desc   Error message status; acts as Title text for error dialog
+             */
+            this.status = errorNames.unauthorized;
+
+            /**
+            * @type {string}
+            * @desc   Error message
+             */
+            this.message = errorMessages.unauthorizedAssetRetrieval;
+        }
+
+        UnauthorizedAssetAccess.prototype = Object.create(Error.prototype);
+        UnauthorizedAssetAccess.prototype.constructor = UnauthorizedAssetAccess;
+
         /**
          * CustomError - throw custom error from Apps outside Chaise.
          *
@@ -189,6 +207,7 @@
             noRecordError:noRecordError,
             InvalidInputError: InvalidInputError,
             MalformedUriError: MalformedUriError,
+            UnauthorizedAssetAccess: UnauthorizedAssetAccess,
             CustomError: CustomError
         };
     }])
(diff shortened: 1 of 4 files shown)


common/templates/errorDialog.modal.html
```diff
@@ -3,15 +3,17 @@
     <h2 class="modal-title">{{::ctrl.params.errorStatus}}</h2>
 </div>
 <div class="modal-body">
-  <span ng-bind-html="ctrl.params.message"></span>
-  <span ng-if="ctrl.params.showLogin"><a id="login-link" ng-click="ctrl.login()"> Please login</a> to continue.</span>
+    <span ng-bind-html="ctrl.params.message"></span>
+    <span ng-if="ctrl.params.showLogin"><a id="login-link" ng-click="ctrl.login()"> Please login</a> to continue.</span>
+    <span ng-if="ctrl.showDownloadPolicy">Download policy can be found <a id="download-policy-link" ng-href="{{ctrl.downloadPolicy}}">here</a>.</span>
```
jrchudy: I only added this line, the other changes in this file were indenting from 2 spaces to 4.

common/errors.js
```diff
@@ -293,7 +293,7 @@
         var exceptionFlag = false;
 
         // TODO: implement hierarchies of exceptions in ermrestJS and use that hierarchy to conditionally check for certain exceptions
-        function handleException(exception, isDismissible) {
+        function handleException(exception, isDismissible, skipLoginDialog) {
```
jrchudy: I couldn't think of a better way to handle this except by adding a new flag to the handleException function.

common/utils.js
```diff
@@ -1751,9 +1753,17 @@
                 }).catch(function (exception) {
                     // error/login modal was closed
                     if (typeof exception == 'string') return;
+                    var ermrestError = ERMrest.responseToError(exception);
+                    var skipLoginDialog = false;
+
+                    if (ermrestError instanceof ERMrest.UnauthorizedError) {
```
jrchudy: It seems like instead of using `responseToError` here, I should check what "code" is returned and if it's `401` I should call `new ERMrest.UnauthorizedError(messageMap.unauthorizedErrorCode, messageMap.unauthenticatedAssetRetrieval)`. 

That's what we do in `recordedit.app.js` at least, or rather used to do before we stated we should use `_responseToError`. I guess in that case though, it is returning an `ermrestJS` error and we are sending it back to `ermrestJS` to rewrap it instead of attaching the properties like I did..

RFSH: Instead of changing the existing attributes in the `ERMrest.UnauthorizedError` in the catch block, you can create a chaise specific error class. Call it the same as `UnauthorizedAssetAccess` (or any other name. I used this name because it's 401 and asset related so "unauthorized access" + "asset"). Then you don't need to add the `skipLoginDialog` and chaise can just show that newly added error class.

```javascript
if (ermrestError instanceof ERMrest. UnauthorizedError) {
  ermrestError = new Errors.UnauthorizedAssetAccess();
}

ErrorService.handleException(ermrestError, true);
```

About authenticate vs. authorize:
I know based on our discussion unauthenticated makes more sense. But the error in http is called "unauthorized access" whether we like it or not. And that's the name that we've used for the error class. So I think we should stick with it and call it the same way (`UnauthorizedAssetAccess`). It will be more confusing as you might think that unauthorized is different from unauthenticated (which in this context, in our code, they are the same).

jrchudy: I like this suggestion, and I think it makes sense to handle it this way because it is a `chaise` error, `ermrestJS` wouldn't be trying to do something special for it. And I like that it means not adding another flag to the `handleException` function.

jrchudy:
@RFSH  changes have been made based on your suggestions and what Hongsuda asked for for the text. This can can be reviewed again.

Issue #1788, opened 2019-06-07
Add rid template button for md editor


jrchudy:
This [issue](https://github.com/informatics-isi-edu/ermrestjs/issues/791) in `ermrestJS` adds support for an rid template in markdown. A button should be added to our markdown editors that inputs `[[<rid>]]` into the form.

jrchudy:
The corresponding changes for this issue have been merged into master. A button was added to the editor input. Information was also added to the mdHelp page explaining how this new template can be used. Reassigning to @hongsudt.

PR #1792, opened 2019-06-08
Md rid button

jrchudy:
Added a button to the markdown text editor in `chaise` that will insert the rid template (`[[<RID>]]`) when clicked. If text is highlighted before clicking this button, the highlighted text will be treated as the RID. I also added a couple lines to the mdHelp page detailing this functionality.

This PR resolves issue #1788 .

Also done in this PR was updating the `mdHelp` app so that it also includes the configuration code. The app was actually broken because `dcctx` wasn't being set and any templating in the navbar was not being done.


RFSH:
The code looks good, I'm just not sure about the new button.

- The other buttons in the editor are `btn-default btn-sm btn` (black font, white background. grey background on hover), so I think we should be consistent with this button too.

- I feel like this button should be grouped with link and image, so you most probably have to use another group name (I'm not familiar with how this works).

- I'm not sure if pencil is the best icon here. We're using pencil to indicate "edit" buttons. We might want to just use "RID" instead of icon.

Although all of these comments are about how the button looks and I think @hongsudt should be reviewing that part. For reference this is how the new button looks like (the new button is the forth one from the left)

common/vendor/MarkdownEditor/angular-markdown-editor.js
```diff
@@ -179,25 +179,63 @@ function addNewButtons() {
   //       }]
   // },
   {
-        name: "groupPrev",
-        data: [{
+      name: "groupUtil",
```
jrchudy: Moved this to be next to the inline review button. I think it looks a lot better and probably something no one noticed.

jrchudy:
Changes have been made based on comments from hongsuda. This can be reviewed once more. Not many major changes were made since last time:
 - button is black with grey like other buttons
 - button is part of link/image button group
 - use the text `RID` for this button

image

PR #1796, opened 2019-06-13
Pseudo column in templating in recordset

RFSH:
This PR will add pseudo columns in templating support to recordset. It should be merged alongside the [ermrestjs PR#797](https://github.com/informatics-isi-edu/ermrestjs/pull/797).  The set of changes include:

- Modifying the `table.js` to use the activelist for generating requests for aggregate columns.
- Adding new schema to test a table that has waitfor in its columns.


common/table.js
```diff
@@ -186,6 +203,74 @@
...
+         */
+        function afterReadAggregate(vm, colIndex, values) {
+            var agg = vm.reference.activeList.aggregates[colIndex];
+            var sourceDefinitions = vm.reference.table.sourceDefinitions;
+
+            values.forEach(function (val, valIndex) {
+
+                // update the templateVariables
+                if (agg.objects.length > 0 && Array.isArray(sourceDefinitions.sourceMapping[agg.column.name])) {
+                    // NOTE: not needed
```
jrchudy: It looks like you expect `templateVariables` to be defined at this point. If it isn't, should we throw an error instead?

There's no harm in initializing it here, but if this case happens, it means we have an inconsistency being covered up.

Issue #1798, opened 2019-06-15
add canonical tag for google search


hongsudt:
- should have chaise config to control this tag.

jrchudy:
The corresponding PR has been merged into master. The canonical tag will be injected into the `head` tag of each app if `includeCanonicalTag` is defined in the chaise config for the current catalog or at the global level.

PR #1799, opened 2019-06-18
added the canonical tag

jrchudy:
This PR adds the canonical tag to each app `index.html` as mentioned in issue #1798 . This PR also adds a new `chaiseConfig` property called `includeCanonicalTag`. The canonical tag will only show in the html templates if `includeCanonicalTag=true`.


common/utils.js
```diff
@@ -1724,6 +1737,18 @@
             }
         }
 
+        function addCanonicalTag() {
+            var chaiseConfig = ConfigUtils.getConfigJSON();
+            if (chaiseConfig['includeCanonicalTag'] == true) {
+                var canonicalTag = document.createElement("link");
+                canonicalTag.setAttribute("rel", "canonical");
+                var canonicalURL = $window.location.origin + $window.location.pathname + UriUtils.stripSortAndQueryParams($window.location.hash);
```
RFSH: The window location might be using `?` for the hash portion of the url (remember the changes that we made to the UriUtils to make sure they call the same function that is taking care of this logic).
But why do you need to reconstruct the whole url piece by piece? Why are you not passing the whole `$window.location.href` to the `stripSortAndQueryParams` function that you've wrote?
Then that function should take care of the rest.

jrchudy: I added strip query params after I had done the first part and didn't realize I could just call it on the whole url.

jrchudy:
I made some changes based on your comment. I added a function that checks for the location hash and returns the proper value if `#` was replaced with `?`.

I didn't change what was passed to `stripSortAndQueryParams()` because of the `getHash()` function I mentioned I added. I couldn't reuse `chaiseURItoErmrestURI` because it throws an error when there's no catalog, schema, or table defined. I believe, we need this tag for static pages, as well as the deriva webapps (which may not have catalog, schema, or table defined).

Issue #67, opened 2015-09-26
Spec out vocabularies and add support in Chaise


carlkesselman:
We want to have uniform way of representing vocabulary in Chaise which should include support for description, synonym and relationships. Need to define Chaise behavior for vocabulary tables.

robes:
We should convert this into the UI side of the issue. I created a separate issue for defining the data meta-model.

https://github.com/informatics-isi-edu/ermrest/issues/42

robes:
@hongsudt this should be prioritized for next/upcoming sprint

PR #1800, opened 2019-06-18
support hideNavbar query parameter

jrchudy:
Added the logic to hide the navbar if `hideNavbar=true` is present as a URL query parmaeter. In this case, the navbar won't be shown, but the logic will continue to run.

Opening this PR for review so we can discuss whether this is the best way to go about this or not. While investigating this issue, I wanted to make sure it would work for the webapps, upon testing it for `boolean-search`, I came across [this issue](https://github.com/informatics-isi-edu/deriva-webapps/issues/67).


common/navbar.js
```diff
@@ -32,6 +32,7 @@
     ])
     .directive('navbar', ['ConfigUtils', 'ERMrest', 'UriUtils', '$rootScope', '$window', function(ConfigUtils, ERMrest, UriUtils, $rootScope, $window) {
         var chaiseConfig = ConfigUtils.getConfigJSON();
+        $rootScope.hideNavbar = UriUtils.getQueryParams($window.location.hash).hideNavbar === "true" ? true : false;
```
RFSH: We should avoid these kind of changes. Only the `UriUtils` functions should look at the urls since the `location.hash` might be empty (if url is using `?` instead of `#`). I think it's better if you call the `chaiseURItoErmrestURI` function. Then in that function you can add the `hideNavbar` property. This is just to ensure one singular function is looking at the url. 
We might want to change `chaiseURItoErmrestURI` to cache the returned value, but I don't think that's important.

jrchudy: That's what I wanted to ask you about with this PR, I think `chaiseURItoErmrestURI` should be used, which means changing it to understand more of the query parameters. Should I continue to add return values to that function and treat it as a `chaise location` object?

RFSH: The same way that the function is returning `ppid` and `pcid` that is derived from query parameter, you can add this attribute too. 

About creating a new chaise location prototype, it's up to you. I feel like a proper prototype (custom object) wouldn't add any value for now. It would be useful if in the future we decided to add more functionalities to that returned object. Currently we're just using the returned attributes directly without calling any functions.

jrchudy: From what I ran into in the other PR I am working on, using `chaiseURItoErmrestURI` would break the navbar app because that function throws an error when no catalog, schema, or table is available.

jrchudy:
This PR can be reviewed again. I merged `canonical-tag` into this branch because I thought I was going to need the `getHash` function I added in the other one, but realized I didn't need it after we discussed the nuance about using `?` instead of `#`. 

Some apps (navbar and boolean-search) don't have a hash so it would replace the query string with the hash and not parse the query params properly. It doesn't make sense to hide the navbar in the navbar app, but I did want to make sure the functionality was there if someone wanted to embed a static page in an iframe.

common/utils.js
```diff
@@ -857,11 +857,56 @@
...
+        /**
+         * Given a location href and key, return the query param value that matches that key
+         * @param {String} url - the full url for the current page
+         * @param {String} key - the key of the query parameter you want the value of
+         * @returns {String} the value of that key or null, if that key doesn't exist as a query parameter
+         *
+         * Note: This won't handle the case where the url might be like this:
+         * '?catalog/schema:table/limit=20' where limit is a column name
+         */
+        function getQueryParam(url, key) {
```
RFSH: Did you try this with urls that start with `?`? I feel like this might not behave as you expect since you're splitting them by `&` (and if you have filters with & this will not behave properly). 

Since the function is accepting a "key", you don't need to extract all the query parameters, you just need to look for the `key + "="` in the substring that starts from the `url.lastIndexOf("?")`. And then return anything that comes after the `=` and before `&` if exists.

jrchudy: I agree, it's probably better to just look for the key being present and strip the contents after the `=`. 

About your first point, do you mean if the function were called like this `getQueryParam("www.domain.org/chaise/record/?catalog/schema:table/limit=20", "limit")`?

jrchudy:
The changes to hide the navbar after checking if it's contained within an iframe have been made. By default, the navbar will be hidden inside of an iframe unless the inverse of the property is defined (`hideNavbar=false`). If no query parameter is defined, it is treated as `null` and the default behavior will take place.

Outside of an iframe, the navbar will show by default, and only be hidden if `hideNavbar=true`. 

This can be tested by navigating to a place where we have iframe content, [here](https://dev.rebuildingakidney.org/chaise/record/#2/Gene_Expression:Specimen/RID=N-GXA6) in `record` app and [here](https://dev.rebuildingakidney.org/deriva-webapps/boolean-search/) in `boolean-search`. By changing the `src` or `data` attribute, you can force the iframe to load new content and see how our pages behave when injected into these frames.

I also made sure to test the `hideNavbar` param for all `chaise` apps when it's part of the current browser's URL. Currently, it's not easy to test this for boolean-search itself because of the way the content is loaded, detailed [here in issue #67](https://github.com/informatics-isi-edu/deriva-webapps/issues/67).

common/utils.js
```diff
@@ -385,6 +385,7 @@
             var contextObj = ConfigUtils.getContextJSON();
             pcontext.push("pcid=" + contextObj.cid);
             pcontext.push("ppid=" + contextObj.pid);
+            if (contextObj.hideNavbar != undefined) pcontext.push("hideNavbar=" + contextObj.hideNavbar)
```
RFSH: This is not very important but just to be consistent with record and recordset, you might want to add `hideNavbar` only when its value is `true`.

common/navbar.js
```diff
@@ -32,6 +32,9 @@
     ])
     .directive('navbar', ['ConfigUtils', 'ERMrest', 'UriUtils', '$rootScope', '$window', function(ConfigUtils, ERMrest, UriUtils, $rootScope, $window) {
         var chaiseConfig = ConfigUtils.getConfigJSON();
+        var dcctx = ConfigUtils.getContextJSON();
+
+        $rootScope.hideNavbar = dcctx.hideNavbar;
```
RFSH: This is not important either, but you don't need to attach this to rootScope anymore, you can attach it to the scope in the link function.

PR #1808, opened 2019-06-28
Update jquery to 3.4.1

vipul-21:
Created the PR to check if the test cases are passing.


vipul-21:
Created the PR to check if the test cases are passing.

RFSH:
@vipul-21 Make sure to test all the different apps manually too. And if they work as expected you can change the pipeline to review and ask for review, so we can merge this.

jrchudy:
When running the e2e tests locally, a few of them are failing. In `recordset/presentation.spec.js`, the last test (`"should throw a malformed URI error when no default schema:table is set for a catalog."`)  fails. After investigating this issue further, I noticed that the error was ignoring our normal error flow (going through `ErrorService.handleException`) and instead directly calling `window.onerror`. 

After speaking with @RFSH about it, he mentioned that [jquery 3](https://www.telerik.com/blogs/whats-new-in-jquery-3) changes the way exceptions are managed. From the link:
```jQuery 3 follows the same pattern followed by the native Promise object. Therefore, a thrown exception is translated into a rejection and the failure callback is executed. Once done, the process continues and the subsequent success callbacks are executed.```

This info can be found under the heading `The Deferred Object`.


There is one other failing test case in `recordset/misc-facet.spec.js`, the one with heading `"alert should be displayed upon reaching the URL limit and the request should not be completed."`. I didn't investigate why this test case was failing for me locally.

@vipul-21 please run these two specs locally and see if these same test fail for you. Let me know if you are running into any issues with this.

**Comment by vipul-21** [comment]
Both the test cases are failing:
1) As @jrchudy mentioned the jquery  `jQuery 3 follows the same pattern followed by the native Promise object`, that is the reason behind the failing test case. The exception being thrown to global exception handler is an deferred exception which makes the [jquery error handler](https://github.com/jquery/jquery/blob/master/src/deferred/exceptionHook.js) being called instead of the `angular exception handler`.
To solve the issue, `ErrorService.handleException` was called directly instead of throwing the exception which lead the issue https://github.com/informatics-isi-edu/chaise/issues/1822.

**Comment by RFSH** [comment]
It seems like jQuery3 is already included in chaise: https://github.com/informatics-isi-edu/chaise/blob/master/scripts/vendor/jquery-3.3.1.min.js

So now there are three versions of jQuery in chaise. It makes sense to keep the jQuery1 for the old apps that we don't want tot modify. But for jQuery3 we should only keep one version. You have two options:

- Remove jQuery 3.3.1 and make sure it's not used by chaise or deriva-webapps.
- Remove the jQuery 3.4.1 that you've added and use the existing jQuery 3.3.1 instead.

If these two versions are not very different, I would suggest going with jQuery 3.3.1 specially if it's used in deriva-webapps.

**Comment by RFSH** [comment]
After investigating this, I realized only errors that are thrown in the run block are going through the`jQuery.Deferred.exceptionHook` and therefore by-passing the AngularJS `$exceptionHandler`. For example [this error](https://dev.isrd.isi.edu/chaise/recordset/#1/isa:dataset/*::facets::N4IghgdgJiBcDaoDOB7ArgJwMYFM6JAEsIAjdafIpMEAGhCjABcwkcmB9Acxwh2978AZgGscATxABdAL61Q6JmTQUEVGvUYs2nHnw5bW7DoSgdRE6XJAAlAJIARafSwALFIVxJKARgC0AMwAGgAMALJWskA) (going to recordset with facets that have invalid path in them) is calling the `$exceptionHandler`. But [this error](https://dev.isrd.isi.edu/chaise/recordset/#1) (going to recordset without any table name) is not calling the `$exceptionHandler`.

I couldn't really figure out why this is the behavior, but overriding the `window.onerror` inside the `ErrorService` solved the issue as @jrchudy suggested. This shouldn't cause any side effects and will actually ensure that all of our errors will go through our exception handler

The suggested change: 
```javascript
.factory('ErrorService', [..., function () {

  // The following should be added:
  $window.onerror = function () {
    return handleException(arguments[4]);
  };

  // ... the rest of the function
});
```

So @vipul-21 please add this to the `errors.js` also revert back the change that you made to `recordset.app.js` as that is not needed.

vipul-21:
As per the comment of @RFSH, the jquery is downgraded to `3.4.0` and the suggested code has been added to the `error.js`.

Issue #798, opened 2016-11-23
Buttons vs links in Chaise apps


howdyjessie:
This continues from [issue #782](https://github.com/informatics-isi-edu/chaise/issues/782#issuecomment-262390637), where @robes said:

>One other thing. I believe that karl and hongsuda felt that everything except "Permalink" should use a style that looks more like a button than simply a link. They felt that simply looking like a link didn't give the user the right indication of its affordances.

In the front-end standup today, we discussed whether we should be using buttons or links in our Chaise apps. In Record, we use links (e.g. Add, Edit, Delete, Show All...) while in RecordEdit we use buttons (e.g. Submit Data, Delete) to indicate the similar actions a user can take. 

For the sake of consistency, we need to decide on a unified UX to expose these actions to Chaise users. At the end of the meeting, we decided to punt until next week when everybody will be back in office and can weigh in on this; this issue can track that discussion.

hongsudt:
I am marking this as a wishlist. We all are kind of aware of the consistency issues, and hopefully, we can eventually merge on consistent style.. I don't see us spending a lot of time refactoring all apsp on this.

PR #1810, opened 2019-06-29
System columns heuristic chaise config property

jrchudy:
This PR should be merged in conjunction with the `ermrestJS` [PR #798](https://github.com/informatics-isi-edu/ermrestjs/pull/798).

Changes in chaise were made to register a chaise function with `ermrestJS` so that chaise config values can be consumed within the execution environment of `ermrestJS`. Includes test cases for this feature as well to make sure it applies in different apps (`recordset` and `record`).

Also added a test case for the `hideNavbar` query parameter because I forgot to in my other PR we merged earlier this week.


common/utils.js
```diff
@@ -1551,11 +1552,32 @@
...
+         * @params {String} context - the current app context
+         * @return {boolean|Array|null} boolean - value to show or hide all system columns
+         *                              Array - order of which columns to show
+         *                              null - no value defined for the current context
+         */
+        function systemColumnsMode(context) {
+            var cc = getConfigJSON();
+
+            var mode = null;
+            if (context == 'compact' && cc.SystemColumnsDisplayCompact)  {
```
RFSH: I think this should include all the sub-contexts of `compact` too. So the check should be `context.indexOf("compact") === 0`.

RFSH: I realized now that you've handled this in ermrestjs. But as I mentioned in the other PR, if you're sending a string called context, you should just send the whole context to chaise and let chaise deal with it. You don't need to do a pre-check in ermrestjs.

PR #1819, opened 2019-07-09
Log info added to navbar links

jrchudy:
The navbar links will include the `pcid` and `ppid` if the link is to the same origin. The `ppid` will be set as `navbar` and the `ppid` will be the same as the one generated for the current page.

@RFSH  this can be reviewed, I'm not sure if I should add log info to the other navbar links/buttons (clicking the logo, clicking on the profile dropdown menu and it's subsequent options, or clicking the "return to live" button if looking at versioned data).

### DIFF
--- a/common/navbar.js
+++ b/common/navbar.js
@@ -84,12 +84,17 @@
                         // append "/" if not present
                         if (path[path.length-1] !== "/") path += "/";
                     }
+                    path = $window.location.host + path;
+
+                    // parses the url into a location object
+                    var eleUrl = document.createElement('a');
+                    eleUrl.href = url;
 
                     var isChaise = false;
                     for (var i=0; i<appNames.length; i++) {
                         var name = appNames[i];
                         // path/appName exists in our url
-                        if (url.indexOf(path + name) !== -1) isChaise = true;
+                        if (eleUrl.href.indexOf(path + name) !== -1) isChaise = true;
                     }
 
                     return isChaise;
--- a/common/navbar.js
+++ b/common/navbar.js
@@ -59,7 +59,24 @@
             var parentNewTab = obj.newTab;
             // template the url
             // TODO: This is done here to prevent writing a recursive function (again) in `setConfigJSON()`
-            if (obj.url && isCatalogDefined(catalogId)) obj.url = ERMrest.renderHandlebarsTemplate(obj.url, null, {id: catalogId});
+            if (obj.url && isCatalogDefined(catalogId)) {
+                obj.url = ERMrest.renderHandlebarsTemplate(obj.url, null, {id: catalogId});
+                // only append pcid/ppid if link is to the same origin
+                // same origin urls will be relative, check for http instead(?)
+                // if http, make sure not same origin
+                if (obj.url.indexOf("http") == -1 || obj.url.indexOf($window.location.origin) != -1) {
+                    var paramChar = "";
+                    if (obj.url.lastIndexOf("?") !== -1) {
+                        // already a `?` in the url
+                        paramChar += '&';
+                    } else {
+                        paramChar += '?';
+                    }
+                    // pcid should always be navbar
+                    // ppid should be the pid for the current page
+                    obj.url += paramChar + "pcid=navbar&ppid=" + $window.dcctx.pid;
+                }
+            }
             // If current node has children, set each child's newTab to its own existing newTab or parent's newTab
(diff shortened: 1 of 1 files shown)


common/navbar.js
```diff
@@ -59,7 +59,24 @@
...
+                    var paramChar = "";
+                    if (obj.url.lastIndexOf("?") !== -1) {
+                        // already a `?` in the url
+                        paramChar += '&';
+                    } else {
+                        paramChar += '?';
+                    }
+                    // pcid should always be navbar
+                    // ppid should be the pid for the current page
+                    obj.url += paramChar + "pcid=navbar&ppid=" + $window.dcctx.pid;
```
RFSH: It's better if we stick with `ConfigUtils.getContextJSON()` instead of directly using `$window` object. So if later we decided to change this, we just have to change one function.

common/navbar.js
```diff
@@ -59,7 +59,24 @@
             var parentNewTab = obj.newTab;
             // template the url
             // TODO: This is done here to prevent writing a recursive function (again) in `setConfigJSON()`
-            if (obj.url && isCatalogDefined(catalogId)) obj.url = ERMrest.renderHandlebarsTemplate(obj.url, null, {id: catalogId});
+            if (obj.url && isCatalogDefined(catalogId)) {
+                obj.url = ERMrest.renderHandlebarsTemplate(obj.url, null, {id: catalogId});
+                // only append pcid/ppid if link is to the same origin
+                // same origin urls will be relative, check for http instead(?)
+                // if http, make sure not same origin
+                if (obj.url.indexOf("http") == -1 || obj.url.indexOf($window.location.origin) != -1) {
```
RFSH: We should check for "chaise" urls. We don't want to add these query parameters to static urls and just the origin check is not enough.

jrchudy:
@RFSH  I made changes to the way we check for a chaise url to be similar to the way we check in the `chaiseDeploymentPath` function. I also added this log functionality to the "back to live" button. I think the nature of that button should change to be a link rather than a function, but that can be handled by a different task.

I wasn't sure if we should be doing something similar here for the deriva-webapps. Having a similar array to loop through and check if the path included `deriva-webapps/appname`. We can discuss that in the morning with @hongsudt .

common/navbar.js
```diff
@@ -62,28 +75,30 @@
             if (obj.url && isCatalogDefined(catalogId)) {
                 obj.url = ERMrest.renderHandlebarsTemplate(obj.url, null, {id: catalogId});
 
-                function isChaiseUrl (url) {
-                    function isChaiseApp (url) {
-                        return (url.indexOf("record") != -1 || url.indexOf("recordset") != -1 || url.indexOf("recordedit") != -1);
+                // same origin urls may be relative
+                function isChaiseApp (url) {
```
RFSH: Do you think it would be valuable to move this to `utils.js`?

common/navbar.js
```diff
@@ -59,7 +72,35 @@
...
+                        var path = chaiseConfig.chaiseBasePath;
+                        // append "/" if not present
+                        if (path[path.length-1] !== "/") path += "/";
+                    }
+
+                    var isChaise = false;
+                    for (var i=0; i<appNames.length; i++) {
+                        var name = appNames[i];
+                        // path/appName exists in our url
+                        if (url.indexOf(path + name) !== -1) isChaise = true;
```
RFSH: Is it possible to be more specific for this check? Do we want to add this to any chaise urls? or chaise urls that are in the same origin? 
Assuming we're in `gudmap.org`, should we add these parameters to all of the following? 

- `gudmap.org/chaise/recordset/#1/schema:table`
- `dev.gudmap.org/chaise/recordset/#1/schema:table`
- `rebuildingakidney.org/chaise/recordset/#1/schema:table`
- `gudmap.org/~jchudy/chaise/recordset/#1/schema:table`

jrchudy:
@RFSH  More changes have been made. We decided to check for same origin as well for the urls so that logging makes sense when looking at the request. This way you can track the `ppid` of a navbar link to a proper `pid` for a page on the same server.

The examples for the location object [here](https://developer.mozilla.org/en-US/docs/Web/API/Location#Examples), show that you can create an anchor tag (`<a>`) and attach the current url as an href to parse the url into a location object. That's how I'm comparing the supplied url with the current host + chaisePath (`/chaise/<appname>`).

PR #1825, opened 2019-07-23
Check asset origin

jrchudy:
The following changes were made:

 - added a same origin check before performing the head request to verify the user is authorized to view/download the asset
   - this should not occur anymore for most cases because ermrestJS will check for same origin before applying the `asset-permission` class
   - this class can still be applied via a display-annotation that won't check what origin the user is in when generating the html template for chaise to show
 - if not same origin, redirect modal dialog added
   - user can wait for 5 seconds and it will auto redirect (open the link in another tab to show or download)
   - click "here" to cause the download/redirect to occur right away
 - handle 403 the same way we handle 401
   - state the user is not authorized and point out the user policy
 - fixed bug in `error.js` when message would be undefined

Fixed a bug reported in [rbk-project](https://github.com/informatics-isi-edu/rbk-project/issues/570).

Links for testing:
 - [asset hosted on www](https://dev.gudmap.org/chaise/record/#2/Common:Publication/RID=17-EZ6C)
 - [asset that returns 401 if not logged in](https://dev.rebuildingakidney.org/chaise/record/#2/RNASeq:File/RID=16-1YDJ)

More info related to this PR can be found in the [previous PR](https://github.com/informatics-isi-edu/chaise/pull/1785) for this issue.


common/errors.js
```diff
@@ -146,29 +154,39 @@
...
-            this.errorData.clickActionMessage = messageMap.clickActionMessage.loginOrDismissDialog;
+            this.errorData.clickActionMessage = null;
 
             /**
              * @type {boolean}
              * @desc Set true to dismiss the error modal on clicking the OK button
              */
             this.clickOkToDismiss = true;
+
+            if (code == 401) {
```
RFSH: Isn't it better to create two different error types if these two cases have completely different status, message, and clickActionMessage?

jrchudy:
@RFSH  this has been updated to have 2 separate error classes for 401 and 403 case when trying to access an asset.

PR #1833, opened 2019-08-06
Error dialog on pure and binary add error

jrchudy:
The change for this was simple. If the `isModalUpdate` flag is true, we know we are in `record` app. This might not be the best way to test for this if `isModalUpdate` gets used in other app contexts later. I was thinking I could check the log action but don't want to tie log actions together with other functionality of our app.

I'm not sure if it's possible to test this error because of our disabled rows function.


hongsudt:
@jrchudy @RFSH if this fix is good, please merge it.

PR #1834, opened 2019-08-08
fix submit button in modal pickers

RFSH:
This PR,

- Adds `search-popup` class to all the modal pickers that we have. This was causing inconsistency in UI of the different modal pickers that we have.
- Moves the "submit" button to be beside the "x" button in the header.
- Changes the background color of the modal pickers to be consistent with page headers.
- Removes the secondary scrollbar that we had for the recordset table in modal pickers, to make it consistent with the recordset page.
- Changes the button labels.

What should be done after this: Make sure the scrolling on modals works the same as recordset app (two scrollbars, one for main content and the other for facets).

### DIFF
--- a/common/inputs.js
+++ b/common/inputs.js
@@ -400,6 +400,7 @@
                     modalUtils.showModal({
                         animation: false,
                         controller: "SearchPopupController",
+                        windowClass: "search-popup",
                         controllerAs: "ctrl",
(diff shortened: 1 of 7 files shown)


RFSH:
This PR is not needed anymore. The changes are done in the new style changes that we did.

Issue #1843, opened 2019-09-04
Client event logging


jrchudy:
- [x] Review the logging information in [docs/user-docs/logging.md](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/logging.md#button-action-list). 
- [x] Change chaise/ermrestJS to make HEAD requests to the server to store that log info.

hongsudt:
@RFSH since we went over your document structure, do we still need to keep this issue open?

RFSH:
No, we don't. I am going to close the issues related to log changes.

Issue #1836, opened 2019-08-13
Histogram size changes


jrchudy:
The histogram facet opens properly the first time it is opened. After collapsing the facet and reopening it, it opens to be much larger than it was the first time.

![image](https://user-images.githubusercontent.com/2932600/62982204-53564f00-bde0-11e9-8135-0b0f4cef13f0.png)
![image](https://user-images.githubusercontent.com/2932600/62982216-60733e00-bde0-11e9-8ca3-61538cca7886.png)

jrchudy:
This issue was related to the height of the graph being set to `150px` while the graph also had the property, `autosize: true`. Plotly takes both of these arguments as layout object values:
```
plot = {
  ...
  layout: {
    autosize: true,
    height: 150px,
    ...
  }
  ...
}
```

Removing the set height, and attaching a CSS property to the parent container of the graph resolved this problem. Autosize will try to resize the graph to fill its parent container. This issue will be resolved when PR #1847 is merged.

Issue #1835, opened 2019-08-09
Chaise navbar flashes black


jrchudy:
When a custom CSS document is defined that includes a rule to change the color of the navbar, the navbar flashes its default color first before changing to the custom color.

jrchudy:
This has been fixed as part of PR #1847 .

Issue #1680, opened 2018-11-14
Clear control in all input types


jrchudy:
We want to have a control that allows the users to easily clear the value for all input types. Currently in `recordedit`, there is a clear control for date inputs, timestamp inputs, Asset selector inputs, and Foreign Key selector inputs. In the range inputs, we already have integer/float inputs and date/timestamp inputs with an `x` control to remove the values.

hongsudt:
@jrchudy @RFSH  Should this be part of the style guide?

RFSH:
This has been done as part of #1847.

PR #1847, opened 2019-09-14
UX mitigation

RFSH:
[ermrestjs PR](https://github.com/informatics-isi-edu/ermrestjs/pull/809)

Fixed issues as part of this PR: #1843 #1851 #1680 #1836 #1835 #1849 #1507


jrchudy:
### download buttons:
  - update download documentation for “btn btn-primary”
  - have a separate class for new download styling (default heuristic for generated content by ermrestJS)
    * allows for old deprecated styles to still be used
  - user-doc "style guide" that points out classes that users can apply in MD environment to add styling

### other changes:
 - TOOLTIPS !!! 
 - side bar button bug on record
 - verify spacing for record

jrchudy:
- user-doc "style guide" that points out classes that users can apply in MD environment to add styling

 - toc
    * toc change name to "Sections"
    * move too title above to top-left panel
    * tooltips on right of them  (why tooltip?)
    * change "hide side panel" to "hide panel
  - facet panel
    * hide filter panel button should move left a little bit
    * change "hide filter panel" to "hide panel"
  - sed commands in makefile for version so it's a variable
  - chaise-icons are misaligned because of font size changes
  - *-sm classes for more responsive
  - i icon
    * headers should be smaller and superscript
    * move into : for record page
    * clean up uncurled edge in icon

After the above is done, send url link to ISRD to ask team what they think about the state of chaise with the mitigation changes.

  - apply mitigation styles to recordedit page
    * title update to be consistent with other pages
      ~ create/edit remains up top
      ~ row name is 2nd heading
    * clean up

RFSH:
We have a discussion about these changes and the following are requested changes:

General
- [ ] tooltip placement
- [ ] Export button should be disabled instead of hidden when data is loading

Recordedit
- [ ] The markdown editor height is not set properly on load
- [ ] Boolean null should be "No Value"

Record page
- [ ] Maybe add border to the left of table of contents?
- [x] ~Show empty sections maybe needs to go to the right? Not now.~

Recordset
- [ ] Placeholder for search box should not go to next line.
- [ ] You shouldn’t be able to select show more when match-not-null and then it should have a tooltip why it’s disabled.
- [ ] Facet title spinner goes to next line when shrinked
- [ ] Remove the spinner from searchInput 
- [ ] Fix the responsiveness of main search box (use the space)
- [ ] Bullets in the markdown_container for the recordset are clipped (Gene)
- [ ] Document hide_num_occurrences and sort for facets

Table header (displaying..)
- [ ] Use digit grouping for total number of records (comma separator)
- [ ] the spacing of the cart should be the same as the care in export button (more space
- [ ] The dropdown -> checkbox should be right? (Maybe we should change the background color instead of having checkmark)
- [ ] Disable while the page is loading


Modals
- [ ] Remove the border bottom for the modal-header
- [ ] Change "X" button to "close"

Facet modal
- [ ] The spacing for the top-right-panel is wrong.
- [ ] Squeeze number of occurrences column.
- [ ] Scalar modal should use the smaller modal version
- [ ] Double spinner issue
- [ ] Remove match-null and match-not-null

Add pure and binary modal
- [ ] Submit should be disabled when there's nothing selected
- [ ] The disabled rows link colors should be the same grey as other rows
- [ ] For selected rows: decrease the spacing + remove the border (it should be diff. from filters)
- [ ] Clear all selected rows -> Reset selection
- [ ] Title of the modal should behave the same as recordset page title
- [ ] Reduce the spacing on top of the modal
- [ ] Title should be "Add TABLE to PARENT_TABLE: <small>ROWNAME</small>"
- [ ] Title should be just one line and we should show ellipsis with tooltip

Profile modal
- [ ] Should be the same as the record main content
- [ ] Change order: Displayname, full name, email, id, identities, groups

RFSH:
Newly requested changes:

Record
- [ ] Decrease the padding of main entity table rows (should be the same as the old design which is 5px padding).
- [ ] The "Create" button in modal popups should be using the secondary styles.
- [ ] The logic for turning of the sticky area should be called sooner. The minimum height should be increased.
- [ ] The caret in dropdown for changing the page size is too small.
- [ ] The plus button in recordedit should be redesigned.

RFSH:
Popup:
- [ ] The modal width should be based on the number of columns, later we might want to do annotation, or we might want to make the modal draggable.

Facet popup:
- [ ] Search box in the facet popup should be bigger (bigger min-width?)
- [ ] Modal sizes should be based on screen size?? Currently it’s based on vh and vw, but if you squeeze the whole page it will be small.
- [ ] The modal header must be sticky no matter what… the search and selected rows are not as important
- [ ]  Squeeze space between “X” and selected rows
- [ ]  the selected rows not using the 
- [ ] The spacing between “Reset selection” and the above is not correct.
- [ ] where should we put the x if the selected row is bigger than the line and wraps?
- [ ] selected rows shouldn’t be button group.
- [ ] The value of selected row should white space wrap.
- [ ] When you go from one mode of scroll to the other one, it’s not updating the styles somehow, it’s adding extra scrollbar.
- [ ] Go to protocol, search title for PCR-based -> and open modal -> fix the spacing of that one.

Recordset
- [ ] Change “Edit” button to “Bulk edit"

Recordedit:
- [ ] boolean
    - [ ] Adjust the width of dropdown to be the same as input
    - [ ] Close the dropdown to the input
- [ ] Align “Set value for all records:” vertically in the middle.
- [ ] Change the icon of the removing forms.
- [ ] The “X” icon for closing the “apply all” to be something else.
- [ ] “Add more forms:”
    - [ ] Use the same style as date and time to be like the following: (Qty.)[number](Clone), or “Clone: "
    - [ ] The tooltip should be more informative: duplicate rightmost form the specified number of times.
    - [ ] Add mask to it that you can only use 1-200
- [ ] The horizontal and vertical scrollbar should be in the same container.
- [ ] The upper horizontal scrollbar should be sticky.
- [ ] Decrease the space between lower and the sticky area
- [ ] The YYY-MM-DD and HH:mm:ss 
    - [ ] should use the correct color for placeholder
    - [ ] Should be all capital-case
- [ ] Markdown editor:
    - [ ] The "?" button should be grouped with the other two so we have 4 groups.
    - [ ] Decrease the spacing so we can have only two rows of buttons.

Foreignkey picker
- [ ] The Create button is misspelled
- [ ] Create: Tooltip: Create new “name"
- [ ] The vertical alignment is not properly set (the table is smaller).

Record page
- [ ] Bullets are not disabled for markdown.
- [ ] Move the buttons inside the span and align to the right…?

Navbar:
- [ ] The space between caret and text for the dropdowns that have another menu is not enough.

jrchudy:
- [ ] autoscroll on next/previous click for recordset app
  - [ ] make sure this happens when new filters are selected too

- column headers in tables
  - [ ] sort icon overlaps info icon

- recordedit
  - [ ] white box to left of scroll bar for top sticky scrollbar
  - [ ] focus on add x forms input grays right border

PR #1874, opened 2019-11-14
Fix integer and float input validators

jrchudy:
In this PR, I changed the integer and float inputs to have `type=text`. I learned that `type=number` helps to mask what values can be typed in but causes problems with HTML validation breaking the angular validations we were doing.

Because of the above change to type text, `integer` inputs no longer had proper min/max validators. I added a new pair of validators for min and max called `chaiseMin` and `chaiseMax`. 

Added an expectation to properly check for an error when an invalid integer is typed that includes letters from the alphabet.

One other change included in this PR is to conditionally place the export tooltip above the export button if navbar is visible, and to the left if navbar is not visible.


common/templates/inputs/rangeInputs.html
```diff
@@ -9,22 +9,22 @@
...
-                            <input type="number" class="range-min" min="{{intMin}}" max="{{intMax}}" placeholder="{{absMin}}" ng-model="model.min" integer disable-arrows>
+                            <input type="text" class="range-min" chaise-min="{{intMin}}" chaise-max="{{intMax}}" placeholder="{{absMin}}" ng-model="model.min" integer disable-arrows>
                             <chaise-clear-input btn-class="min-clear" click-callback="model.min=''" show="model.min || (numericMin.$invalid && numericMin.$dirty)"></chaise-clear-input>
                         </div>
                     </div>
                     <!-- int4 -->
                     <div ng-switch-when="int4">
                         <div class="chaise-input-control has-feedback">
-                            <input type="number" class="range-min" min="{{intMin}}" max="{{intMax}}" placeholder="{{absMin}}" ng-model="model.min" integer disable-arrows>
+                            <input type="text" class="range-min" chaise-min="{{intMin}}" chaise-max="{{intMax}}" placeholder="{{absMin}}" ng-model="model.min" integer disable-arrows>
```
RFSH: I think you can also remove the `disable-arrows` directive from here. It seems like it's meant to " disables the actions that HTML attaches to those input types for up/down", so it's not needed anymore.

common/validators.js
```diff
@@ -31,6 +31,38 @@
...
+                        // use the `required` attribute in the HTML
+                        return true;
+                    }
+                    return parseInt(viewValue, 10) <= parseInt(attrs.chaiseMax, 10);
+                };
+            }
+        }
+    })
+
     // TODO: this is not working as expected, 1.1.1 is "valid"
```
RFSH: Does this bug still exist with the changes in this PR? If not, you should update this comment. 

You might want to update the last line that explains how to use it. Although technically that's still correct.

jrchudy: Right, I will remove this comment and update the other one.

PR #1875, opened 2019-11-21
Display search columns as placeholder

RFSH:
This PR is based on [the latest ermrestjs PR](https://github.com/informatics-isi-edu/ermrestjs/pull/822). It will use the newly added API in ermrestjs to show the search columns as placeholder.


common/templates/searchInput.html
```diff
@@ -1,8 +1,15 @@
 <div class="chaise-search-box chaise-input-group" ng-class="{'disabled-element': disabled}">
     <div class="chaise-input-control has-feedback">
         <input type="text" ng-model="searchTerm" ng-class="inputClass" ng-disabled="disabled" on-enter="enterPressed()" ng-change="inputChanged()" ng-attr-autofocus="{{focus === true}}">
-        <span ng-hide="searchTerm" class="chaise-input-placeholder" ng-click="changeFocus()" ng-if="placeholder.isHTML === true" ng-bind-html="'Search ' + placeholder.value"></span>
-        <span ng-hide="searchTerm" class="chaise-input-placeholder" ng-click="changeFocus()" ng-if="placeholder.isHTML === false" ng-bind="'Search '  + placeholder.value"></span>
+        <span ng-hide="searchTerm" class="chaise-input-placeholder" ng-click="changeFocus()" ng-if="!isArray(searchColumns)" ng-bind="'Search' + (searchColumns === false ? ' all columns': '')"></span>
```
jrchudy: Is this suggesting that `searchColumns` could be a single column name? Just curious because the conditional here could be moved down below and conditionally add " all columns" OR  ng-repeat through `searchColumns`.

recordedit/recordEdit.app.js
```diff
@@ -160,7 +160,7 @@
                 // we are using filter to determine app mode, the logic for getting filter
                 // should be in the parser and we should not duplicate it in here
                 // NOTE: we might need to change this line (we're parsing the whole url just for fidinig if there's filter)
-                var location = ERMrest.parse(ermrestUri);
+                var location = reference.location;
```
jrchudy: This is a good change. I don't think `chaise` should have ever been calling `parse` directly.

.travis.yml
```diff
@@ -106,6 +106,7 @@ before_install:
   - cd ..
   - git clone https://github.com/informatics-isi-edu/ermrestjs.git
   - cd ermrestjs
+  - git checkout search-columns
```
jrchudy: Don't forget to undo this change.

common/templates/searchInput.html
RFSH: `searchColumns` could be `false` (if it's used in entity mode and the `search-box` annotation is missing).

PR #1882, opened 2019-12-04
Boolean dropdowns can flow outside of container

jrchudy:
This PR fixed an issue that @ljpearlman found in `recordedit` with boolean dropdowns being clipped when they are the last input field.

I added `dropdown-append-to` for boolean inputs so they are attached to the input container outside of the table (which has `overlow: hidden;`). I added a click event to the dropdown toggle element to recalculate the width of the dropdown each time it's opened because the number of forms can change.

This can be tested here:
https://dev.rebuildingakidney.org/chaise/recordedit/#2/Common:Principal_Investigator

### DIFF
--- a/common/inputs.js
+++ b/common/inputs.js
@@ -488,7 +488,7 @@
                     scope.model.value.time = null;
                 }
 
-                scope.inputContainer = document.querySelector('.input-container');
+                scope.inputContainerForDropdowns = document.querySelector('.input-container');
 
                 // used to increase the width of boolean dropdowns to the size of the input
                 scope.setDropdownWidth = function () {
--- a/common/inputs.js
+++ b/common/inputs.js
@@ -492,10 +492,7 @@
 
                 scope.setDropdownWidth = function () {
                     var inputSelector = scope.columnIndex + '-boolean-input',
-                        dropdownSelector = scope.columnIndex + '-dropdown';
-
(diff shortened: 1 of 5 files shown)


common/inputs.js
```diff
@@ -486,6 +487,18 @@
                 scope.clearTime = function () {
                     scope.model.value.time = null;
                 }
+
+                scope.inputContainer = document.querySelector('.input-container');
+
+                scope.setDropdownWidth = function () {
```
RFSH: For the other case, you're assuming that there will be just one dropdown on the page and therefore you're just using a class. You could make the assumption here and you could. Then you can use a class instead (preferably the same class name to avoid confusion) and remove the `columnIndex` that you've attached to scope just for this change.

Also can you please add some comments to explain why you have to do this and what are the assumptions in this code?

jrchudy: The `columnIndex` has to be used in this case because relying on the class may fail. One of the other hidden boolean inputs can be selected and the width will be unchanged.

RFSH: hmm. so that means class only works for the other case because all the dropdowns are the same size, and the first dropdown on the page is always one of the main dropdown inputs (rather than select-all dropdowns). If all the select-all dropdowns have the same size, you could use a class that is different from other types of dropdowns on the page. right?

PR #1883, opened 2019-12-04
Truncation one more time

jrchudy:
In this PR I refactored the logic for truncation to no longer rely on a watch event that's triggered when data changes. Instead it will trigger the truncation logic when the height of the table row changes. This also handles the case to reset the truncation logic if the content of the table changes from a page size change or more faceting criteria is added.

This is more performant and works more flawlessly for text cases. There are mixed results for images and iframes which can render after the digest cycle has finished, increasing the row height without triggering the sensor (which checks during the next digest cycle which doesn't occur).


common/ellipsis.js
```diff
@@ -206,64 +201,48 @@
...
+                    });
+
+                    // reset overflows because new rows are available
+                    $rootScope.$on('reference-modified', function() {
+                        scope.overflow = [];
+                        scope.hideContent = false;
+                        scope.maxHeightStyle = null;
+                        userClicked = false;
+
+                        $timeout(function () {
```
RFSH: Is this intentional (empty timeout)?

RFSH:
I found an inconsistency in the behavior of this logic that might help us debug this further. 

On load, the ellipsis logic works flawlessly and consistently for text columns. But if you go to the next page, and get back to the first page, it won't trigger the ellipsis logic at all. This could mean that the Angularjs is not calling the `link` function of the ellipsis directive for the newly drawn rows.

jrchudy:
@RFSH  I made changes to the reference modified listener to also check and set the overflows. This appears to address the cases when the user changes the page size, changes a facet, or using next/previous and the overflows don't update automatically.

PR #1888, opened 2020-01-03
Navbar show and enable

jrchudy:
This PR extends the support of the chaise config property `navbarMenu` by allowing another attribute to be defined for each node or leaf that gives further control over showing the option and enabling the click event for the option.

As outlined in the issue, if the option's `acls.show` list includes a globus group or globus user that matches one of the attributes for the current session, it won't be visible in the DOM. When the option can be shown, but the `acls.enable` list doesn't have a match, the option will be shown but greyed out and not clickable. 

In the same case, shown but disabled, if the node spawned a sub menu with more options, the caret will be hidden and the menu will not open.

I wasn't sure about testing this in travis. The tests for this are only being run for local environments currently because I think the globus group hash value used in the travis tests changes each time we recreate that environment.


common/navbar.js
```diff
@@ -140,6 +192,14 @@
                 scope.toggleMenu = toggleMenu;
                 scope.onToggle = onToggle;
 
+                scope.canShow = function (item) {
+                    return canShow(item, dcctx.session);
```
RFSH: I might be wrong, but the only place that I saw that we're attaching `session` to the context object is [here in login.js](https://github.com/informatics-isi-edu/chaise/blob/master/common/login.js#L21), which is inside a callback. So while we're running this `canShow` and `canEnable` functions, the `dcctx.session` might not have the proper value. Is that correct? Shouldn't these functions take care of that?

This is not directly related to your changes, but why do we even need to attach the session to dcctx? Shouldn't any place that needs to look at the session value, call the `Session.subscribeOnChange`?  Shouldn't these two `canShow` and `canEnable` call the `Session.subscribeOnChange` as well??
Also even if we want to have `session` under that object, we should directly attach extra attributes to the result of `getContextJSON`, and instead, we should call an appropriate setter function.

RFSH:
For testing, we currently don't have any automatic test cases in chaise based on user session, so we could also skip adding test cases for this as well.

Although if you want to add these types of test cases, you can start doing the same thing that Chirag did in ermrestjs. In ermrestjs, to resolve the issue that you mentioned, [during the test setup he's getting the user id associated with the cookie](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/utils/jasmine-wrapper.js#L9). 

If we're not adding automatic test cases for this, we need to make sure to have at least one manual test for this (or steps to recreate this manual test scenario).

jrchudy:
Changes based on suggestions have been made. @RFSH this may be reviewed again.

Issue #1879, opened 2019-11-27
Cleanup server logs


RFSH:
We recently added client logs to log actions that are not generating any server-side requests. While doing so, we made some changes to the way we're naming the actions and we decided to add new attributes. 

We should do the same exercise on the server-side as well and review what's been done, based on the new changes. 

Some of the changes that came up and we should think about:

- [x] We might want to attach a specific action to the `preauthn` requests (requests generated when users click on the login button). We might want to be able to distinguish between using login link in navbar, in errors, or login warning. 
- [x] Instead of using `referrer` we should change the child requests (facet or related requests) to log `source` and `main_rid` instead.
- [x] Foreignkey popup should log which foreign key column the request belongs to.

RFSH:
We have a long discussion about the changes that we should make to capture richer information about the steps user take without the need to look at multiple requests. This is the proposed solution:

For each server requests, alongside `pid`, `cid`, `wid`, `action`, and `catalog`, we should have 

- `schema_table`: To capture the projected (leaf) table in the request.
- `stack`: An array of objects that each can have the following attributes:

```javascript
{
  "type": "",// the type of the node request (m: main, re: related, f: facet, fk: foreignkey, pc: pseudo-column)
  "s_t": "", // the end table of this node
  "filters": {}, // the facet object
  // only applicable in the first node:
  "cfacet": {},
  "custom_filters": "", // the filters that cannot be translated into facet syntax
  // not applicable for the first node:
  "src": [], // the source path
  "agg": "" // agg function
}
```

Based on this, the main request for a record page would look like the following:

```javascript
{
  "pid": "",
  "cid": "",
  "wid": "",
  "schema_table": "Gene_Expression:Specimen",
  "catalog": "2",
  "action": "",
  "stack": [
    {
      "type": "m",
      "s_t": "Gene_Expression:Specimen",
      "facet": {
        "and": [{"src": "RID", "c": "16-WHTP"}]
      }
    },
  ]
}
```



And the following is an extreme example to show how the whole object should look like. This is the request that was fired for an aggregate column, that is in a facet popup of an add pure and binary popup. So we are capturing the main entity, related entity, facet popup, and the aggregate column:

```javascript
{
  "pid": "",
  "cid": "",
  "wid": "",
  "schema_table": "Common:File",
  "catalog": "2",
  "action": "",
  "stack": [
    {
      "type": "m",
      "s_t": "Gene_Expression:Specimen",
      "facet": {
[...]

PR #1889, opened 2020-01-17
Log changes

RFSH:
These changes are part of #1879. To summarize this PR,

- Adds `stack`  information to all the requests and modifies the action string to include the `stack-path`.
- Merges all the update requests in recordset and instead logs that information in the `causes` array. Having different action strings for updates was also wrong since it would only capture the last user action and not the changes that they might have made along the way. Also the `stack` includes the `start_ms` and `end_ms`. The `start_ms` is the elapsed time of when the page became dirty, and `end_ms` is when we are sending this request. This will allow us to join the logs by `start_ms` and then the request with the highest `current_ms` would be the request that users are seeing. While the result of the rest of the requests are ignored by flow-control (same attributes was added to any update request throughout chaise).
- Introduces more client logs: facet open, facet close, facet breadcrumb modifications, selected breadcrumb modifications, sort, page-size changed, page next, page previous, edit intend (in tables), add intend (in modals and related entities), unlink intend, unlink cancel.
- Modifies the authentication requests to send different action strings depending on where the user logged in from (alert, error modal, navbar).
- Modifies navbar openMenu api to only log the opening event instead of every button click.
- Adds a new `names` attribute to navbar logs that captures the name of the menu options that the user had to click to end up with the current option (each option now has information of all its parent and is being logged).
- Relocates the `logActions` and everything related to logs to `logService`.
- Adds new APIs to `logService` to support the newly added features. We could obviously refactor this further. For example, I didn't need to pass `action` and `stack` all the time and instead I could write a function that would do it for you. 

Other non-log related changes are:

- Changed the behavior of the edit button in association tables (#1889).
- Fixed the usage of `.server` in chaise to avoid terminal error (#1895).
- Fixed the recordedit bug that wouldn't call the updated function, and therefore record page was not being updated after editing a related entity.
- Moved `cid`, `pid`, and `wid`  to their own namespace (`dcctx.contextHeaderParams`) and introduce a new API for this in `ConfigUtils`.

The test cases are passing locally for me.


RFSH:
I added more changes and the PR is almost ready to be merged. The following are the changes that we might want to include as part of this PR but needs more discussion:

- Consider the case that the user is in record page, they click on “add” for multiple related tables, and might click on “edit” in the displayed rows of some related tables as well. Now if they end up doing some of these “add” and “edit”s, when they focus on the record page again, we will update the page. But what should be the cause in this case? If there was just one “add” or “edit”, we could simply use that action as the cause, but in this case we’re updating the page because of multiple changes.
We could use related-create  (entity-create for the table that was changed) and related-edit  (entity-edit) in the cases that there was just one action, and use related-update (or something else) in the cases that there are both edit and add.

- When users click on "add" for a related table, we're adding some information about which foreign keys we can prefill. Then in the recordedit, we will use this information to prefill those foreign keys. The first issue is that we're not sending the foreign key "source" information, so in order to log those we have to send them through cookie. Also, there might be multiple foreignkeys, so instead of `source` in the stack element, we have to add `sources`. For these reasons, is this even necessary to log the foreign key information? We're already logging the client action of pressing the "add" button. Besides logging the foreignkey information is just logging the internal logic of chaise.

- The server requests described in the issue: https://github.com/informatics-isi-edu/chaise/issues/1879#issuecomment-575403465


Apart from this, I have to add make sure the test cases are passing locally and in traivs. Also documentation! I have no idea how we want to approach documenting this and I think we should discuss it in a meeting.

RFSH:
I finished the remaining tasks that I explained in the previous comment as well as fixing the test cases. This PR is ready to be reviewed and I only need to add documentation to it. I'm going to update the first comment to reflect all these changes.

RFSH:
I added more comments, refactored a bit, and added documentation.

For documentation, I moved the old doc to `logging-pre-feb-20.md` and modified the `logging.md` to have the latest information. At first, I wanted to list all the requests in the markdown but I realized that would not be helpful at all. Instead, I added a link to google sheet alongside some explanations and an example. I think this amount of documentation should be enough. If we add more, it will be more confusing as to where they should be looking for information. It will also be difficult for us to keep them in sync with each other.

PR #1891, opened 2020-01-23
fix range picker bug related to clearing facets

RFSH:
The function that was responsible for syncing the checkboxes was not properly working. It had two flaws:

-  was assuming that the not-null choice is always available;
- and also wouldn't clear the not-null filter if it was selected. 

I modified the function to properly take care of the not-null and the rest of the options.

### DIFF
--- a/common/faceting.js
+++ b/common/faceting.js
@@ -490,18 +490,28 @@
 
                     // some of the facets might have been cleared, this function will unselect those
                     scope.syncSelected = function () {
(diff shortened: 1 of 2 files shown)


RFSH:
I ran Travis test cases multiple times, and I couldn't get it to pass. I saw some 504gateways, and some builds just ran the time limit. The test cases are passing locally for me, and I didn't notice any slow down. I also didn't change that much of a test case to cause a slowdown. So I'm going to merge this PR.

PR #1898, opened 2020-02-11
Truncation one more time

jrchudy:
Overhauled the truncation logic to be much more specific to when it is triggered. Each row will have a resize sensor added that triggers when the height changes instead of when new row values are present. This causes the logic to trigger less often and be more performant.

Made the "update" event trigger on `reference-modified` as well (instead of when rowValues changes) for the same reasons as above.

To handle the specific case with aggregates that we saw most of the problems with, after aggregates data is loaded, an event is emitted that adds a resize sensor to the specific cell in the table.


common/table.js
```diff
@@ -310,6 +310,12 @@
                         vm.pendingRowValues[valIndex][obj.index] = displayValue;
                     } else {
                         vm.rowValues[valIndex][obj.index] = displayValue;
+                        // emit aggregates loaded event for [row][column]
+                        var uniqueIndex = valIndex;
+                        if (vm.config.containerIndex) {
+                            uniqueIndex = vm.config.containerIndex + "-" + uniqueIndex;
+                        }
+                        $rootScope.$emit("aggLoaded-" + uniqueIndex, obj.index);
```
RFSH: This is just nitpicky but to be consistent with the rest of the event names let's use `agg-loaded` (or `aggergate-loaded` to be more clear) instead of camel casing the name.

You should also document this event [in this jsdoc](https://github.com/informatics-isi-edu/chaise/blob/802b60b99f85d2b7abb831ff44530d3a798438aa/common/table.js#L99) (where the rest of events are documented. )

common/ellipsis.js
```diff
@@ -217,64 +213,83 @@
...
+                        var uniqueIndex = scope.rowIndex;
+                        if (scope.config.containerIndex) {
+                            uniqueIndex = scope.config.containerIndex + "-" + uniqueIndex;
+                        }
+                        $rootScope.$on('aggLoaded-' + uniqueIndex, function(events, data) {
+                            var columnModelIndex = data;
+                            // +1 to account for the actions column
+                            var columnUiIndex = columnModelIndex + 1;
+                            // var hasPostLoadClass = scope.rowValues[columnModelIndex] && scope.rowValues[columnModelIndex].value.indexOf('chaise-post-load') > -1;
+                            var hasPostLoadClass =scope.rowValues[columnModelIndex].value.indexOf('chaise-post-load') > -1;
```
RFSH: Just to do extra check you should add `scope.rowValues[columnModelIndex].isHTML` to this.

RFSH:
You should also document the class name that you've used [in the markdown documentation](https://github.com/informatics-isi-edu/ermrestjs/blob/master/docs/user-docs/markdown-formatting.md#special-classes) to warn data-modelers where this class name is used. You should also mention that we're looking for the string, so it shouldn't even be part of their values. And I think because of that we should change this class and at least add a `$` before it.

common/ellipsis.js
```diff
@@ -217,64 +213,83 @@
...
+                        } else if (dimensions.height < (maxHeight + 10)) {
+                            scope.overflow = [];
                         }
-                    };
-
+                    });
+
+                    // schemaTable is per table (needs to be done for record app with related tables)
+                    // rowIndex is per row in each table
+                    $timeout(function () {
```
RFSH: Is this `timeout` needed? I don't know if what I'm seeing is related to this, but on load even though the event is broadcasted, the `on` callback that you've added down here is not being called. It's working properly after load though.

common/ellipsis.js
```diff
@@ -217,64 +213,83 @@
...
+                        $rootScope.$on('aggLoaded-' + uniqueIndex, function(events, data) {
+                            var columnModelIndex = data;
+                            // +1 to account for the actions column
+                            var columnUiIndex = columnModelIndex + 1;
+                            // var hasPostLoadClass = scope.rowValues[columnModelIndex] && scope.rowValues[columnModelIndex].value.indexOf('chaise-post-load') > -1;
+                            var hasPostLoadClass =scope.rowValues[columnModelIndex].value.indexOf('chaise-post-load') > -1;
+
+                            if (scope.tableModel.columnModels[columnModelIndex].column.hasAggregate && hasPostLoadClass) {
+                                var aggTD = element[0].children[columnUiIndex];
+                                new ResizeSensor(aggTD, function (dimensions) {
```
RFSH: You can capture the resizeSensor objects that you're creating here, so if the updated value emptied the cell, you can destroy its sensor as well (by calling `.detach()`).

common/ellipsis.js
```diff
@@ -217,64 +213,83 @@
...
-                                updateHeight(i, element[0].children[i].children[0]);
+                    new ResizeSensor(element[0], function (dimensions) {
+                        // if TR.offsetHeight > the calculated maxRecordsetRowHeight
+                        // +10 to account for padding on TD element
+                        if (dimensions.height > (maxHeight + 10)) {
+                            // iterate over each cell (TD), check it's height, and set overflow if necessary
+                            if (!userClicked) {
+                                initializeOverflowLogic();
                             }
+                        } else if (dimensions.height < (maxHeight + 10)) {
```
RFSH: Instead of adding just 10 here, you should define constant variables somewhere so it's easier to find and change them. You should name (or comment) them in such a way that it makes it easier to map it to the css rule that is defining these paddings.

PR #1906, opened 2020-02-22
changed default CSV option

jrchudy:
Rewrote `makeSafeIdAttr` function to properly make class names and id attributes safe so that selectors will continue to work.

Ran the tests after and it all seems to be passing after messing around with the regular expression for far too long.

### DIFF
--- a/docs/user-docs/custom-css.md
+++ b/docs/user-docs/custom-css.md
@@ -52,6 +52,15 @@ With the current HTML structure, it is possible to apply different styles to the
 
 More often than not, you will want to apply styling for each of the columns rather than at the table level, so the `c_<column.name>` identifier should almost always be included in the selector.
 
+### Character replacement
+For both class names and id attributes:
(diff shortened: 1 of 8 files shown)


docs/user-docs/custom-css.md
```diff
@@ -52,6 +52,15 @@ With the current HTML structure, it is possible to apply different styles to the
 
 More often than not, you will want to apply styling for each of the columns rather than at the table level, so the `c_<column.name>` identifier should almost always be included in the selector.
 
+### Character replacement
```
RFSH: There's a section called `Caveats` in this document that is not relevant anymore and you should remove it. 

Also, I think it's better if you change your example to be based on the patterns that are explained in this document. For instance, you can say if your column name is `text (name)`, chaise will rewrite this as `text-name-` and then will use `c_text-name-` as the class name (your example is missing the ending `-` too).

PR #1908, opened 2020-02-26
Working Globus links

jrchudy:
Modified the globus links to work after globus changed their url naming convention from `www.globus.org/app/groups/<ID>/about` to `app.globus.org/groups/<ID>/about`.

Other changes included in this PR:
 - filtered non-globus groups into a different array to show as plain text
 - removed max-height for profile modal to allow scrolling
 - added styles to cite dialog so urls will properly wrap to the next line


common/templates/profile.modal.html
```diff
@@ -20,15 +20,15 @@ <h2 class="modal-title text-center">{{ctrl.user}}</h2>
                     </ul>
                 </td>
             </tr>
-            <tr ng-if="ctrl.groupList">
-                <td rowspan="{{ctrl.groupList.length}}" class="text-left">Groups</td>
+            <tr ng-if="ctrl.globusGroupList">
```
RFSH: This is not handling the case that we have only `otherGroups`.

common/templates/profile.modal.html
```diff
@@ -20,15 +20,15 @@ <h2 class="modal-title text-center">{{ctrl.user}}</h2>
...
+                <td rowspan="{{ctrl.globusGroupList.length}}" class="text-left">Groups</td>
```
RFSH: What is this rowspan used for? do we need it? And if we do, should this include the `otherGroups` as well?

RFSH: I don't think we need to define rowspan here. It has no effect.

common/modal.js
```diff
@@ -347,11 +348,14 @@
             vm.identities.push(session.client.identities[i]);
         }
 
-
         for(var i = 0; i<session.attributes.length; i++){
-            if(session.attributes[i].display_name && session.attributes[i].display_name !== user.display_name){
-                session.attributes[i].id = session.attributes[i].id.substring(24);
-                vm.groupList.push(session.attributes[i]);
+            if(session.attributes[i].display_name && session.attributes[i].display_name !== user.display_name && vm.identities.indexOf(session.attributes[i].id) == -1){
+                if (session.attributes[i].id.indexOf("globus") > -1) {
```
RFSH: Since the next line is looking for the exact position of "24",  you could change this check to look for the whole prefix instead so

```javascript
session.attributes[i].id.indexOf("https://auth.globus.org/") === 0
```

PR #1913, opened 2020-03-03
Pseudo-column in templating in record page

RFSH:
This PR will add pseudo-column in templating support in record page which means visible-columns in `detailed`, visible-foreign-keys, and citation annotations. The summary of changes are as follows:

- Changed the flow-control logic in record page. Now it's ermrestjs that is dictating the order of secondary requests and chaise just uses the `Reference.activeList.requests` to get the secondary requests that it needs. Based on the type of request, chaise will call the nesseccary function to get the values. 

- Changed how the share popup button works and how we're showing the citation. The citation will be disabled while the main entity is loading. After that, users can click on the "share and cite" button to open the share popup. But the value of citation might not be available yet. To signal this, I added a `$rootScope.citationReady` boolean, which the modal can use to show a loaded. The modal also watches for this boolean, to show the citation as soon as it's available.

- Modified the logic of showing/hiding columns/inline tables since now the `Page.content` might be waiting for extra requests. So I had to add an extra `pageContentInitialized` boolean to signal when we can look at the value of `pageContent`.

- Refactored the `table.js` functions responsible for getting the values of aggregate to be aligned with the new functions that I added in the `record.utils.js` (this has been done just for readability of code).

- Since I modified the `Reference.activeList` API in ermetsjs, I had to update `table.js` to use the new APIs.

- Added new test cases to recordset presentation, record presentation, and related table specs.

related issue: [ermrestjs#617](https://github.com/informatics-isi-edu/ermrestjs/issues/617)

_The only remaining tasks in chaise are making sure Travis is passing and the code is readable and documented._


common/table.js
```diff
@@ -134,195 +134,193 @@
...
             }
             var logObj = {
                 action: getTableLogAction(vm, action, logService.logStackPaths.PSEUDO_COLUMN),
                 stack: stack
             }
-            pcol.column.getAggregatedValue(vm.page, logObj).then(function (values) {
+            activeListModel.column.getAggregatedValue(vm.page, logObj).then(function (values) {
                 if (vm.flowControlObject.counter !== current) {
-                    return defer.resolve(false);
+                    return defer.resolve(false), defer.promise;
```
jrchudy: I didn't know you could do this.

docs/dev-docs/flow-control.md
```diff
@@ -41,19 +45,111 @@ The best way to handle these would be using states and state-management. But in
...
+    ```
+    (es_4), (i_3, es_1, es_2), (i_4), (agg_1), (i_1, agg_2), (i_2, es_3, agg_3), (agg_4)
+    (r_1, agg_4, agg_5), (r_2, es_1, es_5, agg_6, agg_7), (r_3, es_1, es_6), (r4)
+    ```
+
+
+
+3. Send requests for all the aggregate columns that are used in the tabular mode of inline tables from to bottom.
+
+4. Send requests for all the aggregate columns that are used in the tabular mode of inline tables from top to bottom.
```
jrchudy: I think you repeated this line from the one above.

RFSH: I forgot to update the second one to say related tables (instead of inline tables). I'll fix this.

Issue #1918, opened 2020-03-11
Session expire or change while interacting with chaise pages


RFSH:
We're currently working on some improvement for 401 handling. 401 error can happen if the user session expires and they don't have the required access to the data anymore. But in read cases, even if the user session expires, we might not get a 401. For example, assume that you opened a recordset page and didn't interact with it for some time. If later you get back to that page and open a facet modal, the data that you're going to be presented with might be different if your session has been expired.

This not only going to be confusing to the users but also it might also cause 409 errors in ermrestjs/chaise since all the Reference objects (and therefore requests) have been created based on the old user permissions.

The same issue can also happen if the user permissions change while they are interacting with the page, but that's not as frequent as the session expiring and it might be harder to guard against.

We should think about ways to fix this as it can be confusing to the users (or might cause unexpected 409 errors).  One way to fix this is to periodically check the user session in the background and prompt users with the same 401 error if it has been expired.

jrchudy:
In the case that an unexpected 409 error occurs because of a session timeout, I don't think we want to show a "session timed out login dialog". Instead we should do a check to see if the session is timed out, then modify the error dialog to prepend text suggesting that this error may have occurred because a previous session existed but is no longer valid.

Issue #1916, opened 2020-03-10
Session timeout on page with read requests


jrchudy:
This requires further discussion and is related to issue #1838 .

The scenario is that a timeout/logout occurred while a user was looking at a page that can trigger further read requests (`record`/`recordset`) that require certain permissions to retrieve the data.

The user loaded the page with permission to see the content, a timeout (or logout) occurs, then the user tries to facet (`recordset`) or page to the next page of data or change the page size (`record`/`recordset`). The user is then prompted with a login modal:
  1. The same user signs in again
    - we take a similar approach as we are doing for `recordedit` and continue processing the requests that previously caused an error (post login)
  2. A different user signs in, this user does NOT have permission to read the table an action was taken on
     - error modal shown to user with appropriate error message, do not allow dismissal and provide a "reload" and "home" navigation option
     - Alternative (HT): 
       - error message: Your login account xxx is different from the existing account. 
       - Click OK to continue with account xxx. Continue as xxx will refresh the page with new credential. Dismiss will leave them where they were. 
     - @RFSH  suggested that not having permission could also be an issue with a pseudo column which touches a table the user can't read from. Not sure if this changes how this should be handled
  3. A different user signs in, this user does have permission to read the table an action was taken on
     - should this be handled like the above case for a user without permission?
  4. A different user signs in, this user has MORE permissions to read the table an action was taken on
     - This means the user could have more columns visible to them that aren't being shown
     - Same could happen with a user with less privileges, they may not be able to see a specific set of columns shown
   5. The user returns to an anonymous user without logging back in
      - facet popups may have less options visible to them and not be aware of being logged out and why the options seem to be less than usual

The current functionality is to automatically reload the page when a user logs in. This was done to simplify the process and not have to worry about handling it differently.

This is related to issue #1918 .

@karlcz  @hongsudt  could you review this and add comments?

Issue #1915, opened 2020-03-09
Timeout with pure and binary add


jrchudy:
In record app, when a user goes to link more content via pure and binary add popup, if a user gets timed out the selections are lost.

The workflow: 
The user will open the modal popup for add, make a few selections, then walk away from the machine and coming back to a timed out session. This can also occur from the user doing the same above, but then logging out in another tab. Upon clicking the "submit" button, the modal dialog closes and the login modal popup will be shown.

This workflow is counterproductive to the login workflow. We want to preserve the state (selections made) of the pure and binary modal. The login modal should either popup on top of the p&b modal, or the login error/warning needs to be added to the modal as an alert at the very top.

Issue #1838, opened 2019-08-13
Do not reload on timeout


jrchudy:
If the user's session is timed out on recordedit and they log back in (same user), do not reload the page. Check to see if the previous user (reference.session info) is the same as the new user that just logged in to accomplish this. Related to issue #1837 .

jrchudy:
There are 3 cases that can occur with respect to a user being timed out/logged out. The following is how the 3 cases will be handled in `recordedit`:
    1. The same user signs in again, the data submission request is "retried" and the workflow remains the same as previous if errors occur
    2. A different user signs in, this user does NOT have permission to create/edit for that table or specific row of data.
       - notify the user the specific record was being created/edited by someone else (globus name)
       - error dialog popup that can NOT be dismissed
    3. A different user signs in, this user HAS permission to create/edit for that table or specific row of data.
      - the same as detailed above about showing an error dialog and saying who the user was


To further the conversation, how should this be handled in the case of read requests?

I created an issue (#1916) to keep track of the case with read requests.

jrchudy:
Another case that can occur is that the user X (original user) has multiple recordedit pages open at once. The submission in one page will trigger the session timeout login dialog to popup, if the user logs in with a different user (user Y) that CAN submit this data, then blocking them in 1 tab won't keep user Y from submitting requests in the other tabs that were started by user X. 

The UX will show them as user X still in those other tabs, but upon data submission, the user information will be fetched from webauthn, and it will be user Y that has submitted that data creation or update when they shouldn't be able to. This can be mitigated by having an onfocus event that gets the current session from webauthn and does a check to see if the previous user is the same as the current user before allowing the submission to occur.

jrchudy:
Error modal:
![image](https://user-images.githubusercontent.com/2932600/77695658-5a922d80-6f69-11ea-9e01-1dcc90f90949.png)

Log out page instructions:
> Open a new tab to any chaise page or in this tab click the log out button in the top right corner.
>
> If you are using a shared computer, please ensure you have also logged out of any identity providers used during your session.
>
>
>
> This can be accomplished by navigating to Globus to logout - <Log out>.
>
> If another identity provider was used to log into globus, make sure you have also logged out of that provider as well.

PR #1920, opened 2020-03-21
Login after timeout while modifying data

jrchudy:
The changes in this PR are done to improve the workflow of creating and editing data after a timeout or logout occurs. This applies to both create and update in `recordedit` and pure & binary create in `record`.  This is for issues #1838 and #1915 .

For most use cases, the required changes were made to `recordCreate.js`. Both apps use the same function to call the `ermrestJS` libraries to communicate with `ermrest`. 2 major changes were made here:
  - rewrote part of the `addRecords` function to remove the following call `reference[create|update].apply(<scope>, arguments)`
    - when this request errored out, it would ignore our reject callback and catch clause and go directly to the global exception handler.
    - it would then call the catch clause causing 2 pieces of code to execute and causing a race condition of which error handling code was triggered first (catch block vs general exception handler)
  - Modified the way submitting data in a modal behaves. The data is now submitted before the modal is closed so that the modal stays open if a timeout occurs (or other blocking error that isn't set as an alert)
    - I did this by adding a function to call in the submit function of the popup modal
    - That function that is called will handle closing the modal after submission has succeeded

Some other changes made to facilitate this functionality:
  - refactored success/cancel callbacks of session timeout popup dialog since it no longer blindly reloads. The success case was not being called before so in the code, it was calling the reject callback for closing a modal
  - `_getSession()` to the body of the session service so it could be called from within the service
  - store previous session information (used to track if a user was logged in when the page loaded)
  - set the session only if it is not set yet or it was changed to `null` (error case of getSession request)
    - not entirely sure about this
    - later, if we are allowing the user to continue, we should update the session info so the app reflects that state
  - created a new error object in chaise called `DifferentUserConflictError`
  - boolean that is resolved in the `setHTTP401Handler` that will decide whether requests should be retried or not

Other changes that were made:
  - refactored the functions that communicate with the local storage
    - did this to prepare for the changes that were discussed for #1916 and #1918 
  - commented logic path for managing timeouts in the future
    - started this while we were discussing implementing #1916 and #1918 
  - modified the z-index of the navbar menu when an error is thrown
    - as more modals get opened, z-index increases by 10 for each modal
    - we tried to be safe about it before by increasing it by only 1


common/authen.js
```diff
@@ -404,7 +471,20 @@
...
                     Session.getSession("401").then(function(_session) {
-                        defer.resolve();
+                        var prevSession = Session.getPrevSessionValue();
+                        // TODO: limiting this to entry may not be the right condition?
+                        // Not sure if this will trigger on page load, if not I think it's safe to run this in all contexts
+                        // recordset  - read should throw a 409 if there's a permission issue
+                        // record     - same issue with read above
+                        //            - p&b add throws a conflict error in most cases, so won't get sent here either (RBK)
+                        // recordedit - add/update return here in some cases, and in other cases will return a 409 and ignore this case
+                        var differentUser = (ConfigUtils.getContextJSON().appContext.indexOf("entry") != -1 && prevSession && !Session.isSameSessionAsPrevious())
```
RFSH: I don't think you need to check for `entry` context. This error applies to all the 401 error cases (regardless of the app) and as you mentioned even the recordedit case that we were trying to solve doesn't go through this and actually throws a 409.

common/authen.js
```diff
@@ -240,49 +268,78 @@
             logInHelper(loginWindowCb, win, reloadCb, 'popUp', null, logAction);
         };
 
-        var shouldReloadPageAfterLogin = function(newSession) {
-            if (_session === null) return true;
+        var shouldReloadPageAfterLogin = function() {
```
RFSH: Can you please add a comment to explain the cases that we're actually reloading after login? it's a bit confusing right now and you have to follow the code path to understand it.

common/errors.js
```diff
@@ -217,6 +219,52 @@
         ForbiddenAssetAccess.prototype = Object.create(Error.prototype);
         ForbiddenAssetAccess.prototype.constructor = ForbiddenAssetAccess;
 
+        /**
+         * Error class that is used when a user logs , but the page's previous session was another user
+         *
+         * @return {object}        Error Object
+         */
+        function DifferentUserConflictError(sessionInfo, prevSessionInfo) {
```
RFSH: The options for the error based on the drawing that Karl did were "reload" and "continue". Based on my assumption the "continue" button was meant to do the session check again and if the issue resolved, the error should be dismissed automatically to allow users to get back to their old state.

We could reuse the "ok" button that currently is directing users to the home page for this purpose.

(This comment doesn't necessarily belong to this part of the code and I added here because it's related to this new error class)

common/authen.js
```diff
@@ -240,49 +268,78 @@
...
-                    action: logService.getActionString(logService.logActions.SESSION_RETRIEVE, "", "")
+            config.headers[ERMrest.contextHeaderName] = {
+                action: logService.getActionString(logService.logActions.SESSION_RETRIEVE, "", "")
+            }
+
+            // see if there's a token before doing any decision making
+
+            // if there's no webauthnCookie || no cookieFromStorage, login flow
+
+            // if there's a webauthnCookie user was logged in at some point on this machine, check to see if it matches _cookie
```
RFSH: The code is not doing any of the explained comment here, right? Is this comment supposed to explain how you want to implement it later? I so, I don't think it should be part of the code.

jrchudy: From above:

> commented logic path for managing timeouts in the future
started this while we were discussing implementing #1916 and #1918

lib/switchUserAccounts.css
```diff
@@ -0,0 +1,9 @@
+.chaise-body li {
```
RFSH, line 1: Let's add this to the `app.scss` that we have. You can create a different SCSS file and then import it in the `app.scss` but we shouldn't create separate CSS files anymore.

Makefile
```diff
@@ -468,6 +469,32 @@ RECSET_SHARED_CSS_DEPS=$(CSS)/vendor/bootstrap.min.css \
 RECSET_CSS_SOURCE=$(COMMON)/styles/app.css \
 	$(COMMON)/vendor/MarkdownEditor/styles/github-markdown.css
 
+DEFAULT_JS_DEPS=$(JS)/vendor/angular.js \
```
RFSH, line 472: Eventually, when we get to refactor the makefile, most of the JS_DEPS can be changed to this one.  This is a lot cleaner than having multiple sources of JS.

common/utils.js
```diff
@@ -2350,6 +2356,8 @@
             SESSION_VALIDATE: "session" + clientPathActionSeparator + "validate",
             SESSION_RETRIEVE: "session" + clientPathActionSeparator + "retrieve",
 
+            SWITCH_USER_ACCOUNTS_LOGIN: "switch-accounts" + clientPathActionSeparator + "login",
```
RFSH, line 2359: I don't think this is used.

common/styles/scss/_navbar.scss
```diff
@@ -17,7 +17,7 @@ body > .container {
 }
 
 .error-open #mainnav {
-    z-index: 1051;
+    z-index: 1501;
```
RFSH, line 20: Now that you know why you're using this specific value. Can you please comment on the reason for it here as well?

common/modal.js
```diff
@@ -145,6 +146,27 @@
             }
         }
 
+        // NOTE: only used for DifferentUserConflictError
+        if (exception.showContinueBtn) {
```
RFSH, line 150: Why are you not using the same pattern of `expection instanceof DifferentUserConflictError` that we have in the lines above?  (I'm not sure which is better/cleaner and I'm just asking).

common/utils.js
jrchudy: I forgot to remove the `"action"` string places I added. Also after changing this variable name, I didn't go back and update it in the code.

Made the change and pushed it.

common/authen.js
```diff
@@ -307,14 +373,86 @@
...
+                        }
+
+                        var onError = function (err) {
+                            // NOTE: user didn't login, what's the error ?
+                            // same error message as duplicate user except "anon"
+                            handleDiffUser();
+                        }
+
+                        // should be modal login
+                        logInHelper(loginWindowCb, "", onSuccess, 'modal', onError, LOGIN_LOGIN_MODAL);
```
RFSH: `LOGIN_LOGIN_MODAL` should be `logService.logActions.LOGIN_LOGIN_MODAL`. Currently, It's throwing a terminal error because of this.

common/modal.js
jrchudy: I could now. Before my changes included me relying on the else case in the above if/else chain to show the reload button and append the text for it. But since we decided to use different text for reload in this case, that has changed.

common/utils.js
jrchudy: Which means I need to update the logging document with the 2 new actions..

common/authen.js
jrchudy: Addressed this when you mentioned above that the log actin I defined was unused.

.gitignore
```diff
@@ -26,6 +26,7 @@ mdHelp.html
 test/e2e/data_setup/uploaded_files/*.*
 package-lock.json
 entities.json
+lib/switchUserAccounts.html
```
RFSH, line 29: I wish we had a better place for this. If we want to continue this path, then the lib function will be full of .html and .js files.

I feel like `lib` might not be the best place for this, but I really cannot think of a better place for it so maybe it's fine. But can you please create a folder for this `switchUserAccounts.html` (with the same name) and then you can rename the HTML to `index.html`. Then the URL will be just `/lib/switchUserAccounts`

common/modal.js
RFSH: I don't know if what I'm seeing is related to this or not. But in recordedit I'm seeing the "reload" message twice.

Screen Shot 2020-04-06 at 11 43 32 AM

common/utils.js
RFSH: Oh right. You just need to update the google sheet and add a new row after this one:

common/styles/scss/_switch-user-accounts.scss
```diff
@@ -0,0 +1,12 @@
+#switchUserAccounts {
```
RFSH, line 1: Since by default `app.scss` is disabling the `overflow-y` (and therefore disabling scroll), you should add a rule to override this. That's how I did it in mdHelp page before:

https://github.com/informatics-isi-edu/chaise/blob/master/recordedit/mdHelpStyle.min.css#L6

common/modal.js
```diff
@@ -129,6 +130,24 @@
...
+
+            vm.continue = function () {
+                exception.errorData.continueCB($uibModalInstance);
+            }
+
+            vm.switchUserAccounts = function () {
+                // NOTE: should this be relative to this location?
+                // error thrown in record or recordedit means './logout/logoutInstruction.html'
+                //    goes to /chaise/record/logout/logoutInstructions
+                $window.open('../lib/switchUserAccounts.html', '_blank');
```
RFSH, line 149: You should use the `UriUtils.chaiseDeploymentPath()` function

.gitignore
jrchudy: I tried doing this and had issues with makefile.. Not sure exactly the error but some of the paths weren't working properly.

common/modal.js
jrchudy: I'm not sure why this happened for you. Since changing the code to check the exception type, the latter reload message should never be appended because the else case won't be reached.

RFSH: I'm not sure what changed but I cannot reproduce this anymore. I'll mark this as "resolved".

Issue #1932, opened 2020-04-13
buggy behavior related to file picker and clone


RFSH:
One of our users reported a buggy behavior in recordedit that was caused as a combination of picking a file and using the clone feature. 

To recreate this bug,
- Go to a recordedit page that has foreignkey and file picker in create mode.
- Use the foreignkey picker and select a value.
- Use the file picker and select a file.
- Press the "clone" button.
- Use the file picker and select a file for the second form.
- Click the "select-all" button for any of the values. Upon clicking, the file picker values will change and user is not able to select new files anymore.

I am still not sure what the cause of this bug is and currently am investigating this and will update this issue with more information. My best guess is that when users click on "clone", we're not properly copying the value and that is causing a referential bug.

hongsudt:
@ljpearlman @crisaless @robes @bugacov @svoinea So you guys are aware of this bug!

RFSH:
### How to reproduce
I think I found the core of the problem. This bug is not a result of "clone" button and it's just a combination of opening foreign key picker and file picker. So

- Go to recordedit page in muti-edit mode that has foreignkey and file picker (make sure the file picker has data).
- Use a foreignkey picker.
- Click the "select-all" button for any of the columns apart from the file picker. The value of the file picker will be removed.

I didn't update the original issue because there might be some other issues related to "clone" specifically.


### Chaise background
To better understand this, let me first explain the following data structure/functions that chaise is using:

- `rows`: An array of "displayed" values. These are the values that users see in the UI. In most cases, each value is a simple string, but in the case of Files, it's an object that has multiple attributes that uploader uses (and we also display).

- `submissionRows`: An array of "submission" values. These values are hidden from users and are only meant to be used when chaise is talking with ermrestjs. This includes data submission and when users open a search popup (needed for`domain_filter_pattern` annotation).  For these two cases, before sending the data to ermrestjs recordedit is always calling `transformRowValues` to turn a `row` into `submissionRow`. 

-  `transformRowValues`: The function used for transforming `rows` into `submissionRows`. For instance, in this function, we're turning the string representation of a JSON column into actual JSON or turning the objects used for timestamp columns into the string representation. 

[...]

PR #1933, opened 2020-04-14
Refactor recordedit codes related to select all and uploader

RFSH:
This PR was first introduced to fix the #1932 issue but since I was working on recordedit, I did some other refactoring as well. The changes are:

- Modified the select-all implementation regarding asset columns which fixed the #1932  issue. To make the implementation simpler, recordedit was using the simple disabled input for all the "select-all" forms, including assets. But this means that the `rows` value had to be a simple string and couldn't be the object that we store for assets. To solve this issue, recordedit was abusing `submissionRows` to hold the object. But this had an impact on other columns, and it was buggy. The code assumed that `sumbissionRows` is only populated after submit but that's not necessarily the case. For example, in the case of foreignkey pickers, we need `submissionRows` for the `domain_filter_pattern` support. Therefore we shouldn't have used `submissionRows` to hold these values. To fix this, I changed the input from a simple disabled to an actual upload directive that is disabled. A while ago we added disabled support to this directive, so I just had to make sure the boolean states are properly set. After this change, I could simply pass the rows to the upload directive and didn't need the `submissionRows` hack.

- `transformRowValues` function that we have is not what it was intended to be at first. So I updated the function JSDOC to explain its new implementation. We should not use this function on its own and it only should be used in `populateSubmissionRows`.  So I spend some time to try and merge these two functions, but it introduced some other bugs. I didn't want to debug it any further, so I just left comments about how I think we can improve this function later.

- `transformRowValues` is not modifying the asset columns anymore.

- While testing this, I found out that the `copyFormRow` is very slow and realized that it's copying all the row values blindly. It was copying `Upload` objects that we have as part of `hatracObj` attribute of asset values. It was copying all the reference and column attributes inside this object, which I have no idea how it was working because there are circular dependencies in reference and cannot simply be deep-copied. So I changed its implementation to copy or create new objects selectively. This improved performance a lot. From my testing, if your form had two assets, it would take around 10 seconds to copy the form. With the new code, it takes less than a second.

### DIFF
--- a/common/templates/inputs/inputSwitch.html
+++ b/common/templates/inputs/inputSwitch.html
@@ -130,7 +130,7 @@
 
     <!-- File input -->
     <div ng-switch-when="file">
-        <upload reference="::$root.reference" column="::column" placeholder="getDisabledInputValue()" value="model.value" is-required="false"></upload>
+        <upload reference="::$root.reference" column="::column" placeholder="getDisabledInputValue()" value="model.value" is-required="false" is-disabled="columnModel.isDisabled"></upload>
     </div>
 
     <!--JSON input -->
--- a/common/upload.js
+++ b/common/upload.js
@@ -14,7 +14,7 @@
                     values: '=?',
                     value: '=',
                     reference: '=',
-                    isDisabled: "=?",
+                    isDisabled: "=",
                     placeholder: "=",
                     isRequired: "="
                 },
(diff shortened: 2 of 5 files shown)


RFSH:
@jrchudy This PR is ready to be reviewed. I modified the description to reflect all the changes.

Issue #1937, opened 2020-04-17
Incorrect view details and edit links in recordset


RFSH:
How to reproduce this bug:

- Go to a recordset page in a deployment that is not setting `maxRecordsetRowHeight` chaise-config property to `false` (disabled the truncation logic).
- Make a note of the RID value of the first row.
- Select a facet that changes the resultset values.
- The edit and view links of the rows are still based on the old resultset while the rest of the table content has been updated.

The cause of this bug:

When we modified the ellipsis logic, we mistakenly moved all the watchers inside [this check](https://github.com/informatics-isi-edu/chaise/blob/master/common/ellipsis.js#L199). We should at least move the last watcher outside this check. We should investigate this more and see which parts should be outside this check.

jrchudy:
The corresponding PR has been merged, assigning this to @karlcz to verify it has been fixed.

RFSH:
I am going to close this issue due to inactivity. If the issue still persists, we should create a new issue for it.

PR #1938, opened 2020-04-17
run init function regardless of maxRecordsetRowHeight value

RFSH:
This PR addresses the issue #1937.


jrchudy:
Made sure the tests passed on travis for the cases I added. One failed last time which I fixed. Tests ran in travis and only 1 failed in error spec (that should not have failed).

PR #1944, opened 2020-05-01
remove trackby to rerender ellipsis on every data update

RFSH:
### How to reproduce the bug:
- Go to a record page that has a pure and binary table.
- Click on "add" button to open the modal.
- Select a facet, so the number of rows that will be displayed after is less than the current number of rows.
- Chaise will throw a terminal error

(Sorry in advance for the following wall of text. I tried to explain it with as much detail as possible.)

#### `vm.page` vs `vm.rowValues`:

When the data changes, we're first setting the value of `vm.page`, and then later in the promise chain, we will populate the values of `vm.rowValues`.
- `vm.rowValues` is what's displayed in each row of a recordset table.
- `vm.page` is used the `tuple` object that is used in ellipsis directive.

So what's displayed on the page won't change until we have populated the `vm.rowValues`. In the case of a normal recordset page, the logic to populate these two will be done in the same digest cycle and therefore, there's no issue. But in the case of pure and binary, since what we're displaying is a result of two separate queries (one for showing the rows and the other one for disabling some of them), AngularJS will do this in two different digest cycles. 

#### ellipsis directive:
This directive is used to represent a row of data in the recordset table. We're passing `tuple` and `rowValues` to this directive, among other parameters. `tuple` is coming from the `vm.page.tuples` and `rowValues` is each element in the array of `vm.rowValues` (`vm.rowValues` is technically an array of arrays). When ellipsis directive loads, we're using `tuple` to generate the view, edit, and delete links. When the data changes, AngularJS is using the same ellipsis template to show the new rows without actually recomputing the ellipsis. So we have to manually recompute ellipsis logic to generate appropriate action links manually.

### Why this is happening:
On each digest cycle, Angularjs will recompute the templates that need to be updated. As I mentioned, it takes two digest cycles to show the results in add pure and binary modal. So after the first digest cycle since the value of `vm.page` has been changed, the ellipsis directive will use the new value of `vm.tuple`. So At this point, the `tuple` and `rowValues` that the ellipsis directive has are not in sync. But we're firing an event in the same digest cycle to signal ellipsis to update itself. At this point, since `rowValue` has not been updated yet, for the ellipsis rows that are supposed to be hidden (because the returned result has fewer rows), we will update the ellipsis code. But since we updated the `tuple` in this digest cycle, its value is undefined and therefore the terminal error is displayed.

### How I fixed it in this PR:
We were using `reference-modified` to rerender the ellipsis logic manually. This event, in some cases as it was explained above, was called in the middle of two digest cycles.

At first, I changed this logic to the old method that we had, which is watching for the `rowValues` in ellipsis.  This solved the issue, but I realized by removing `track by` I can ensure AngularJS will run the link function for us, and then we don't need to run the ellipsis manually. AngularJS will always run the ellipsis directive based on correct values of `tuple` and `rowValues` after each data update and solves the issue.

Although by doing this, the DOM will keep destroying and creating new table rows each time the data changes. But during my manual testing, I didn't see any noticeable performance issues with this. We could add "track by" the `tuple.uniqueId` value, but I think we still want to recompute the ellipsis because the data might have changed, and therefore the truncation logic might need to be changed.

I created the PR so I can write the explanation somewhere and so we can do more manual testing.

### What's missing from this PR:
e2e testing!

P.S. I think some test cases are broken because of the chaise-config change that recently was done in some specs. I fixed one simple case in recordset presentation, but there are some more that are still broken (I tested with master so I'm sure the changes here didn't cause them). I'll look into them tomorrow.

### DIFF
--- a/common/ellipsis.js
+++ b/common/ellipsis.js
@@ -281,25 +281,6 @@
                         }
                     });
                 }
-
-                // when the row data updates we have to:
-                //  - run the init function to make sure functions for action columns point to proper reference
-                //  - reset the overflow (truncation) logic
-                $rootScope.$on('reference-modified', function() {
-                    if (removeCellSensors) {
-                        removeCellSensors();
-                    }
-
-                    $timeout(function () {
-                        init();
-
-                        userClicked = false;
-                        if (initializeOverflowLogic) {
-                            initializeOverflowLogic();
-                        }
-                    });
-                });
-
             }
         };
     }])
--- a/common/recordCreate.js
+++ b/common/recordCreate.js
@@ -359,14 +359,13 @@
              * Callback to get the list of disabled tuples.
              * This is only applicable in case of adding related entities.
              * @param  {Object} tableModel the table model
+             * @param {Object} page the new page object
              * @param  {Array} requestCauses array of string that indicates why the request is fired
              * @return {Promise} Promise is resolved with a list of disabled rows (array of tuple objects).
              */
-            params.getDisabledTuples = function (tableModel, requestCauses, reloadStartTime) {
+            params.getDisabledTuples = function (tableModel, page, requestCauses, reloadStartTime) {
                 var defer = $q.defer();
-                var page = tableModel.page, pageSize = tableModel.pageLimit;
-
-                var disabledRows = [], index, newStack = tableModel.logStack;
+                var disabledRows = [], index, newStack = tableModel.logStack, pageSize = tableModel.pageLimit;
 
                 var action = logService.logActions.LOAD;
                 if (Array.isArray(requestCauses) && requestCauses.length > 0) {
@@ -387,7 +386,7 @@
                         if (index > -1) disabledRows.push(page.tuples[index]);
                     });
 
-                    defer.resolve(disabledRows);
+                    defer.resolve({disabledRows: disabledRows, page: page});
                 }).catch(function (err) {
                     defer.reject(err);
                 });
--- a/common/table.js
+++ b/common/table.js
@@ -461,9 +461,29 @@
                     }
 
                     $log.debug("counter", current, ": read main successful.");
-                    vm.page = page;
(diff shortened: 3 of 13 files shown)


RFSH:
While adding test cases I realized that we already have test cases for this. In [misc-facet.spec.js]() we're actually making sure facets in foreign key and add pure & binary popups are functional (So only testing search functionality in facet popup is missing). But even in master these test cases are passing despite the terminal error. That's just because the displayed values are fine and therefore the condition of the test is met. I'm not sure what the best way to handle this is. I could add a check to make sure the error modal is not displayed after selecting a facet, but this is very specific to this case and I don't think we should just add a check because of a bug that some part of the code had and might not be reproducible anymore. 

The fact that tests can still pass while the code has thrown a terminal error, shows that we cannot easily rely on our automatic tests in all general cases. I am going to investigate this further and see what I can find. The ideal improvement would be a watcher for error modal that can be applied universally to all our e2e tests. 


That being said, since I already wrote the test case for searching in add pure and binary modal, I didn't remove it and pushed it (I can remove it if you think it's redundant). I also added testing of search feature in the facet modal.

RFSH:
During my manual testings, since `vm.page` and `vm.rowValues` are populated in two different digest cycles, I noticed some stale and invalid data while the `rowValues` loads. Because of my slow network, the request for getting the disabled rows in add pure and binary takes a few seconds and in that few seconds, I could see some stale data. 

So I moved the code that is responsible for populating `vm.page` to the same place that we're populating `vm.rowValues`. This shouldn't have any negative impact on the data load since we have to wait for the values of `vm.rowValues` regardless and populating `vm.page` in the first digest cycle has no benefit.

RFSH:
After discussing this, we decided that we should this PR as soon as we can. There were some test cases that were failing because of the `maxRecordsetRowHeight `chaise-config property that we recently added to one of the specs. So we decided to revert that change for now (@jrchudy made that changed in the latest commit of this PR).

PR #1946, opened 2020-05-07
Update test config

jrchudy:
Modified configuration property in one chaise-config to turn off ellipsis feature to make sure turning it off doesn't break anything else.

Other changes were made to make tests run more consistently. There are 3 test cases that are still causing issues but are inconsistent and very hard to reproduce. Two of them are noted by `NOTE:` comments I left in the code. 

The last case happens randomly when a search popup is opened in recordedit. Sometimes `$rootScope.tuples[0]` is not defined and throws a terminal error.


RFSH:
I might be wrong but I don't think we have a test case that is specifically making sure the truncation logic is working properly. If that is the case, I think we should add one to the specs that are doing truncation. It could be as simple as having a large text and making sure it's showing `...more` (we could test what clicking on `...more` and `...less` does too if we want to completely test this feature).

Do you remember which spec was throwing the random error that you mentioned? Since the error is obviously coming from chaise I think we should look into it. I think it's just an issue in the test that we're not waiting properly for the page to load and we're clicking with script (and not actually clicking on the page). I think when we redesigned the recordedit, we didn't change the condition of making sure the page is loaded. The [recordedit-helpers](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/utils/recordedit-helpers.js#L43) is just waiting for the submit button to show up. But the submit button will be displayed as soon as we have the `reference`. I think we should change the condition to look for the `spinner` instead to be more accurate. But should we even show the submit button while the page is loading?

jrchudy:
> Do you remember which spec was throwing the random error that you mentioned?
[Here's an example](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/delete-prohibited/recordset/misc-facet.spec.js#L798-L819)

> Since the error is obviously coming from chaise I think we should look into it. I think it's just an issue in the test that we're not waiting properly for the page to load and we're clicking with script (and not actually clicking on the page). I think when we redesigned the recordedit, we didn't change the condition of making sure the page is loaded., only issue now
Yeaup. It seems that we aren't waiting for the rest of the page content to be ready before clicking on a search popup button. Most of our RE tests relied on waiting for submit, but that only waits for reference (not the data). The search popup relies on the data row being there, so we need to wait for submit then wait for the spinner. 

I have made this change to `recordeditPageReady`, only issue now is that there are multiple places in tests for recordedit that don't call this function yet.

Issue #1940, opened 2020-04-29
ermrestjs and chaise build assumptions


RFSH:
While working on #1936, I wanted to deploy multiple ermrestjs and chaise versions, so I can test the different variants. But that wasn't possible because of the assumptions that chaise and ermrestjs are making about their relative path. To remove these assumptions, I was thinking about adding some environment variables. 

Before explaining these assumptions, let's take a look at the current list of environment variables that we have :

| variable                 | category | description                                   | default value             |
|--------------------------|----------|-----------------------------------------------|---------------------------|
| `ERMRESTJSDIR`           | build    | where to install ermrestjs                    | `/var/www/html/ermrestjs` |
| `CHAISEDIR`              | build    | where to install chaise                       | `/var/www/html/chaise`    |
| `REMOTE_CHAISE_DIR_PATH` | tests    | if defined, scp chaise-config to the location |                           |
| `CHAISE_BASE_URL`        | tests    | used in e2e tests in chaise                   |                           |
| `SHARDING`               | tests    | used in chaise test suit                      |                           |
| `ERMREST_URL`            | tests    | used in test suit (chaise and ermrestjs)      |                           |
| `AUTH_COOKIE`            | tests    | used in test suit (chaise and ermrestjs)      |                           |
| `RESTRICTED_AUTH_COOKIE` | tests    | used in ermrestjs test suit                   |                           |

 The following are the different places that ermrestjs and chaise are making assumptions about their paths:

-  Upon loading, when ermrestjs tries to dynamically fetch the dependencies using the `../../ermrestjs/` URL prefix. This works because chaise is always two levels deeper than ermrestjs when it's running this code (`/chaise/<app-name>` vs `/ermrestjs/`). This won't work if a static site is more than two levels deep.
    - Solution: We can add `ERMRESTJS_BASE_PATH` that uses the default value of `/ermrestjs/`. I chose this name because a while ago we added `chaiseBasePath` chaise-config property that is used to locate the templates in chaise.  

- Chaise creates the URLs with a `../` prefix for all the assets that are injected. This should be fine if the created URLs are only injected into chaise templates as all the current chaise apps are two levels deep (and therefore one level deeper from the assets fires). But if we introduce a new app that is not two levels deep, it won't work. Also if we decide to use makefile to create the list of assets that can be prefetched in navbar (or deriva web apps), this assumption won't work.
[...]

Issue #1935, opened 2020-04-16
navbar app: not appearing on page


crisaless:
On the FaceBase website, the navbar app is not appearing at all on one particular page - currently on the staging server:

https://staging.facebase.org/facial_norms/search/

From the console in Chrome dev tools I can see that it failed to load common/templates/navbar.html

AND it's looking for it in an unexpected location:

https://staging.facebase.org/facial_norms/common/templates/navbar.html?v=null

At first I thought it was because this is page is two directories down from the root. But there are other pages at a similar level, even in that same directory that are not experiencing this problem - for example:

https://staging.facebase.org/facial_norms/tools/

One thing about this facial norms search page is that is a javascript app - so it may be some JS clashing going on...

jrchudy:
We assume certain app names to extract the deployment path. In this case, `facial-norms/search` is being run through our function, but since it has search in it's path, `facial-norms` is being treated as the chaise root folder. We need to think about the `chaiseDeploymentPath` function when it is used in other projects that use our navbar embedded on pages that might have the same name in the url as our apps.

I think these are the offending lines:
https://github.com/informatics-isi-edu/chaise/blob/master/common/utils.js#L830-L834

RFSH:
@jrchudy One simple fix is moving up the next `else if` case so we first check for that instead of app name. Then we should make sure that chaise-config property is defined for all deployments.

RFSH:
@crisaless 
As part of the changes that I'm working on for #1936, I will improve this logic in the code. But for now, I removed `search` from that function so it works in this case.

Also if your static site has  `record`, `recordset`, `recordedit`, or `login` as part of the URL, it will break because of the same function.  As I mentioned this will be fixed later and this is done just so we can have a functional navbar for now.

crisaless:
I confirmed there are no urls in the facebase static site with those words. 

ok, sounds good!

RFSH:
This has been properly fixed by the latest changes in Chaise. We're not making any assumption about the app names and we're using the environment variables that are used to build the chaise to signal chaise javascript code where the code is actually installed.

Issue #1494, opened 2018-04-13
Cleanup App Dependencies in Makefile


RFSH:
In `Makefile` we're listing all the dependencies that are required for each app, and also the common dependencies. But some of these seem redundant. We should be able to clean up these list and change the way we're creating the dependency list. 

We recently found that one of the apps are not working because of a common dependency not being included. If we clean up the lists and make it more readable we can avoid such incident from happening again.

hongsudt:
I can't find the issue that @karlcz brought up related to tagging the deployment files with versions. So I am just going to add the scope of this one..  Feel free to update my comments.. 

The chaise/ermrestjs deployment code should rename the files to include the file checksum (e.g. ermrest_<checksum>.js) and get referenced properly in any scripts that depend on those files. To do this, we need to: 
- [ ] ....

@RFSH @jrchudy please fill in the detail..

RFSH:
I edited the previous comments, I feel like we need more discussion or at least more information about how we can actually achieve this. I don't think it's feasible to add version number to all of the dependencies in chaise. That can easily go out of control. We should consider merging the dependencies during the build, and then adding version.

jrchudy:
Regarding tagging deployment files with versions, this has been done for most of the chaise assets excluding the vendor files.

PR #1951, opened 2020-05-14
Cleanup Makefile and dependencies

RFSH:
This PR will address #1494, #1935, and #1940 issues. The changes include:

- Remove unused files. Since we're copying almost all the files under chaise directory, I removed the old and unnecessary files. 
  - There are a few more that I would like to remove but wasn't sure if they were used.
  - wiki-images was not used at all. There were two images that I thought might be useful, so I created [this wiki page](https://github.com/informatics-isi-edu/chaise/wiki/Whiteboard-dicussions) for them.
- Modify Makefile to use the new environment variables.
- Add chaiseBasePath to meta tag so it’s consistent and can be used in js files as well.
- Clean up Makefile structure
- Minify chaise dependencies

You can find the ermrestjs PR related this [here](https://github.com/informatics-isi-edu/ermrestjs/pull/841).


RFSH:
@jrchudy I think this PR is in a state that we should discuss the changes. I still need to update some documentation but the code should be ready to review. There is a lot to discuss and I need someone else to look at the new makefile to see if the structure that I used makes sense or not.

RFSH:
After discussing the changes in this PR, @hongsudt suggested that we should try and minimize the number of things that are included in the `chaise-includes.html` file. Most specifically she suggested using other methods instead of the `meta` tags that we've used.

So I modified chaise to do the same thing that ermrestjs is doing. I created a `makefile_variables.js` to include the variables that I want to pass from `Makefile` to chaise code. Since there are multiple variables I'm adding them to an object. I called the object `chaiseBuildVariables` which will have the following attributes:

- `buildVersion`: the version used for installation. This will be used for fetching templates.
- `chaiseBasePath`: This will be used in navbar/login apps as well as `utils.js` to determine the prefix of chaise URLs.
`ermrestjsBasePath`: Used in navbar/login apps to determine the prefix of ermrestjs URLs.

I also updated navbar and login apps documentation and moved them to be under `docs/user-docs` folder.

RFSH:
To make it easier for developers that are using navbar app, we decided to add `angular.js` to list of chaise dependencies that are optional, and `navbar.app.js` will fetch itself.

PR #1976, opened 2020-07-08
(1953) Issues with Print in record app

DamCousik:
This PR consists the code for the following fixes:

- [x]  Record action buttons are hidden
- [x]  Removed justification for all text
- [x]  Screen Action buttons are hidden
- [x]  Removed the 'Displaying information' button
- [x]  Side panel is hidden
- [x]  Made the collapsible arrows smaller
- [x]  Vocab class presented as a bullet separated list
- [x]  Changed the font size for all the headers
- [x]  Removed the Action/View column from the record header table
- [x]  Whitespace reduced between the columns
- [x]  Table borders are restored in print preview
- [x]  Top arrow key hidden
- [x]  Add a class to custom style elements according to browser type and OS


RFSH:
@DamCousik I will look at your changes later, but at first glance, I noticed that for every simple change in your code, you'd done a commit. This is not harmful by itself but will pollute the repo and will make git history completely useless.  Also, all of your messages seem to be generated by the GitHub desktop app.

Everyone has their preference of when to commit a change. But as a general rule, you would commit something when you want to add that change to history, and not something that you're just locally testing. Also, your commit message should just summarize your change and shouldn't be some automatically generated string. 

That being said, we can easily "squash" your commits when we get to merge this PR. But that's not always ideal as we want to keep your history of changes in the repo.

common/utils.js
```diff
@@ -2630,6 +2630,27 @@
...
+
+          if ((navigator.platform.indexOf("Mac")!= -1) && (navigator.userAgent.indexOf("Firefox")!=-1))
+          {
+            osName="MacOS";
+            browserName = "Firefox";
+          }
+
+          if((osName === "MacOS") && (browserName === "Firefox"))
+          {
+              var newCustomClass = "chaise-mac-firefox";
```
jrchudy, line 2646: Instead of forcing the conditional check to be for BOTH mac and firefox, it's better if we do each check separately. We want to add a separate class for both the OS and the browser. So in the case that we had an issue with, we want to add 2 classes to the element, `chaise-mac` and `chaise-firefox`.

Let's change this whole function to something like the following:
```
function addMacFirefoxClass () {
  var osClass = ( navigator.platform.indexOf("Mac") != -1 ? "chaise-mac" : undefined );
  var browserClass = ( navigator.userAgent.indexOf("Firefox") != -1 ? "chaise-firefox" : undefined );

  var bodyElement = document.querySelector(".chaise-body");
  if (bodyElement){
    if (osClass) UiUtils.addClass(bodyElement, osClass);
    if (browserClass) UiUtils.addClass(bodyElement, browserClass);
  }
}
```

After making this change, you will have to update your selector in the css document as well to `.chaise-mac.chaise-firefox`.

We want 2 separate classes here so we can later expand this for each OS and browser type and more easily define combination of rules and classes without iterating over every combination.

jrchudy:
Please document the OS and browser class information here:
https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/custom-css.md

Please document the findings you learned that could help the team and future developers here:
https://github.com/informatics-isi-edu/chaise/blob/master/docs/dev-docs/dev-guide.md

common/styles/scss/_record-app.scss
```diff
@@ -304,41 +304,247 @@
...
+            }
+        }
+
+        #page-title{
+            font-size: 1.75rem !important;
+        }
+
+        #rt-container{
+            .vocab::after{
+                content: "\2002\2022" !important;
```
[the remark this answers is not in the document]
DamCousik: Yes, I will add the comment. This is getting the markdown-vocab class to be displayed without the bubbles and just as a dot separated list

docs/dev-docs/dev-guide.md
```diff
@@ -34,6 +34,30 @@ To make sure you're familiar with both CSS and SASS. The following are some guid
...
+### Useful CSS rules
+How each browser renders printing styles is different from the other. Mac and Windows behave differently for the same browser type (Firefox, Chrome, etc). Hence we need to keep in mind the following while writing print rules in css.
+
+  - If table borders or other line elements are not visible in the print preview or the PDF, check if there exists any overriding
+    bootstrap rules. These bootstrap rules could be a `background-color`, `background-border`, etc. and they always take precedence over the custom css rules that are defined in the @media-print section of the css file.
+
+  - If yes, then we must override those rules with `!important` to get the desired effect.
+
+  - A new class has been defined to apply custom styling to the case of Firefox browser in combination with MacOs which can be found here
+    :(https://github.com/DamCousik/chaise/blob/origin/isi-isrd-local-dev/docs/user-docs/custom-css.md)
```
DamCousik, line 54: I will change this GitHub link to point to the main repo once the changes have been approved.

docs/dev-docs/dev-guide.md
```diff
@@ -34,6 +34,30 @@ To make sure you're familiar with both CSS and SASS. The following are some guid
...
+
+  - Use the print mode in the rendering tab to see how the document looks when printed in Chrome browser. On Firefox, this can be achieved
+    by clicking on a small page icon in the "Inspect Element mode".
+
+  - The print preview that is seen when doing a `Ctrl-P` on Windows or a `™£¡•-P` on Mac doesn't necessarily give you the right picture of
```
jrchudy, line 59: The icon might not show in browser for others not on mac. Easier to write out `Cmd-P` instead

webauthn=QPNl0msVe56uERFbSTNs8ttI
```diff
@@ -0,0 +1,5 @@
+# Netscape HTTP Cookie File
+# https://curl.haxx.se/docs/http-cookies.html
+# This file was generated by libcurl! Edit at your own risk.
```
jrchudy, line 3: you should remove this file. It has your cookie info in it for rbk

docs/user-docs/custom-css.md
```diff
@@ -87,8 +87,16 @@ For example, if your column name is `text (name)`, chaise will rewrite this as `
 
 ### Examples:
 
+- Added a class to with deal column spacing in Firefox browser on Mac
```
jrchudy, line 90: It might be better to move this to the top section. https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/custom-css.md#changing-chaise-table-styles

Can you change the title to `Changing chaise styles` and modify the first paragraph after that heading to:
 - With the current HTML structure, it is possible to apply different styles to different parts of chaise and to further style the chaise tables based on model information.

Make sure to use the classes you added since this has `.addMacFirefoxClass` in your example.

DamCousik: Corrected this

PR #1977, opened 2020-07-09
Annotation drawing

RFSH:
This PR will add annotation manipulating features to the viewer app.  The changes were first done by @iamyle0710  and I made more changes to make sure it can be merged with master.

There are a lot of changes in this branch so I'm not exactly sure how we should go about reviewing this PR, but I will try to summarize the changes in the following list:

### Chaise (non-viewer) changes:
- Modified `inputSwitch` directive to support a more general case. The existing directive is working in a special mode. I didn't want to completely change the select-all implementation, so now the `inputSwitch` is used in two different modes (it's determined by the passed attributes):
    - select-all: doesn’t have an initial value. The value that is stored and displayed is not necessarily what will be sent and we’re using an extra translation layer.
    - normal field (used in viewer): can have initial value and the stored value is exactly what will be sent.

    When I said "sent" I didn't mean sent to ermrestjs. No matter what we’re still going to have a translation layer that is called `populateSubmissionRows`. So the problem with select-all case is that it has an EXTRA translation layer.

- Fixed the implementation of markdown preview in the `inputSwitch` directive. 
    - The code was naively clicking on a button in recordedit page to show the markdown preview. I modified it to directly call the `showModal` Function that we have. This way it can be used in other places as well. While doing this, I also refactored the code a bit (removed the `markdownPreview.js` and moved the implementation to `modal.js`) and removed other unused packages (`angular-datepicker`, `rzslider`.)

- Moved some of the functionalities of recordedit app to the common `recordCreate.js`  file. So both recordedit and viewer can use them for showing the form.

- Fixed the failing test cases.

### Viewer related changes
- Removed the `viewer.css` and moved the new styles that were added to the `_viewer-app.scss`.

- Removed alerts and other old/unused services that were specific to the viewer app in favor of the general Chaise versions.

- Added a new `viewerConstant` constant attribute so it's easier to update.

- Refactored `viewer.app.js` by removing unused code and making the use of `Reference` API.

- Changed the HTML structure of viewer app to be aligned with the rest of Chaise.

- Fixed the spacing issue in the popup (it was because of the missing DOCTYPE definition.)

- Added edit/create forms to the viewer app using the same directive/servies that recordedit is using.

The focus of this PR was to get the annotation drawing feature to a point that is "mergable". After merging this PR, I should spend some time refactoring the code (You might notice the `TODO`s that I left in the code.) We also are not properly supporting logging in this app (which should be addressed later).


viewer/annotations/annotations.controller.js
```diff
@@ -477,35 +476,108 @@
...
+
+            item["groupID"] = data[idColName] + "," + data[nameColName];
+            item["name"] = data[nameColName];
+            item["id"] = data[idColName];
+            item["url"] = url;
             return true;
         }
 
+        /**
+         * disable the
```
jrchudy, line 527: half of a comment

viewer/constant.js
```diff
@@ -4,13 +4,35 @@
     angular.module('chaise.viewer')
 
     .constant('viewerConstant', {
```
jrchudy, line 6: Could you add a comment explaining why this is hard coded (I know the app is only used by RBK).  Maybe we should discuss a way to expand this to be customizable in the future.

RFSH: We're still not sure how this will be used in other deployments. So for now, @hongsudt suggested using something inside the viewer code but centralized that can later be moved to a configuration file. Since we were not sure we want to reuse the chaise-config method, or introduce a new method for this; I just decided to add a constant in the code.

I will add this comment to the code as well.

RFSH:
@jrchudy I addressed your comments and pushed another change to address the performance issue that we were seeing on the page. 

I tried to explain this in my commit, so In the following, I will copy what I wrote in there:

There are three different events that happen that should be ordered properly:

- (1) When all the requests are done and iframe location is updated, so we tell the viewer app that its "displayReady".
- (2) When we initially set the height of the bottom-container.
- (3) When we show the iframe.

With the previous approach, we were doing (3) as soon as the HTML page loads. Then (1) and (2) together. Since we were doing (1) and (2) together, by the time that (2) was done, the page wasn't still showing anything and therefore it couldn't set proper height. So the iframe would show up with a small height. If the iframe had to load a lot of images, the lag would delay the logic to update the height and the user could see the small iframe for a few seconds.

I modified the code to do (1) and in the second digest cycle, it would do (2) and (3) together. So by the time that we're setting the initial height the rest of the page is properly displayed and the initial height is what we want. And we're also going to show the iframe at the end of this second digest cycle, so you won't see a small iframe on load.

PR #1989, opened 2020-07-30
Hiding iframe and video markdowns during print

DamCousik:
This PR includes the following fixes

- [x] Fixed the print for iframes with Treeview document to show the tree hierarchy during print.
- [x] Persisted scrolling during print
- [x] Parsing the URL to check if there is an embedded YouTube video. If yes, then add a class to apply the above discussed custom styles.
- [x] The same has been processed for the <video> markdown pattern as well.

- Applying the thumbnail to the YouTube video on print was complicated as the styles are applied via YouTube and we have limited access to it. Hence, the alternative was to display a note informing the user that the video is hidden during print and possibly provide the URL of the hidden video.

### DIFF
--- a/common/styles/scss/_record-app.scss
+++ b/common/styles/scss/_record-app.scss
@@ -536,7 +536,7 @@
           visibility: visible !important;
         }
 
-        alerts {
+        .alert {
           display: none !important;
         }
 
--- a/common/styles/scss/_record-app.scss
+++ b/common/styles/scss/_record-app.scss
@@ -531,11 +531,12 @@
           display: none !important;
         }
 
+        //In print we hide YouTube and other videos. This class is then made visible to display the relevant info in print mode
         .video-info-in-print{
           visibility: visible !important;
         }
 
-        .alert-success{
+        alerts {
           display: none !important;
         }
 
--- a/common/styles/scss/_record-app.scss
+++ b/common/styles/scss/_record-app.scss
@@ -531,6 +531,10 @@
           display: none !important;
         }
 
+        .video-info-in-print{
+          visibility: visible !important;
+        }
+
         @page {
             size: auto;
             margin: 5% 3% 5% 3%;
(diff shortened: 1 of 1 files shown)


RFSH:
@DamCousik Can you please mention the issues that you've fixed in the description of your PR? That way Github will link the issue to the PR and it's easier to follow.  You can do this by using `#<github_issue_number>` format (`#1989`). You can find examples of this in the previous PRs that we've created in this repo or ermrestjs.

DamCousik:
@RFSH : I have mentioned all the issues that I have fixed in this PR. Sending the list of fixed issues again for your reference:

- [x] Fixed the print for iframes with Treeview document to show the tree hierarchy during print.
- [x] Persisted scrolling during print
- [x] Parsing the URL to check if there is an embedded YouTube video. If yes, then add a class to apply the above discussed custom styles.
- [x] The same has been processed for the markdown pattern as well.
- [x] Fixed banner not displaying on record page during print

hongsudt:
@DamCousik please read what @RFSH asked carefully. He wants you to link an issue by mentioning the issue number in the text. He even gave you an example.

common/styles/scss/_record-app.scss
```diff
@@ -531,6 +531,14 @@
           display: none !important;
         }
 
+        .video-info-in-print{
+          visibility: visible !important;
+        }
+
+        .alert-success{
```
RFSH, line 538: Since we want to hide all the alerts, you can just use the directive name. So

```css
alerts {
  display: none !important;
}
```

common/styles/scss/_record-app.scss
```diff
@@ -531,6 +531,14 @@
           display: none !important;
         }
 
+        .video-info-in-print{
```
RFSH, line 534: Add a comment above this to explain what's this class and what's its purpose.

Issue #1979, opened 2020-07-10
Chaise config naming inconsistencies


jrchudy:
It was brought up that almost all of the chaise config properties use camel case starting with a lower case letter (`camelCase`). There are 2 properties that do not, `SystemColumnsDisplayCompact` and `SystemColumnsDisplayDetailed`.

karlcz:
We certainly won't have different configuration parameters differing
only by case. Should we make chaise case-insensitive in its
interpretation of the config...?

hongsudt:
case-insensitive makes sense.. We will proceed with camel case for documentation though.
________________________________

RFSH:
#1943 is not directly related to this but can be done together. Both are related to chaise-config.

jrchudy:
Both, `SystemColumnsDisplayCompact` and `SystemColumnsDisplayDetailed`, have been updated to use proper camel case in the code and documentation. 

Chaise config properties that are a part of `chaise-config.js` and the catalog chaise config annotation including `configRules` are case-insensitive.

Issue #1943, opened 2020-05-01
Default chaise-config values and documentation


ljpearlman:
Some of the default chaise-config values seem to be different from what I think are the most commonly desired values. Specifically, showFaceting, editRecord, and deleteRecord should probably all default to true.

It's also not entirely clear to me from the documentation what some of the parameters are used for (and which were used by the old search app and are ignored by the current apps); specifically:

`catalog` -- this is distinct from `defaultCatalog`, and I'm not sure what it's used for
`feedbackURL`,  `helpURL`,  and ` logoutURL` - are these still in use?
`dataBrowser` - is that still in use? If so, the default needs to change
`maxColumns` - is this still in use?
`showUnfilteredResults` - either I'm misinterpreting the definition, the description is wrong, or it's no longer in use
`defaultCatalog` - I don't think the description is clear.
`signUpURL` - is this still in use?
`profileURL` - is this in use?

It also might be nice to indicate which apps each parameter is used for (some of them have that already, but others don't).

jrchudy:
Still used by chaise:
 - logoutURL
 - dataBrowser
   - needs update to default to `recordset`. Still used for a default redirect in case window.error is encountered
 - defaultCatalog
   - needs update the language in the description
 - signUpURL
 - profileURL

The following properties are not used by chaise anymore:
 - catalog
 - feedbackURL
 - helpURL
   - still consumed by `scripts/tour.js` which was only used in search
 - maxColumns
   - still consumed by `scripts/utils.js` which was only used in search
 - showUnfilteredResults
   - still consumed by `scripts/controller/ermrestResultsController.js` which was only used in search

jrchudy:
TODO: have a discussion about the "unused" properties and remove them after if that's what we decide.

jrchudy:
The changes detailed above have been addressed. The list of properties no longer used by chaise have been removed from the documentation and marked as deprecated in `RELEASE.md`.

Reassigning to @ljpearlman to review the changes to the documentation.

PR #1994, opened 2020-08-12
Chaise config properties can be case insensitive, support for systemColumnsDisplayEntry added

jrchudy:
This PR tackles 3 issues in github, [#853](https://github.com/informatics-isi-edu/ermrestjs/issues/853) in ermrestJS, #1979, and #1943 .

- Support for `chaiseConfig.systemColumnsDisplayEntry` has been added.
- naming inconsistencies for chaise config properties has been addressed.
  - `SystemColumnsDisplayCompact` and `SystemColumnsDisplayDetailed` have been updated to use proper camel case.
  - chaise config properties are now case-insensitive
- clean up some of the default config values and documentation for them


common/utils.js
```diff
@@ -1760,7 +1761,18 @@
         }
     }])
 
-    .factory("ConfigUtils", ['defaultChaiseConfig', '$http', '$rootScope', '$window', function(defaultConfig, $http, $rootScope, $window) {
+    .factory("ConfigUtils", ['defaultChaiseConfig', '$http', '$log', '$rootScope', '$window', function(defaultConfig, $http, $log, $rootScope, $window) {
+        // List of all accepted chaiseConfig properties in defined case from chaise-config.md
+        var chaiseConfigPropertyNames = [
```
RFSH, line 1766: I think it's better if you define this as a constant value.

common/utils.js
```diff
@@ -1794,14 +1806,39 @@
         function setConfigJSON(catalogAnnotation) {
             var cc = {};
             // check to see if global chaise-config (chaise-config.js) is available
-            if (typeof chaiseConfig != 'undefined') cc = Object.assign({}, chaiseConfig);
+            if (typeof chaiseConfig != 'undefined') {
+                // loop through properties and compare to defaultConfig to see if they are valid
+                // chaiseConfigPropertyNames is a whitelist of all accepted values
+                for (var key in chaiseConfig) {
+                    // see if returned key is in the list we accept
+                    var matchedKey = chaiseConfigPropertyNames.filter(function (defaultKey) {
```
RFSH, line 1814: You've already done in multiple places, but you could define a private helper function in this factory that does this for you (so you don't have to copy/paste the code)

jrchudy:
> Are you going to include the changes to chaise-config.md in this PR? (the changes regarding the old attributes)

Yes I will include these changes in this PR. The idea is to include those changes in this PR then update RELEASE.md to include the properties that have been deprected.

PR #1995, opened 2020-08-13
Viewer app improvements

RFSH:
This PR will,

- Add a spinner to the share button to signal users that we're waiting for the popup (in both record and viewer app).
- Introduce a new method of passing annotations to osd viewer by using messages (instead of URL parameter).
- Fix the `Image_Annotation` read request to properly send multiple requests (instead of assuming there will be less than 200 rows).
- Display spinner for the screenshot button based on changes in [osd viewer#40](https://github.com/informatics-isi-edu/openseadragon-viewer/pull/40) PR.
- Change the color of buttons in the annotation sidebar to be aligned with the rest of the chaise buttons.


RFSH:
@jrchudy I don't think that's needed. Apart from the share popup change, the rest were mainly refactoring and moving code around.

PR #2000, opened 2020-08-22
Rid search

jrchudy:
This PR adds the functionality to the navbar to include an input in the navbar that tables an RID and goes directly to that record page on submit. To show this input box, `resolverImplicitCatalog` must be on, so `!== null` based on the definition we have in [chaise-config.md](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/chaise-config.md). A new property is added to chaise-config to allow for deployments to hide this input using `hideGoToRID`.

Added a test case that only runs in travis.


jrchudy:
@RFSH I moved the styles around and rewrote the rid function. I commented how each condition pertained to each case that the resolver can be defined following the comment I wrote in `utils.js`.

common/navbar.js
```diff
@@ -284,6 +284,59 @@
...
+                        var resolverId = chaiseConfig.resolverImplicitCatalog,
+                            url = "/id/", catId, splitId;
+
+                        if (isCatalogDefined(catalogId)) {
+                            splitId = UriUtils.splitVersionFromCatalog(catalogId);
+
+                            // use `/id/catalog/ridSearchTerm` format if:
+                            //   - resolver id is a different catalog id than current page
+                            //   - resolver id is NaN and !null
+                            if (isNaN(resolverId) || resolverId !== catalogId) {
```
RFSH, line 314: In `resolvePermalink` this check is done by `==`. So to be consistent let's change this to `!=`. Then we will support both string and integer in both places

common/navbar.js
```diff
@@ -284,6 +284,59 @@
...
+                                url += splitId.catalog + "/"
+                            }
+                        } else {
+                            // we don't have a catalog id, check if reolverId is a number and use as catalog id
+                            // basically catalogId !== resolverId for this case
+                            // use `/id/catalog/ridSearchTerm` format if:
+                            //   - resolver id is a number
+                            if (!isNaN(resolverId)) {
```
RFSH, line 322: This is not consistent with `resolvePermalink`. We're either going to add the current catalog or nothing. So you should remove this code path.

RFSH:
We discussed about using consistent terms for the following:

- Button placeholder: `Go to RID`
- chaise-config property: `hideGoToRID`
- log action: `:,navbar/go-to-rid;navigate`
- extra logged attribute: `rid`

Issue #1917, opened 2020-03-10
log eye v.s. link when direct to a record page


hongsudt:
- Consider adding a new url parameter to pass to the record page e.g. `paction=view|link` or `plink`, `pclick`
- Extend the `pcid` to add this extra info makes the string different from `cid` which makes the join during the analysis harder.

hongsudt:
@RFSH Did we do this in the log already? If so, can you comment and close this issue?

RFSH:
No, we haven't implemented this.

RFSH:
All the links in chaise are generated by the `appLink` function. This includes that all the buttons (create, edit, view) as well as links that are generated based on the pseudo-column type (or the markdown display settings).

We can make modifications in different places to pass different query parameters. For example for the view button (eye icon) in a tabular format, I can pass this extra query parameter. But I don't see the benefit of adding the query parameter to the rest of the links as well. I think we limit the change for now and just add a specific query parameter to the view button.

RFSH:
This has been done as part of the #2001 PR. I added a section called ["Find number of clicks on the `view` button in recordset app"](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/logging.md#find-number-of-clicks-on-the-view-button-in-recordset-app) to provide more information.

@hongsudt Can you please verify and close this issue?

RFSH:
Closing this issue due to inactivity.

Issue #1910, opened 2020-02-27
Logging in viewer app


RFSH:
We're currently working on adding new features to viewer app. In the first phase, we're adding modal pickers to viewer app so that users can edit the existing annotations or create new ones. We're using the same modal picker that we have in other parts of chaise in viewer app. We should make sure we're logging these new reqeuest correctly and consistently. 

I created this issue as a reminder for myself so that I have to make sure we're setting the `stack` and `stack-path` of these requests correctly.

This is used in "annotation" part of the page, so we decided to change the `stack` structure of getting the "annotation" list, to include a new element (apart from the element related to the main request). Since "annotation" is behaving the same way as "related entities" in the record page, we thought about using `related` in the stack path. But later we changed it to use `annotation` to be more clear. But we're still not sure about the `type` of stack element that we're adding. For now, I used `annotation` to be consistent. 
So the request for getting the annotation is going to change from 

```javascript
// old
{
  "action": ":entity,annotation;load",
   "stack": [
    {"type": "entity", "s_t": "main_table", "filters": {"and": [{"source": "RID", "choices": ["value"]}]}}
  ]
}
```

to

```javascript
// new
{
  "action": ":entity/annotation,;load",
   "stack": [
    {"type": "entity", "s_t": "main_table", "filters": {"and": [{"source": "RID", "choices": ["value"]}]}},
    {"type": "annotation", "s_t": "annotation table"}
  ]
}
```

And about the modal picker,s  they behave exactly like `fk-picker` in recordedit and we're just filling out the rest of values. So we're going to use the same `fk-picker`, and since they are part of annotation the whole stack path will be `entity/annotation/fk-picker`. We also want to be able to capture edit vs. create. We've ] this by using `app-mode` in recordedit, so we could use the same convention. Although with this change, we're diverging from the actual definition of `app-mode` since now this won't be affecting the other requests and will only be used for these fk pickers. I am not going to list all the requests in this case and in the following you can only just see how load, update and create requests will look like in this case.

```javascript
// load edit picker
{
  "action": "edit:entity/annotation/fk-picker,;load",
   "stack": [
    {"type": "entity", "s_t": "main_table", "filters": {"and": [{"source": "RID", "choices": ["value"]}]}},
    {"type": "annotation", "s_t": "annotation table"},
    {"type": "fk", "s_t": "fk table", "picker": 1}
  ]
}

// update request of edit picker
{
  "action": "edit:entity/annotation/fk-picker,;update",
[...]

PR #2001, opened 2020-08-25
Logging support for viewer app

RFSH:
This PR,

- Adds proper log support to the viewer app (#1910). For the complete list of actions please [refer to this google sheet](https://docs.google.com/spreadsheets/d/1cHhPR0AacvuH2o3QavWlJCIIyPajfn93fJzFnc7VA5M).
- Add extra log info to distinguish between view button and the rest links (#1917)
- Fix a bug in the viewer app which would cause users to be stuck in the edit form (Open edit form for an annotation, don't change anything and click on save.)
- Unhighlight annotations when switching to edit mode. 
- Removed the unnecessary usage of `slide_id` in the code.
- Fixed a small bug in the logic of disabling the "show/hide annotation" button. It wasn't properly checking the value that osd viewer is returning.

### DIFF
--- a/common/ellipsis.js
+++ b/common/ellipsis.js
@@ -121,7 +121,9 @@
 
                     // view link
                     if (scope.config.viewable) {
-                        scope.viewLink = tupleReference.contextualize.detailed.appLink;
+                        var viewLink = tupleReference.contextualize.detailed.appLink;
+                        var qCharacter = viewLink.indexOf("?") !== -1 ? "&" : "?";
+                        scope.viewLink = viewLink + qCharacter + "paction=view";
                     }
 
                     // if tupleReference is not defined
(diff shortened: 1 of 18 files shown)


jrchudy:
The only thing I had a comment about was adding the logging functions to the annotation service. I think it makes sense to do it there but was wondering why in annotation service instead of the existing log service?

RFSH:
@jrchudy I've done the same in table.js and faceting.js as well. I also did the same in recordedit in this PR (in recordCreate.js). 

These functions are more specific to the ones in log service. They're mainly some wrappers that use the contextualized data in each place (for example in table.js it's using the `logStack` attached to `vm` or in annotation service I'm using the `logStackNode` that is attached to each annotation model) and call the general functions in log service. We could eventually move all these functions to log service and pass the contextualized data as well if we find a more general pattern in them.

PR #2008, opened 2020-09-11
Navbar dropdowns scrolling

jrchudy:
This PR adds scrolling support to the navbar dropdown menus and sub-menus. Changes include:
 - setting a fixed height for the navbar dropdown if it's too long
 - adjusting this fixed height when the page is resized
 - adding a space to the right side of the caret elements for scrollbar space
 - keeping the clicked parent option highlighted
   - removing the highlight when another is clicked or the menu is closed


common/navbar.js
```diff
@@ -372,6 +438,12 @@
                                 names: menuObject.names
                             });
                         }
+
+                        var openSubmenues = document.querySelectorAll(".dropdown-menu.show");
```
RFSH, line 442: What's the point of this piece of code? and why is it not part of the `toggleMenu` function? 

I'm assuming it's making sure the scrollbar is added by setting the height, but it's also forcing the submenu to scroll to top.

common/navbar.js
```diff
@@ -253,12 +297,20 @@
...
+                            // when menu opens, calculate height needed
                             logService.logClientAction({
                                 action: logService.getActionString(logService.logActions.NAVBAR_MENU_OPEN, "", ""),
                                 names: menuObject.names
                             });
                         }
+
                         onToggle(open);
+                        // unset height on dropdowns first
+                        var openDropdowns = document.querySelectorAll(".dropdown.open ul");
```
RFSH, line 309: Shouldn't this be part of `onToggle` function?

RFSH (approved):
Everything works fine and as expected. Since you've spent some time figuring out the code, I think it might be worth adding comments to these functions. So the next time we can pick it up faster. 

The only issue that I found (and I'm not sure how it can be solved or if it's even an issue) is when you open a second level menu, it will scroll to top. You can test it here:
https://dev.rebuildingakidney.org/~ashafaei/chaise/record/#2/Common:Collection/RID=Q-3K5C

Open "Gene Expression Data" and then scroll down to any menu item with children (I copy/pasted the existing navbar so the children are duplicated). As soon as you click on the submenu, it will show the third level submenu in a proper position. But it also scrolls the second level to top.

common/navbar.js
jrchudy: Calling it here because of needing `$window.innerHeight` for the `checkHeight` function. $window is no defined in the scope for the `onToggle` function.

common/navbar.js
jrchudy: That's the intention, it's supposed to reset the height values so new ones can be properly set. I didn't realize this was causing the scroll top bug you mentioned with a nested child of submenus.

This block is defined here for the same reason as above, `$window` is not available in the scope of `toggleMenu`.

common/navbar.js
RFSH: You can pass the $window object to this function. I think it's cleaner to have all the functionality in the same function.

jrchudy:
@RFSH I updated the changes for this PR. I moved the calls to `checkHeight` inside of `onToggle` and `toggleMenu` since we can rely on `window` in that scope. `$window` is just a wrapper for `window` to give it more angular properties. I also removed unsetting the height for all selected elements and instead only unset the height for the element being toggled.

https://dev.isrd.isi.edu/~jchudy/chaise/recordset/#1/isa:dataset@sort(accession::desc::)

RFSH (approved):

PR #2018, opened 2020-10-08
Remove pseudocolumns in multi mutation table view

jrchudy:
When users create or update multiple records at once, the post mutate recordedit set view will no longer show pseudo columns that are also path columns (comment in code). This turned out to be a little more complicated than "just remove the column from the set".


common/table.js
```diff
@@ -900,7 +900,12 @@
             if (vm.attributesAlreadyAttached) return;
             vm.attributesAlreadyAttached = true;
             vm.columnModels = [];
+
             vm.reference.columns.forEach(function (col) {
+                // if flag set, remove pseudo columns that are path columns
+                // NOTE: could check isPseudo, isPathColumn, and isForeignKey
+                if (vm.removePseudoColumns && col.isPseudo && col.isPathColumn) return;
```
jrchudy, line 907: Is this the right way to check if a column should be ignored? If it is a pseudo column and a a path column?

Should it be `isPseudo && !isForeignKey`?

recordedit/form.controller.js
```diff
@@ -128,7 +161,8 @@
                     sortOrder: null,
                     page: page,
                     pageLimit: model.rows.length,
-                    rowValues: DataUtils.getRowValuesFromTuples(page.tuples),
+                    removePseudoColumns: true,
```
jrchudy, line 164: Not sure if this is the best way to modify how `vm.resultsetModel.columnModels` is created since `vm.resultsetModel.columns` is not used for the table of columns anymore.

RFSH:
Based on the conversation that we had it seems like we want to add a specific context to ermrestjs and do this based on the context. Then chaise can just use this specific context.

This is where you can modify the `generateColumnsList` function to check for context and exclude some columns:
https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L3064

You shouldn't just exclude columns based on their types. You should exclude columnes:
- that require some secondary request that we're not going to send in this case (aggregate pseudo-columns and columns with aggregate pseudo-column wait_for.)
- multi-level all-outbound foreign keys that we don't have their values.


So
```
refCol.hasWaitFor || !refCol.isUnique || (refCol.hasPath && refCol.isUnique && refCol.foreignKeys.length > 1)
```

recordedit/form.controller.js
```diff
@@ -145,14 +148,15 @@
                 // NOTE: This case is for a pseudo-failure case
                 // When multiple rows are updated and a smaller set is returned, the user doesn't have permission to update those rows based on row-level security
                 if (failedPage !== null) {
+                    var failedReference = failedPage.reference.contextualize.entryCompact;
```
RFSH, line 151: The page that is used here is created by create/update function. This means that it already is using the structure of reference and changing the context afterward doesn't have any effect.

So you should change the context in ermrestjs instead.

RFSH: for create:
https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L1251
for edit:
https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L1842

RFSH (approved):

PR #2022, opened 2020-10-21
Markdown name support for navbar options

jrchudy:
Added markdown name support to navbar menu and submenu options. This allows for markdown patterns to be defined to attach extra attributes (including class) to the element being defined.


RFSH (commented):
The changes look good and work as expected. I think you should modify our existing navbar test case to test this one as well. It should cover all the cases 
- Just name
- Just markdowName
- both name and markdownName

RFSH (approved):

PR #2026, opened 2020-10-30
Add chaise config property for share cite button

jrchudy:
This PR adds support for the ability to show/hide or enable/disable the button that opens the share and cite dialog. The property is defined with the ability to control both modes using globus group membership similar to how we use acls to show/hide or enable/disable navbar menu options.

```js
shareCiteAcls: {
  show: ["*", ],
  // show: [], // hide
  enable: ["*"],
  // enable: [] // disable
},
```

The follow is how I tested 3 cases with e2e tests:
 - one where it’s not defined and all defaults are applied (no changes needed since specs were testing opening the popup)
 - defining the object with both show and enable as “*” (same reason above since we are opening the popup for tests)
 - one case for resolverImplicitCatalog relied on being in travis, so I added an else case to test config functionality not in travis


record/record.controller.js
```diff
@@ -112,6 +112,40 @@
             });
         };
 
+        vm.canShowSharePopup = function() {
```
RFSH, line 115: This exact functionality of matching group-id with the user session is done in four different places (two here, and two in navbar). I think you should create a function and call it in all these four places (two here, and two in navbar). it could be part of the authen.js

```javascript
// where allowedGroups is showAcls in this case, and then in that function you already have access to _session so there's no need to pass it.
whateverFnName: function (allowedGroups) {
  ...
}
```

One thing that I noticed is that in the navbar code, we're properly handling empty session, while here the check is not included.

jrchudy: In this case the session being available won't limit the button from showing up, but I should be checking for it before calling `some` on `session.attributes`.

RFSH: Why not? If you define

```
showAcls: {
  "show": ["some-user-group"]
}
```

You only want to show it to a special user group, and therefore the user must be logged in (session must be valid). So I think in both cases you need to check the session. Only if the setting is defined as `["*"]` it should be visible regardless of session value.

RFSH (approved):

PR #2027, opened 2020-11-06
Dynamic multi-channel support in viewer

RFSH:
This PR will add the required changes for [multi-channel dynamic configuration](https://github.com/informatics-isi-edu/openseadragon-viewer/issues/49). 

Related PR: [OSD viewer#55](https://github.com/informatics-isi-edu/openseadragon-viewer/pull/55)

### Chaise-wide changes:
- Modified the query parameter related functions in `utils.js` so they can be properly used in the viewer app as well.
  - Add an option to avoid encode/decoding query parameters (in OSD viewer, query parameters are not URL-encoded.)
  - Allow multiple query parameters with the same name. The previous code wouldn't properly work with `?url=val1&url=val2` as it was a simple key-value pair of strings. Now the values will be an array of strings in this case.

### Viewer app changes:
- Add support for reading the Image_Channel table if needed. Also modified the way we're OSD parameters (explained in the next section.) 
- Pass the OSD viewer parameters by message instead of query parameters.
- Refactor the viewer.app.js code a bit and introduce new functions in viewer.utils.js
- Add new constants that are used with the new logic.

---------------
#### New logic for OSD parameters:

The following is the new logic for deriving the OSD parameters (what chaise passes to OSD viewer):

- There are [two different categories of OSD parameters](https://github.com/informatics-isi-edu/openseadragon-viewer/tree/channel-refactor#invoking-openseadragon-viewer), 
  - channel-related: `url`, `channelName`, `pseudoColor`, `isRGB`, `aliasName`
  - general: `waterMark`, etc

- Different OSD parameters might be derived from different sources based on this priority order: 

  1. query parameter on the page

  2. `Image.uri` value

  3. `Image_Channel` (`Image_Channel` is only used for channel-related parameters.)

  This means that if a specific parameter is defined on both query parameter as well as `Image.uri`, we will use the value defined on the query parameter. In this logic, each parameter is treated separately (in this example another parameter might be derived from `Image.uri` or any other sources).

- Channel-related parameters are a bit different as they are treated as all-in or nothing. We first check if channel `url` is defined ([ignoring annotation `url`s](https://github.com/informatics-isi-edu/openseadragon-viewer/issues/54)). If that is the case, all the other channel-related will only be derived from this source and will be ignored by others. For instance, if user navigate to a page with `?url=valid_image_url` while the `Image.uri` value is `?url=another_valid_image_url&channelName=DAPI`, we will use the `url` defined in the query parameter and ignore the `url` and `channelName`s defined on `Image.uri`.


RFSH:
@jrchudy I found [a typo in my code](https://github.com/informatics-isi-edu/chaise/pull/2027/commits/c4a8676551151977c8259e484ee904da01dd20bc) that was causing `limit` (and other query parameters) to be ignored which in turn broke recordset test cases.

This PR is ready to be reviewed. As I mentioned in the description the only chaise related changes are changing functions related to query parameters. The rest of the changes are isolated to the viewer app itself.

jrchudy (approved):
Changes to chaise utility functions looks good. Only major change I see is passing an option to NOT decode query params

Issue #2015, opened 2020-09-30
Color picker support


RFSH:
In the new Image Channel table, we are going to store the color value. And then users should be able to modify this color through the recordedit app. These are the list of tasks that should be done for this change (it's not limited to chaise):

- [x] Determine the web standard color format  (hex, rgb, etc)
- [x] Define a custom type in ERMrest
- [x] Expose the new type to ERMrest
- [x] Add color picker in recordedit
  - We can use [the color input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color). Although this input is not supported by IE.
- [x] For display purposes we should display a rectangle with that color + the raw color value

- [ ] ETL the existing Pseudo_Color column values.
  - The color in dzi images are currently stored as tuples with three values where they respectively represent red, green, and blue. Each value is represented as a number in 0-1 range. To convert to a 0-255 scale, you just need to multiply each one by 255. And then to convert to hexadecimal format, you would convert each value to base 16, and then concatenate them together. For example:

    - the color in dzi image: `(1.000000 0.000000 0.000000)`
    - converted to RGB 0-255 format: `(255, 0, 0)`
    - converted to hexadecimal format: `#FF0000`

RFSH:
Based on [w3 specification](https://www.w3.org/TR/2018/REC-css-color-3-20180619/#colorunits) (and more user friendly [MDN web docs page](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value)), the color CSS3 data type represents a color in the sRGB color space. It may also include an alpha-channel transparency value, indicating how the color should composite with its background. It can be defined,

- using a keyword (such as blue or transparent)
- using the RGB cubic-coordinate system (via the #-hexadecimal or the `rgb()` and `rgba()` functional notations)
- using the HSL cylindrical-coordinate system (via the `hsl()` and `hsla()` functional notations)

I couldn't find any documentation that says one format is better than the other, or one is standard and the other one is not. So I don't think there really is a single "standard" web color format and we just have to pick a format that fits our needs. As was suggested by @karlcz, we should be specific about the format in the name of the custom type and that would also allow us to introduce other color types in the future (if needed.)

Therefore,
- in this case, we most probably want users to pick a fully opaque color. So we don't want to offer `rgba` or `hsla` formats. 
- If we end up using the `color` input in HTML, we can tailor the type based on its specification. This input only accepts 7 character hexadecimal (`#rrggbb`) color format.
[...]

PR #2028, opened 2020-11-16
Color rgb hex

RFSH:
This PR will add the support for color picker in Chaise. [ERMrestJS#864](https://github.com/informatics-isi-edu/ermrestjs/pull/864) makes sure the data is properly served to Chaise.

Summary of changes:

- Added the style for the new  `.chaise-color-preview` class that ERMretJS will use for color preview in read-only modes.
- Added a new `color` input type and `colorPicker` directive that is used in recordedit and `inputSwitch` for showing a color input. I used [this package](https://seballot.github.io/spectrum/) for the color picker functionality. It was the only proper and functional color picker that I could find. The only downside of using this one is the fact that it needs jQuery (which is not an issue really, just something that I hope it didn't need.)
- Modified test cases in both read-only and entry modes to test this new column type.

~P.S. I commented the test case that I added to `recordedit-helper.js` because it's not working properly. I'm currently trying to fix it.~

related issue: #2015


jrchudy (approved):
Changes look good. I'm finishing test cases for an issue I'm working on and will run your test cases and make sure everything is working still once I finish mine.

RFSH:
@jrchudy I fixed my color test cases and they should be working properly now.

While working on this, the broken float test case was causing issues for me (as I couldn't properly run the new tests.) So I spend some time trying to rewrite that `it` with proper promise-chain. But that didn't help at all. Now it's easier to figure out where exactly in the `it` the error is happening, but I couldn't fix it. Besides, this test originally was testing validation. I was setting an invalid value and was making sure that the error is displayed. But ever since it has been modified and wasn't event doing that (it was setting the input to some valid value and was clearing it.) So for now, I just ignored it by doing `xit`. I'll spend a little more time trying to figure this out.

jrchudy:
Tests are passing for me locally.

PR #2029, opened 2020-11-16
Handle 404 for rid search input

jrchudy:
I modified go to RID function to send a head request to the server first to see if the resource exists before trying to navigate there. If the server returns 405 (not allowed), we know the RID exists and should continue with navigating the user to the record page with the input RID.

I added a test that only runs locally to verify the resolver is working properly with a valid RID input into the RID search box. I modified the other test to run in both cases since locally catalog 4 shouldn't exist and in travis the resolver isn't configured even though we have the chaise config property (`resolverImplicitCatalog` set). Both of these cases should throw a warning message.

### DIFF
--- a/common/navbar.js
+++ b/common/navbar.js
@@ -214,7 +214,7 @@
         'chaise.login',
         'chaise.utils'
     ])
-    .directive('navbar', ['ConfigUtils', 'ERMrest', 'logService', 'Session', 'UriUtils', '$rootScope', '$sce', '$timeout', '$window', function(ConfigUtils, ERMrest, logService, Session, UriUtils, $rootScope, $sce, $timeout, $window) {
+    .directive('navbar', ['AlertsService', 'ConfigUtils', 'ERMrest', 'logService', 'Session', 'UriUtils', '$rootScope', '$sce', '$timeout', '$window', function(AlertsService, ConfigUtils, ERMrest, logService, Session, UriUtils, $rootScope, $sce, $timeout, $window) {
         var chaiseConfig = ConfigUtils.getConfigJSON();
 
         // One-time transformation of chaiseConfig.navbarMenu to set the appropriate newTab setting at each node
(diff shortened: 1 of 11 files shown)


common/navbar.js
```diff
@@ -365,6 +365,7 @@
                     //  - if resolverImplicitCatalog === catalogId:   /id/RID
                     //  - otherwise:                                  /id/catalogId/RID
                     scope.ridSearch = function () {
+                        $rootScope.showSpinner = true;
```
RFSH, line 368: Since this will not change the current page location (and will open a tab instead), I think we shouldn't show the spinner that we show for the whole page. I think instead we should just disable the current input and change the icon of it to show spinner. This is how I did it for export and share buttons.

common/navbar.js
```diff
@@ -388,7 +389,24 @@
                             rid: scope.ridSearchTerm
                         });
 
-                        $window.open(url, '_blank');
+                        // try to fetch the resolver link to see if the path resolves before sending the user
+                        // NOTE: .head returns a 405 error (not allowed) for proper RIDs, maybe ask Karl if this is expected
```
RFSH: If that's the case maybe we should just send a get request instead? I agree that we should ask Karl.

common/navbar.js
```diff
@@ -388,7 +389,24 @@
...
+                        ConfigUtils.getHTTPService().head(url, {}).then(function (response) {
```
RFSH: I think you should send `deriva-client-context` header with this request for proper logging. And then you don't need the client log that you're doing before this.


```javascript
var logObj = ConfigUtils.getContextHeaderParams();

logObj.action = logService.getActionString(logService.logActions.NAVBAR_RID_SEARCH, "", "");
logObj.rid = rid: scope.ridSearchTerm;

headers[ERMrest.contextHeaderName] = logObj;

onfigUtils.getHTTPService().head(url, {headers: headers})
```

common/navbar.js
jrchudy: Sounds good. That's better in this case

common/navbar.js
jrchudy: From karl:
> I suspect 405 really means we didn't' implementn HEAD and it doesn't' mean anything

jrchudy:
Changed the error to show a dismissible modal with the updated text for when a user is not logged in.

Also fixed 2 test cases in `errors.spec.js` so they properly redirect to the home page now after adding `dataBrowser: '/'` to the chaise config.

RFSH (approved):

PR #2031, opened 2020-12-02
Preformat boolean

jrchudy:
Changed input utils to format boolean values based on `column.display.preformatConfig` if it is present. This was only an issue in recordedit which didn't show the pre-formatted boolean values in the dropdown. This also applies when reading data in for edit mode. Before submitting the data to ermrestJS for create/update, we have to convert the "formatted" value back to the raw boolean value. This can be tested here:
https://dev.rebuildingakidney.org/~jchudy/chaise/recordedit/#2/Common:Collection/

I think it makes sense to do this in chaise since it applies to the display of the content.

### DIFF
--- a/common/inputs.js
+++ b/common/inputs.js
@@ -4,8 +4,6 @@
     angular.module('chaise.inputs', ['chaise.validators', 'chaise.utils'])
 
     .factory('InputUtils', ['dataFormats', 'defaultDisplayname', '$rootScope', function(dataFormats, defaultDisplayname, $rootScope) {
-        var defaultBooleanValues = [true, false];
-
         /* Functions for all input types */
         // determines if input should be disabled based on ermrestJS API
         function isDisabled(column) {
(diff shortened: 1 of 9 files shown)


common/inputs.js
```diff
@@ -24,6 +22,22 @@
             }
         }
 
+        /* boolean specific function */
+        var defaultBooleanValues = [true, false];
+
+        // checks for preformat config before returning true/false
+        function formatBoolean(column, value, context) {
+            if (column.display.preformatConfig) {
```
RFSH, line 30: I'm going to comment this here, but it's applicable to other places that you've called `formatvalue` or checked `preformatConfig` as well:

- You shouldn't check for `column.display.preformatConfig` and should always call the `formatvalue`. The only thing that we care about is making sure we're showing the same formatted value in the dropdown as we do in record/recordset app.

- The passed `context` is only used for figuring out what the "null" value should be mapped to. I also noticed that it doesn't have to be passed to `formatvalue` function. So I [pushed a change to ermrestjs to use the Reference's context if this parameter is missing](https://github.com/informatics-isi-edu/ermrestjs/commit/8620b51138be64c9e896321f78b17014b6cb1908). And now you can just remove it from all the places that you've called the function. Just `column.formatvalue(value)` should be enough.

RFSH (approved):

PR #2032, opened 2020-12-02
share and cite dialog title

jrchudy:
Changed share and cite dialog title when citation is present to "Share and Cite". Was previously just "Share" regardless of a citation being present.

This change could have been made before calling `openSharePopup` but then would require changes in a few different places. Since the citation we use in the popup is based on the passed reference, I figured it would be more important to set the title to "and Cite" in the method before opening the modal.

In `shareCitation.modal.html` we previously had the following for the title:
```HTML
<h2 class="modal-title text-center">Share {{ctrl.params.title}}</h2>
```

`params.title` was not being used before for this modal so I figured reusing this property would be sufficient rather than changing the whole title.


common/modal.js
```diff
@@ -76,6 +76,7 @@
 
             params.displayname = refTable.name+'_'+tuple.uniqueId,
             params.reference = reference;
+            if (reference.citation) params.title = "and Cite";
```
RFSH, line 79: The `params` might already have `title` in it (in viewer app, I'm changing the title of share popup for annotations. You can find an example [here](https://dev.rebuildingakidney.org/chaise/viewer/#2/Gene_Expression:Image/id=14-47YP)) So you shouldn't blindly just change it. Something like this:

```javascript
if (reference.citation) {
  var title = "and Cite";
  if (params.title) {
    title += " " + params..title;
  }
  params.title = title;
}
```

RFSH (approved):

jrchudy:
Changes to pass in the title to the share popup function were pretty simple so I factored that out and defined a default.

RFSH (approved):

PR #2033, opened 2020-12-03
More Descriptive Head titles

jrchudy:
Modifications to use the heuristics outlined in the attached issue to update the head title for the current page. This required waiting for the reference to be resolved and in some cases, after reading the page data. 

Added tests to each presentation spec for each app to verify these heuristic changes. Also updated 2 navbar specs to make sure `chaiseConfig.headTitle` was being included properly.

Updated spacing in edit mode of recordedit, foreign key popup, and the add pure and binary popup.


jrchudy:
Merged master which should have the fixed test cases now. This is ready for review.

recordedit/recordEdit.app.js
```diff
@@ -253,14 +253,22 @@
...
 
-                            var column, value;
-
+                            var column, value, headTitle;
 
                             // $rootScope.tuples is used for keeping track of changes in the tuple data before it is submitted for update
                             $rootScope.tuples = [];
-                            if (context.mode == context.modes.EDIT && page.tuples.length == 1) {
-                                $rootScope.displayname = page.tuples[0].displayname;
+                            if (context.mode == context.modes.EDIT) {
```
RFSH, line 260: Looking at the code and the way you implemented it, I realized we never considered the copy case.

It seems like in case of copy we use the same page title as create mode (`Create new <table>`) which I think makes sense. So I think you should  also do the same for the head title.

jrchudy: Oh that's a good call. I'll update that as well and add a test case for it.

RFSH (approved):

PR #2049, opened 2021-02-03
GitHub Actions setup

RFSH:
This PR will replace Travis CI with GitHub Actions by removing the `.travis.yml` and creating the equivalent GitHub workflow. 

I also had to change any notion of Travis in our test and documentations. A few things that I don't like and we can improve later:

- We might be able to rely more on docker containers or GitHub marketplace actions. For now I'm manually installing most of the dependencies.
- I had some trouble setting up SauceLabs. With Travis CI, we could use its [sauce_connect](https://docs.travis-ci.com/user/sauce-connect/) feature right out of the box. The same tool [is available in GitHub actions](https://github.com/saucelabs/sauce-connect-action) but I couldn't properly setup the tunnel with it. While running our e2e tests in Travis, SauceLabs creates a machine that talks to the vm that the CI tool is creating. And to make sure it can access the machine, we have to create a "tunnel" with SauceLabs, and then use the same tunnel for our tests as well. This can be done by specifying the same "tunnel-identifier" while creating the tunnel and running test cases. To ensure the usage of same "tunnel-identifier" I had to manually install SauceLabs with GitHub action instead of using the available too.
- Our e2e test is consistent of four different specs. Regardless of failure in each one, we want to run all of them all the times. But as soon as part of a job errors out, GitHub will stop running the remaining commands on the job. To ensure running all the commands you have to use `continue-on-error` but that causes the job to report a successful status. To get around that, I had to add an extra dummy step that reports failure when the test step failed. [The proper support for this is actually one of the requested features](https://github.com/actions/toolkit/issues/399).


RFSH:
@jrchudy The only failing test cases right now are histogram faceting tests. I was never able to run them consistently locally so I'm not entirely sure what's wrong. From what I'm seeing it seems like the expected values are less precise than the actual values, `12.5464` vs `12.546399999999998` for example.  Can you please take a look at them?

PR #2053, opened 2021-02-12
Iframe fullscreen logging

jrchudy:
Changes to add logging to the fullscreen iframe button. I'm not sure what we should use as the log action but what I added seems right.

I think we might want to add something to the logStack that includes the path for the iframe? Maybe also the path to the originating page although that can be inferred from the ids we are attaching.

Also I wasn't sure where to add this information into [this document](https://docs.google.com/spreadsheets/d/1cHhPR0AacvuH2o3QavWlJCIIyPajfn93fJzFnc7VA5M/edit#gid=1180557009). I was thinking it doesn't fit in any existing category since iframes can be a part of any app (record and recordset. even the modal popups in recordedit).


RFSH:
Let's discuss this in a meeting with @hongsudt. Depending on whether we want to capture the stack and/or add another level to it, you would have to make some modifications to this implementation. The current implementation is completely ignoring stack, stack-path, and app-mode. So regardless of where users clicked on the button the action will always be `:;iframe;fullscreen`. Also the last word in the action string is always a verb, so we need to modify the text as well.

jrchudy:
@RFSH this has been updated to include a log stack that communicates what page the action came from. I updated the document as well at the very bottom with it's own section. I wasn't sure where else to put it since iframes can be in any recordset view in any popup.

PR #2054, opened 2021-02-12
Test the session expired alert

RFSH:
This PR will add e2e test for making sure the session expired alert is displayed. To achieve this,

- I had to modify the `performLogin` function that we have to modify the `localStorage`. 
- Since as part of the test I'm performing a logout. I have to make sure anonymous users can see the data. So I made sure test catalogs are readable by anyone. I could make this specific to only one spec, but I feel like this is actually a good change that we should have done a long time ago. We're not doing any testing based on user ACLs so there really is no need to hide the catalogs. That being said, this change might have an effect on CI e2e which I have not anticipated. If that's the case, I have to think about other ways of doing this.
- Added two tests to the errors spec. These two tests open a chaise page (record and recordset), manually remove the cookie, refresh, and check the alert. I'm also logging in after each one to ensure user has all the previous access to data.

P.S. At first I was thinking about doing schema-level ACLs in the test spec. That's why in Slack I mentioned a change in ErmrestDataUtils. But After going through it, the specification and also code was very messy. So I decided against it and now doing the catalog-level ACL.

### DIFF
--- a/test/e2e/data_setup/config/record/links.config.json
+++ b/test/e2e/data_setup/config/record/links.config.json
@@ -1,9 +1,5 @@
 {
-    "catalog": {
-        "acls": {
-            "enumerate": ["*"]
-        }
-    },
+    "catalog": {},
     "schema": {
         "name": "links",
         "createNew": true,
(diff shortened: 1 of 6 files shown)


RFSH:
@jrchudy The actual test case that I added in this PR is only running locally, so the CI build fail is not related to the changes in this PR. I think it's good if you also run the test case in your machine to make sure what I added works.

Since this PR was focused on testing I was fixing some other failing test cases as well. But there's still some test cases that are failing. The `long_text` test case in `ind-facet` spec is constantly failing in CI, but not locally for me. I'm not sure what's wrong with it.

jrchudy:
Sounds good. I saw the build had failed in CI and had planned to run it locally after another day in case you were going to fiddle with the test in CI or not.

jrchudy (approved):
No issues running the test you added locally. Everything looks good to me

Issue #1947, opened 2020-05-08
Disable submit button in recordedit app while loading


RFSH:
Currently, we show the submit button as soon as the `reference` object is created. This is fine in case of create, but it might cause issues in copy and edit.

 In the case of copy and edit, we have to send a request to get the existing entities. Currently, the submit button will be available before this request is done. But that should not be the case and we should disable the button while the data is loading.

PR #2055, opened 2021-02-18
properly test conflict error headers

RFSH:
We recently found out that the header of our error messages are not properly working. On Chrome and Safari, for a `Conflict` error chaise was showing the default `Terminal Error` header. I [fixed this issue in ermrestjs](https://github.com/informatics-isi-edu/ermrestjs/commit/c4e739da22c8b7b16a110c15ec0ef21937aced38) and this PR is just to make sure Chaise is properly testing the header.

The actual change was just two lines, but since I added a new `chaise.page.js` function, I refactored the code a little bit. 

This PR was created mainly to make sure e2e workflow is working as expected.

P.S. I also realized that [this test case](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/delete-prohibited/record/no-delete-btn.spec.js#L74) doesn't work. It's expecting an error while it shouldn't. I modified the given input to make sure it's invalid and test case passes.


RFSH:
As part of this I did #1947 as well since that change only required making sure test cases are passing.

PR #2059, opened 2021-02-24
Change configuration app constant name

jrchudy:
This PR changes the way each app interacts with the configuration app. It changes the constant `appName` to `settings`. It's now an object that has 3 properties (`appName`, `disableHeadTitle`, `disableOnclickBehavior`). Each of these are explained in further detail in `config-app.md`.

Made changes to each individual `app.js` file. I think only navbar app and login app should refrain from modifying the head title and the onclick behavior of links (even thought they relay on specific classes being defined).

Appropriate changes will be made in deriva-webapps to booleansearch, lineplot, and plot app (the three that currently integrate with config.js from chaise).

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -3122,14 +3122,12 @@
             addCanonicalTag();                              // controlled by chaise-config value to turn on/off
             setWindowName();                                // will only update if not already set
 
-            if (!settings.disableHeadTitle) addTitle();     // controlled by settings in config.js
+            if (settings.overrideHeadTitle) addTitle();     // controlled by settings in config.js
 
-            if (!settings.disableOnclickBehavior) {
-                overrideDownloadClickBehavior();            // controlled by settings in config.js
-                overrideExternalLinkBehavior();             // controlled by settings in config.js
-            }
+            if (settings.overrideDownloadClickBehavior) overrideDownloadClickBehavior();    // controlled by settings in config.js
+            if (settings.overrideExternalLinkBehavior) overrideExternalLinkBehavior();      // controlled by settings in config.js
 
-            return addCustomCSS();                          // controlled by chaise-config value to attach or not
+            return addCustomCSS();                          // controlled by chaise-config value to attach
         }
 
         return {
--- a/common/utils.js
+++ b/common/utils.js
@@ -3113,16 +3113,23 @@
 
         /**
          * Will return a promise that is resolved when the setup is done
+         *
+         * NOTE: should only be called by config.js (or equivalent configuration app)
          */
-        function setupHead() {
-            addPolyfills();
-            addCanonicalTag();
-            addTitle();
-            setWindowName();
-            overrideDownloadClickBehavior();
-            overrideExternalLinkBehavior();
-            addBodyClasses();
-            return addCustomCSS();
+        function setupHead(settings) {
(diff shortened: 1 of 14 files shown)


RFSH:
I feel like the names `disableHeadTitle` and `disableOnclickBehavior` are just misleading and confusing as we're not disabling anything. And also just `OnClickBehavior` is not informative enough (we're not doing this for any on-clicks, only some specific cases). So I would suggest flipping this and having specific names for each one,

- `overrideHeadTitle`
- `overrideDownloadClickBehavior`
- `overrideExternalLinkBehavior`

And what about the other functions in the `setupHead`?  Is `addCanonicalTag` for example something that only should be done in chaise apps?

jrchudy:
I wanted to set the configuration flags based on our default setup. By default, we want this functionality to occur, so we shouldn't expect a new app to have to "turn on" functionality. That's similar to the decisions we made with the chaise-config properties. 

This is also something only developers of applications that interface with chaise have to worry about, so it's internal to chaise. The explanation of each property in the config document should be enough to communicate that. If you disagree with both of the above, we can definitely discuss this further. I tried to not make the variable names very lengthy which is why they may be interpreted as misleading or not informative enough.

> And what about the other functions in the setupHead?
I added a comment on the line of each function in the changes that explains which ones should have a setting and which ones shouldn't.

common/utils.js
```diff
@@ -3113,16 +3113,23 @@
...
-            addTitle();
-            setWindowName();
-            overrideDownloadClickBehavior();
-            overrideExternalLinkBehavior();
-            addBodyClasses();
-            return addCustomCSS();
+        function setupHead(settings) {
+            addPolyfills();                                 // needs to be set for navbar functionality (and other chaise functionality)
+            addBodyClasses();                               // doesn't need to be controlled since it relies on .chaise-body class being present
+            addCanonicalTag();                              // controlled by chaise-config value to turn on/off
```
jrchudy, line 3122: > Is addCanonicalTag for example something that only should be done in chaise apps?
I think there are cases that this is purely controlled by the chaise-config. You have to define a chaise-config property to turn it on so calling the function blindly does nothing without the config property.

RFSH:
@jrchudy I understand about the default behavior and how we usually define names to be the opposite of our defaults. In this case, the default is not that important to me, so changing it was the easier option. 

> The explanation of each property in the config document should be enough to communicate 

To some extent yes. But the name itself shouldn't be misleading and should be close to what it actually is supposed to do,

- `disableHeadTitle` literally says "disable head title" but it's not disabling anything. It's just telling chaise to don't override the head title. To me `DontOverrideHeadTitle` is a lot better than `disableHeadTitle` (it's not a thoughtful name and that's why I thought we should just flip it instead of coming up with a name.) 
- `disableOnclickBehavior` sounds like it's disabling any on-click behavior.

I don't have any issues with long names and I actually prefer long names if they avoid this issue. As I mentioned, if we can come up with a better naming, we don't need to flip the behavior and default values.

jrchudy:
I updated the property names based on our discussion about them. Some apps I didn't add `overrideHeadTitle: true` because we hadn't previously talked about how to set the headTitle in those apps. In mdHelp app for example, I think it makes sense to change how the headTitle is being set to prepend the defined value from chaise-config so it read `chaiseConfig.headTitle : Markdown Help`.

RFSH (approved):

PR #2062, opened 2021-03-09
Change behvaior of links in viewer when inside an iframe

RFSH:
### Summary of the issue
Viewer app is currently used in an iframe in the record page. So users can click on the "share" button for any of the annotation and by clicking on the link, the iframe (and not the whole page or new tab) will navigate to the "Image Annotation" table.

This is confusing as users don't have a proper way to get back to the previous page. And also they might think that clicking on "share" in the record page and navigating to that link should show the current state of the page (with "Image Annotation" record being in the iframe and not the viewer app).

### Implementation details

I know that for now we want a solution that is limited to viewer app. But I didn't want to implement this just for the viewer app. So the actual implementation is independent of the app. And then using the config `settings` constant I'm making sure only viewer app is using this.

The general solution was adding a click listened and adding the `blank="_true"` as you can see in `openLinksInTab` function that `setupHead` calls. But this won't support the links in navbar as we've defined a custom `ng-click` for those. And even though navbar is currentlt not visible in the iframe of viewer app in rbk, I wanted my solution to be general enough and you could also have navbar in iframe.

So I had to make some modifications to the `navbar.js` as well. Since `settings` is a constant for the config-app (and not chaise apps), I had to make sure that this variable is stored somewhere in our `$dcctx` object so I can access it in navbar. I added a `$dcctx.settings` object that holds the existing `hideNavbar` and the newly added `openLinksInTab`.  I'm not very good with names and if you thing having two `settings` might be confusing, I can change it. 

To test this you can use the following link:

https://dev.rebuildingakidney.org/~ashafaei/iframe-link.html

### DIFF
--- a/common/config.js
+++ b/common/config.js
@@ -62,7 +62,12 @@
             },
             settings: {
                 hideNavbar: hideNavbar,
-                openLinksInTab: openLinksInTab
+                // the settings constant is not accessible from chaise apps,
+                // therefore we're capturing them here so they can be used in chaise
+                openLinksInTab: openLinksInTab,
+                overrideHeadTitle: settings.overrideHeadTitle,
+                overrideDownloadClickBehavior: settings.overrideDownloadClickBehavior,
+                overrideExternalLinkBehavior: settings.overrideExternalLinkBehavior
             }
         };
         // set chaise configuration based on what is in `chaise-config.js` first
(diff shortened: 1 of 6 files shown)


common/config.js
```diff
@@ -59,7 +60,10 @@
                 pid: MathUtils.uuid(),
                 wid: $window.name
             },
-            hideNavbar: hideNavbar
+            settings: {
```
jrchudy, line 63: Might make sense to store all the "settings" in this object, even if they aren't consumed beyond the config app setup and setupHead function.

jrchudy (approved):
Changes are good. Not sure if it makes sense to store the whole settings object in `$window.settings` or not. Adding `appName` would be redundant since it's in the `$window.contextHeaderParams`.

RFSH:
@jrchudy I think your suggestion makes perfect sense. But instead of just storing the whole settings object, I'm just iterating over what I we want to access in chaise. It's uglier but I think it's the more correct. This would allow more flexibility in what chaise will use. For example in case of this new attribute that I added, I'm doing the iframe check in one place in `config.js`, so I would store `openLinksInTab` in `$window.dcctx` instead of `openIframeLinksInTab`.

jrchudy (approved):

PR #2068, opened 2021-03-23
doc how to use the catalog.id pattern in navbar

robes:
suggested change to doc


jrchudy:
Thank you for pointing this out @robes . I updated `navbar-app.md` as well to include this information (and links to it) since it isn't as clear where all of the explanations and examples for `navbarMenu` are.

PR #2079, opened 2021-04-13
Option to hide empty related tables for writers

jrchudy:
Changes have been made to support a new chaise config property `showWriterEmptyRelatedOnLoad`. This property is added as a boolean property. When set to true, all empty related tables will show on load for writers regardless of permissions to write on those tables. When set to false, all empty related tables will hide on load for writers. When not defined (or anything other than a boolean), we will use the heuristics set based on being logged in and having the ability to create on at least one related table to show them all by default.

Added a test for both `true` and `false` cases.


record/record.app.js
```diff
@@ -278,6 +280,12 @@
                     });
                 });
 
+                // chaiseConfig.showWriterEmptyRelatedOnLoad takes precedence over heuristics above for $rootScope.showEmptyRelatedTables when true or false
+                // showWriterEmptyRelatedOnLoad only applies to users with write permissions for current table
+                if ($rootScope.reference.canCreate && typeof chaiseConfig.showWriterEmptyRelatedOnLoad === "boolean") {
```
jrchudy, line 285: Removed check for `$rootScope.modifyRecord`. This value is based on `chaiseConfig.editRecord`. We don't want that chaise config property affecting how this property behaves. If we want to show all related tables to a writer on load, then having that property set to false shouldn't hide the related tables.

RFSH (approved):

Issue #1958, opened 2020-06-02
Support for Dynamic ACL


svoinea:
It might be useful for `chaise` to handle dynamic ACL in a similar way like `ermrest`.

For example, an user (identified by the column `RCB`), belonging to a group (`https://www.globus.org/app/groups/...`) is allowed to see his own records, but should be allowed to update his own records only if one of its column (for example `status`) has a value specified through a set (for example `[DRAFT,DEPO]`).

Currently, `chaise` gives you access to the `edit` view, and `ermrest` rejects the update request if the user has no permission.

Ideally, `chaise` will prevent to submit an update request if the user has no permission.

karlcz:
I updated an old ermrest issue informatics-isi-edu/ermrest#157  to reflect the API features probably required for this sort of Chaise feature.  @robes   @hongsudt

robes:
@hongsudt this was actually entered by Serban but covers the scenario.

Priority: predicting the edit/delete rights on a per-row basis. For example, not seeing the "Delete" or "Edit" buttons on a record page. Not seeing the "edit" and "delete"/"unlink" controls within the recordset (or embedded recordset) apps.

karlcz:
I've merged the first round of ermrest features which would enable better adaptaiton of the edit/delete and recordedit form activation fields.

See the revised ermrest API docs for details but here's the cheat sheet:

When you GET /ermrest/  you get a service advertisement in JSON format: 
```
{
   ...,
   "features": {
      "trs": true,
      "tcrs": true,
      ...
   }
}
```

when these fields are present and `true` you can try to use the new feature. Being absent means you are talking to an older server and SHOULD NOT attempt the feature as it will cause errors. If it is false you might be talking to a future server that has allowed the feature to be disabled, so also SHOULD NOT attempt it.

The new feature allows new syntax in the /attributegroup/ API list of column projections.  It looks like an aggregate function but requests a rights summary.

- `outalias:=trs(talias:RID)`
- `outalias:=tcrs(talias:RID)`

This adds an output column named by _outalias_ and provides a rights summary for the entity keyed by _talias_ and _RID_.  You should include this in the extra projection list (after the `;`), not in the group-key column list.

The first form is simpler and just provides row-level `update` and `delete` rights info.  This is faster and should be used in recordset listings to allow you to decide whether to show edit/delete controls for each row to this user.

[...]

PR #2081, opened 2021-04-16
Support dynamic ACLs

RFSH:
This PR will add the required changes for #1958. Which are: 
- Changed Reference.canUpdate/canDelete to use Tuple.canUpdate/canDelete where applicable
- Added `canUpdateRows` to recordedit to keep track of columns that can be updated.
- Added a new recordset table to the recoredit result page which would show the "disabled rows". 

Since we didn't have any test cases related to ACL in chaise, I had to make some changes to the test frame work that would allow that. I also added new specs related to testing acl to record, recordset, and recordedit. 

related ERMrestJS issue: https://github.com/informatics-isi-edu/ermrestjs/pull/876


RFSH:
@jrchudy This PR is ready to be reviewed.

While working on the test cases for chaise I found a corner case that we cannot support. And after discussing it with @hongsudt  we decided to show a special alert in that case. 

If there’s a row-based column-level update  acl for a not-null column, recordedit cannot properly handle it in multi-edit case. For example assume there are the following rows in database:

```json
{
  "id": 1,
  "col": "v1",
  "another_col": "another v1"
},
{
  "id": 2,
  "col": "v2",
  "another_col": "another v2"
}
```

Where `col`  is not-null and the user doesn’t have update access to the col  column in row with `id=1`. Now, if users modify the value of `col` in `id=1` row, we have to send the new value of `col`  to ermrest and therefore it should be added the the projection list of the request.  Now,
  - if we add the existing value of `col` for `id=2` to the request, ermrest will complain about the acl.
  - if we skip adding the value of `col`  for `id=2`, ermrest will complain about the not-null requirement.

Technically in this simple example I could completely ignore the row `id=2` since none of its column values has changed. But if the table had other columns that were changed, then we would have to send values for this row as well. So the issue still exists.

In this case, to avoid confusion, we decided to show a warning when users attempt to modify any of the columns.

jrchudy (approved):
Changes look good. Tests passed locally for me.

Issue #2025, opened 2020-10-29
collapsed Menu button doesn't work


crisaless:
I just noticed that if you narrow the width of the browser in Chaise until the menu collapses, the Menu button does nothing when clicked.

Tested on Gudmap/RBK and FaceBase and on Chrome/Firefox/Safari.

More detailed explanation and reasons to fix it : )

When you narrow the browser to less than 768px, the navigation collapses - as designed per Bootstrap - to a Menu button. You should be able to click the button and the menu would appear below. But when I click, nothing happens and usually the word "Menu" disappears.

I did notice that when I go to a chaise page on my iphone - the browser window doesn't get narrower. Instead the page looks zoomed out so you still get a normal view of the navigation, though it's pretty small. So this isn't an issue for a mobile device - and the majority of our users aren't using cell phones.

But a user would definitely have a problem if they wanted to resize the browser window. On my laptop with resolution 1440px x 900px the breakpoint is at about half of my screen, which is a reasonable width someone might use if they want to compare two images side by side, for example.

I can imagine it would not be a simple fix to configure the mobile menu to work well with complex navbar configs like we're doing on staging.gudmap.org - but we do need some kind of workaround.

RFSH:
@kpherwan To provide more context,

- You can test in on any page in rbk-dev or isrd-dev, for example go to [this page](https://dev.rebuildingakidney.org/chaise/recordset/#2/Protocol:Protocol@sort(RMT::desc::,RID)) and you can see the behavior. Just reduce the page width until the "menu" button shows up.
- Related code to navbar are:
  - [`navbar.js`](https://github.com/informatics-isi-edu/chaise/blob/master/common/navbar.js): Where the component is defined and all the js code related to it.
  - [`navbar.html`](https://github.com/informatics-isi-edu/chaise/blob/master/common/templates/navbar.html): Navbar's html template. Navbar is using [`navbarMenu.html`](https://github.com/informatics-isi-edu/chaise/blob/master/common/templates/navbarMenu.html) to implement navbar elements.
- Navbar is using [bootstrap 3's navbar](https://getbootstrap.com/docs/3.3/components/#navbar). And it seems like we didn't properly implement the responsive case. So you would have to most probably change the HTML or JS code to fix the implementation.

kpherwan:
The changes to support this have been merged into master. RBK and Facebase should work as expected though Gudmap has a different way of implementation right now in production. @crisaless can you please review and confirm if changes look good.

[...]

PR #2085, opened 2021-05-24
2025 - Menu button now expands to show vertical navbar

kpherwan:
Issue - https://github.com/informatics-isi-edu/chaise/issues/2025

**Before -** 
Clicking on Menu expands nothing

image

**After -**
Navbar visible

image

**Test link -** 
https://dev.rebuildingakidney.org/~kpherwan/chaise/recordset/#2/Microarray:Series@sort(RID)

https://dev.facebase.org/chaise/recordset/#1/isa:dataset@sort(release_date::desc::,RID)

### DIFF
--- a/common/styles/scss/_navbar.scss
+++ b/common/styles/scss/_navbar.scss
@@ -174,6 +174,7 @@ body > .container {
     overflow-y: auto;
 }
 
+/* This value is in bootstrap for navbar-collapse but it isn't getting applied so its been added here */
 .navbar-collapse {
     max-height: 340px;
 }
--- a/common/styles/scss/_navbar.scss
+++ b/common/styles/scss/_navbar.scss
@@ -163,8 +163,19 @@ body > .container {
     color: #fff;
 }
 
-.child-opened {
-    background-color: $table-header-background-color;
+
+// Horizontal navbar
+@media (min-width: 768px) {
+    .child-opened {
+        background-color: $table-header-background-color;
+    }
+}
+
+// Vertical navbar
+@media (max-width: 768px) {
+    .child-opened > ul {
+        padding-left: 3%;
+    }
 }
 
 /*Adding the scroll bar to the menu */
--- a/common/styles/scss/_navbar.scss
+++ b/common/styles/scss/_navbar.scss
@@ -173,3 +173,7 @@ body > .container {
     overflow-x: hidden;
     overflow-y: auto;
 }
+
+.navbar-collapse {
+    max-height: 340px;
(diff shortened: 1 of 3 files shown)


common/styles/scss/_navbar.scss
```diff
@@ -173,3 +173,7 @@ body > .container {
     overflow-x: hidden;
     overflow-y: auto;
 }
+
+.navbar-collapse {
+    max-height: 340px;
```
RFSH, line 178: Why this specific value? Is this based on RBK's navbar?

kpherwan: This value is in bootstrap for navbar-collapse but it isn't getting applied so I have added it here

RFSH:
The changes look good. Before merging, 

- Test this collapsed version with the default CSS rules hat we have in chaise (don't include any custom css like RBK and Facebase are doing.) And make sure the default UI looks good. For example the default background-color applied to `.child-opened` by bootstrap doesn't work with our navbar (We have a black navbar, so this one most probably should be some shades of grey).

- Test this in RBK and Facebase. It should work without any issues, the reason for testing is making sure the UI looks good. As you can see in different deployments, we're using different CSS rules to customize the navbar and the colors might be off when the navbar is collapsed. You don't need to make any CSS changes in different deployments, just testing it and noting what you think should be changed is enough.

- I think it's good to document the useful css classes and HTML structure of navbar to make the process of customizing navbar easier. We have a document for navbar that you can find [here](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/navbar-app.md). Add a "Customize navbar UI" section before "Notes". You should,

   - Add a minimal representation of HTML structure in there.
   - Explain what are the common classes and their purpose. For example you can mention `navbar-collapse` and explain that the height for the collapsed version is set in there. You don't need to explain every class in there. Just what seems more relevant (How to change color, how to change height, etc)

common/styles/scss/_navbar.scss
kpherwan: Done

kpherwan:
> As Aref mentioned, the default background-color applied to .child-opened doesn't work with our navbar in the vertical setting, changed to show submenu items indented to indicate the expansion.

image

> Added the documentation for "Customize navbar UI" to make it easier for users to find what they would need to change -
https://github.com/informatics-isi-edu/chaise/blob/2025-responsive-menu-bug/docs/user-docs/navbar-app.md

RFSH (approved):

Issue #2071, opened 2021-03-30
Sub-menu opens off-screen


karlcz:
If you get a navbar menu near the right side of the screen, e.g. due to responsive layout on a narrow window, nested sub-menus will still attempt to open to the right and end up clipped by the viewport.

I can recreate this on CFDE by logging in and going to the data review catalog which has a "Submission System" menu towards the right side of the navbar.  Resize the window until the login/username template moves to the next line, then open the Submission System -> Vocabulary sub-menu.  It can appear partially or totally clipped depending on how narrow you make the window.

RFSH:
To provide another example, this can be seen in Facebase as well. Navigate to [this page](https://www.facebase.org/chaise/recordset/#1/isa:dataset@sort(release_date::desc::,RID)) and reduce the page width so that the last option is very close to the edge.

We need to figure out what is the best way to handle this. One obvious solution would be expanding the sub-menu to the left instead right.

kpherwan:
Hi @karlcz can you please verify if the issue is resolved now (it's merged to master). Let me know if something is missing/wrong.

RFSH:
I'm going to close this issue due to inactivity. If this is still not working as expected we should create a new issue for it.

PR #2087, opened 2021-05-29
2071 submenu goes offscreen

kpherwan:
Related to https://github.com/informatics-isi-edu/chaise/issues/2071

Available to test on https://dev.rebuildingakidney.org/~kpherwan/chaise/recordset/#2/Protocol:Protocol@sort(RMT::desc::,RID)

### DIFF
--- a/common/navbar.js
+++ b/common/navbar.js
@@ -27,6 +27,7 @@
             var openDropdowns = document.querySelectorAll(".dropdown.open ul");
             [].forEach.call(openDropdowns, function(el) {
                 checkHeight(el, window.innerHeight);
+                checkWidth(el, window.innerWidth);
             });
         }
     }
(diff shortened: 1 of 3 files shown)


RFSH:
@kpherwan Please add more info to the description. Briefly explain the behavior and how you change it. Summarize the new logic for placement of the popup and what are the corner cases.

kpherwan:
**Problem 1** - The dropdown in navbar or sub-menu of dropdown currently opens off-screen as it always opens towards the right, for example as below - 

image




image

**Solution** - The dropdown will open on the side that has more space 

image


image

**Problem 2** - The text of the dropdown option did not wrap and the longer texts took up a lot of width and hence a wider dropdown.

image


**Solution** - The text wraps after the width of 350px is hit.

image

common/navbar.js
```diff
@@ -83,7 +94,7 @@
           menuTarget.style.top = parseInt((immediateParent.offsetTop + parent.offsetTop) - parent.scrollTop) + 10 + 'px';
       }
 
-      menuTarget.style.left = parseInt(posValues + immediateParent.offsetWidth) + 5 + 'px';
+      menuTarget.style.left = parseInt(posValues + immediateParent.offsetWidth) + 'px';
```
kpherwan, line 97: Removed the additional 5px as the cascading dropdowns showed an unnecessary gap in between currently
 
image

common/styles/scss/_navbar.scss
```diff
@@ -163,29 +165,42 @@ body > .container {
     color: #fff;
 }
 
-
-// Horizontal navbar
-@media (min-width: 768px) {
-    .child-opened {
-        background-color: $table-header-background-color;
-    }
+.dropdown-menu > li >a {
```
kpherwan, line 168: Styling changes to center the arrow, allow text wrap and add a cursor pointer to make it appear clickable.

Before 

image


After

image

RFSH (approved):

Issue #2090, opened 2021-06-11
Navbar issues


kpherwan:
- [ ] The text of the menu option wraps when width gets limited due to the scrollbar i.e. the browser doesn't accommodate width for the scrollbar but instead takes up width of the container/element.

<img width="518" alt="Screen Shot 2021-06-10 at 2 15 36 PM" src="https://user-images.githubusercontent.com/78573171/121604677-d1490880-c9ff-11eb-9a50-c1f10dda25ec.png">

Relevant URLs -

https://stackoverflow.com/questions/5605667/scrollbar-shifts-content

https://stackoverflow.com/questions/1417934/how-to-prevent-scrollbar-from-repositioning-web-page

- [x] In Safari exclusively, the submenu is not visible when the vertical scrollbar is enabled. My understanding is that Safari has some incompatibilities with position:fixed and scollbar. The submenu options are still clickable and inspectable but they are not visible!

Relevant URLs -

https://bugs.webkit.org/show_bug.cgi?id=160953

https://stackoverflow.com/questions/26704903/only-in-safari-positionfixed-child-cut-off-when-parent-is-positionfixed-and/59656203#59656203

jrchudy:
This is fixed in safari with the migration to react. The other issue about text wrapping still occurs

jrchudy:
Both issues still occur. 
 - Depending on the length of the text, the OS/browser will decide to place the scrollbar inside of the element or outside of the element on the right
   - This depends on their OS settings like on mac being able to show scrollbars "Based on using a mouse vs trackpad" which will hide most scrollbars until the element is hovered
   - we can't easily control this and is only a minor inconvenience
     - solution requires dynamic javascript to calculate element sizes and change CSS styles
 - the 2nd issue can be resolved by removing the z-index in safari browsers for the "parent" menu

jrchudy:
The merged pull request fixes the 2nd bullet from the description above. The first issue will be marked as "won't fix". Closing this issue

PR #2095, opened 2021-06-30
Keep table header fixed while scrolling to improve user context of content

kpherwan:
When scrolling all the way to the bottom, the table headings might not be remembered by user so they have been made fixed. 3 scenarios -

1. Recordset
2. Facet 
3. Record

Screen Shot 2021-06-30 at 4 52 15 PM

The CSS way using position:sticky does not work for this scenario since we cannot restrict the height of the table hence we use JavaScript to make the appropriate DOM changes ourselves. 

Note - These changes unfortunately do not work in Safari, the problem is similar to existing problems like mentioned here - https://github.com/informatics-isi-edu/chaise/issues/2090 and a simple fix could not be found.

It can be tested here -
https://dev.facebase.org/~kpherwan/chaise/recordset/#1/isa:experiment@sort(RID)
https://dev.facebase.org/~kpherwan/chaise/record/#1/isa:experiment/RID=1-3SZP

### DIFF
--- a/common/styles/scss/_record-app.scss
+++ b/common/styles/scss/_record-app.scss
@@ -267,6 +267,7 @@
                         padding: 10px 0px 10px 0px;
 
                         .accordion-content {
+                            position: static;
                             padding-left: 40px;
                             padding-right: 8px;
                         }
@@ -419,10 +420,12 @@
           .chaise-table.table {
               border-top: 1px solid #BBBBBB;
               border-bottom: 1px solid #BBBBBB;
+              border-left: 1px solid $table-border-color;
+              border-collapse: initial;
 
               tr {
                     border-left: 1px solid #BBBBBB;
-                    border-bottom: none;
+                    border-bottom: 1px solid #BBBBBB;
                  }
 
               >thead>tr>th,
@@ -438,7 +441,7 @@
               th
               {
                     background-color: unset !important;
-                    border-top: none;
+                    border-top: 1px solid #BBBBBB;
                     border-bottom: 1px solid #BBBBBB;
               }
           }
--- a/common/styles/scss/_recordset.scss
+++ b/common/styles/scss/_recordset.scss
@@ -157,8 +157,6 @@
 // chaise table
 .outer-table {
     overflow-x: auto;
-    height: max-content;
-    max-height: 60vh;
 }
 
 .chaise-table-header {
@@ -187,7 +185,8 @@
 }
 
 .chaise-table.table {
-    border-top: 1px solid $table-border-color;
+    border-collapse: initial;
+    border-left: 1px solid $table-border-color;
     border-bottom: 1px solid $table-border-color;
     margin: 0;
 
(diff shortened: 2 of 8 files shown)


common/styles/scss/_recordset.scss
```diff
@@ -539,3 +544,7 @@ faceting {
 .selected-rows {
     min-height: 45px;
 }
+
+#overflow-container-facet .chaise-checkbox {
+    z-index: 0;
```
RFSH, line 549: I think you've added this because of the z-index:1 that is added to checkboxes in app.scss:587, but I would say the better change is removing that rule. I don't think we need to define a z-index on checkboxes.

common/table.js
```diff
@@ -1477,6 +1477,13 @@
                     UiUtils.attachFooterResizeSensor(0);
                 }
 
+                if (scope.vm.config.displayMode === recordsetDisplayModes.facetPopup) {
+                    var overflowContainerFacet = document.getElementById('overflow-container-facet');
```
RFSH, line 1481: Selecting by id works but is fragile and you shouldn't do this here. We've designed recordset directive so it can be used multiple times on the page without assuming where. If you look at the `attachDOMElementsToScope`, we're creating a `scope.parentContainer` that will return the container of this instance of recordset. So you should use that and then select based on a class:

```js
scope.parentContainer.querySelector("<some selector>")
```

If you do that, then you don't need to do this twice once for popup and once for normal recordset. The newly added `id`s are not needed either.

common/table.js
```diff
@@ -1487,7 +1494,18 @@
...
+
+                var overflowContainer = document.getElementById('overflow-container');
+                if (overflowContainer) {
+                    overflowContainer.onscroll =  fixHeader;
+                }
+
+            };  
+
+            var fixHeader = function(e) {
+                var header = document.getElementById('fixed-table-header');
```
RFSH, line 1506: As mentioned above, change the way you're getting the element. Also comment what this comment is doing.

common/styles/scss/_recordset.scss
```diff
@@ -229,6 +230,9 @@
     }
 
     .table-heading {
+        position: relative;
+        top: 0px;
```
RFSH, line 234: comment why you added this top and z-index.

common/styles/scss/_recordset.scss
```diff
@@ -265,6 +269,7 @@
 
     >thead>tr>th {
         border-bottom: 1px solid $table-border-color;
+        border-top: 1px solid $table-border-color !important;
```
RFSH, line 272: Isn't there a better way to do this? a more specific selector maybe? we should avoid `important` as much as we can.

common/styles/scss/_record-app.scss
```diff
@@ -267,6 +267,7 @@
                         padding: 10px 0px 10px 0px;
 
                         .accordion-content {
+                            position: static;
```
RFSH, line 270: comment why you added this

RFSH (changes requested):

RFSH:
Given that this implementation wasn't performing properly and was jittery specially in record page, we decided to not merge this. So I'm going to close this PR but leave the branch for future references.

More information can be found in #2101.

PR #2097, opened 2021-07-09
1801 google dataset metadata

kpherwan:
Related to PR https://github.com/informatics-isi-edu/ermrestjs/pull/882

- We want the chaise record pages to show up in search results of Google Dataset Search. To facilitate that, we are adding JSON-LD https://developers.google.com/search/docs/data-types/dataset for dataset discovery. 

- The JSON-LD will be based on the annotation called tag:isrd.isi.edu,2021:google-dataset https://github.com/informatics-isi-edu/ermrestjs/blob/1801-chaise-google-dataset/docs/user-docs/annotation.md#tag-2021-google-dataset . 

- A config file called google-dataset-config.js is introduced (similar to chaise-config) and it contains the RIDs of the records that we want to add the JSON-LD to. JSON-LD will only be appended when the RID is in the config (if it exists for the table) and the annotation is setup correctly as well.

### DIFF
--- a/docs/user-docs/google-dataset.md
+++ b/docs/user-docs/google-dataset.md
@@ -1,0 +1,5 @@
+# Google Dataset support
+
+<!-- TODO explain what this is  and what's the benefits of it -->
+
+<!-- TODO explain how to define this by talking about annotation as well as the google-dataset-config.js -->
--- a/docs/user-docs/google-dataset.md
+++ b/docs/user-docs/google-dataset.md
@@ -37,37 +37,48 @@
 2. For more fine grain control of cases where JSON-LD should be added, a config file called [`google-dataset-config.js`](https://github.com/informatics-isi-edu/chaise/blob/master/google-dataset-config-sample.js) is used to specify the cases that should contain the JSON-LD. It can be plugged in similar to chaise-config.js. JSON-LD will only be appended when the row in question is in the config (if it exists for the table) and the annotation is setup correctly as well.  This is an optional step and if no config file exists then the JSON-LD will still be appended if the annotation is setup correctly. Example below:
 
 ```javascript
-var googleDatasetConfigs = {
-    "2": {
-        "Protocol": {
-            "Protocol": {
-                "columns": [
-                    "RID"
-                ],
-                "values": [
-                    "N-H98J","N-H98R","N-H990","N-H99G","N-H996","N-H99J","N-H99P","N-H9A8"
-                ]
-            }
-        },
-        "Common": {
-            "Collection": {
-                "columns": [
-                    "RID"
-                ],
-                "values": [
-                    "Q-3K5E","R-ZD4C","16-26EY","16-QKNG","16-WHS4","16-WMM4","17-DSBR"
-                ]
-            },
-            "Gene": {
-                "columns": [
-                    "RID"
-                ],
-                "values": [
-                    "Q-3TNA","Q-5CPA"
-                ]
+var googleDatasetConfig = {
+    "configRules": [
+        {
+            "host": [
+                "www.gudmap.org"
+            ],
+            "config": {
+                "allowlist": {
+                    "2": {
+                        "Protocol": {
+                            "Protocol": {
+                                "columns": [
+                                    "RID"
+                                ],
+                                "values": [
+                                    "N-H98J","N-H98R"
+                                ]
+                            }
+                        },
+                        "Common": {
+                            "Collection": {
+                                "columns": [
+                                    "RID"
+                                ],
+                                "values": [
+                                    "Q-3K5E","R-ZD4C","16-26EY","16-QKNG","16-WHS4","16-WMM4","17-DSBR"
+                                ]
+                            },
+                            "Gene": {
+                                "columns": [
+                                    "RID"
+                                ],
+                                "values": [
+                                    "Q-4M5E","Q-5G0C","Q-3QBE","Q-46MG","Q-47YR"
+                                ]
+                            }
+                        },
+                    }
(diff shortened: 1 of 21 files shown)


docs/user-docs/google-dataset.md
```diff
@@ -0,0 +1,19 @@
+# Google Dataset Search support
+
+- Google Dataset Search is a search engine from Google that helps researchers locate online data that is freely available for use. Google Dataset Search relies on exposed crawlable structured data via schema.org markup, using the schema.org dataset class https://schema.org/Dataset. In order for our pages to put up in the search results, we need to add markup in the JSON-LD format. 
+
+
+- The JSON-LD will be based on the annotation called tag:isrd.isi.edu,2021:google-dataset https://github.com/informatics-isi-edu/ermrestjs/blob/master/docs/user-docs/annotation.md#tag-2021-google-dataset . Using pattern expansion inside the annotation, the attributes of the JSON-LD will be populated.
```
RFSH, line 6: Use URL syntax (`[caption](url)`)

record/record.utils.js
```diff
@@ -628,14 +628,76 @@
...
+                    }
+                }
+            }
+
+            // use ermrestjs and attach to dom if annotation exists
+            var metadata = $rootScope.reference.googleDatasetMetadata;
+            if (DataUtils.isObjectAndNotNull(metadata)) {
+                metadata = metadata.compute(tuple, $rootScope.templateVariables);
+                if (DataUtils.isObjectAndNotNull(metadata)) {
+                    metadata.url = window.location.href;
```
RFSH, line 683: Does this mean that data-modelers cannot define the `url`? Does url have to be the exact same value as window location? 
We need to discuss about this. We might want to use the resolver instead of window location.

record/record.app.js
```diff
@@ -144,6 +144,9 @@
                     DataUtils.getDisplaynameInnerText(tuple.displayname)
                 );
 
+                // populate the google dataset metadata
+                recordAppUtils.attachGoogleDatasetJsonLd(tuple);
```
RFSH, line 148: Let's move this to be after line that says `$window.history.replaceState({}, '', url);`. If this code is going to use the window.location, we have to make sure it's picking up the correct value.

RFSH (changes requested):
Changes look good. I just suggested some minor changes.

 We also need to discuss about the `url` property and documentation.

record/record.utils.js
```diff
@@ -629,14 +629,75 @@
...
+                DataUtils.isObjectAndNotNull(googleDatasetConfigs[catalogID][schemaName][tableName])
+                ) {
+                var currConfig = googleDatasetConfigs[catalogID][schemaName][tableName];
+                if(currConfig) {
+                    var columns = currConfig.columns;
+                    var allValuesArray = currConfig.values;
+                    var valueInTuple;
+                    
+                    if (Array.isArray(columns) && Array.isArray(allValuesArray) && 
+                    (columns.length == allValuesArray.length || (columns.length == 1 && allValuesArray.length > 0))) {
```
RFSH, line 651: This condition doesn't make sense to me. We could have 
```
{
  "columns": ["col1", "col2", "col3"],
  "values": [ ["col1_1", "col2_1", "col3_1"]]
}
```
Where `columns.length = 3` and `allValuesArray = 1`.  I don't understand why you're checking for something like that. The following is all you need:

```js
// config defined for the table, but the values are invalid: don't add json-ld
if (!(Array.isArray(columns) || !Array.isArray(allValuesArray)) {
  return;
}

var matchFound = allValuesArray.some(function (v) {
  // make sure the value is an array
  var val = Array.isArray(v) ? v : [v];

  // make sure we have value for all the columns
  if (val.length != columns.length) {
    return false;
  }

  // all the values match
  return columns.every(function (col, index) {
    return tuple.data[col] == val[index];
   });

});

// didn't find a match: don't add json-ld
if (!matchFound) {
  return;
}
```

record/record.utils.js
```diff
@@ -629,14 +629,75 @@
...
+            }
+
+            // use ermrestjs and attach to dom if annotation exists
+            var metadata = $rootScope.reference.googleDatasetMetadata;
+            if (DataUtils.isObjectAndNotNull(metadata)) {
+                metadata = metadata.compute(tuple, $rootScope.templateVariables);
+                if (DataUtils.isObjectAndNotNull(metadata)) {
+                    var script = document.createElement('script');
+                    script.setAttribute('type', 'application/ld+json');
+                    script.textContent = JSON.stringify(metadata, null, 4);
```
RFSH, line 686: I know you've added indentation for readability, but if you have a huge JSON-LD wouldn't this cause performance issues? If it can cause performance issues, given that we have online tools that lets you check this object, I don't think we need to worry about readability.

record/record.utils.js
kpherwan: I had misunderstood the config earlier, corrected it now.

RFSH (approved):

RFSH:
Based on the discussion that we had, we realize for our first usecase we also want to limit addition of jsonld based on hostname. So we decided to change the google-dataset-config.js content to following:

```js
var googleDatasetConfig = {
    "configRules": [
        {
            "host": ["www.abc.org", "www.xyz.org"],
            "config": {
                "allowedList": {
                    "<catalog_number>": {
                        "<schema_name>": {
                            "table_name>": {
                                "columns": ["<array of column names>"],
                                "values": ["<array of values>"]
                            }
                        }
                    }
                }
            }
        }
    ]
};
```

RFSH (approved):

PR #2102, opened 2021-07-21
Relative applinks

jrchudy:
Changes to make the appLink function relative. The issue arose when working on the snapshotted static pages for SSR. The pages are being snapshotted on the staging server with data from production. The intention is for the links to be pointed to `www.gudmap.org` and not `staging.gudmap.org` so changing them to be relative means that distinction won't be part of the HTML but instead part of the browser interpreting our relative paths based on the current `location.origin`.

This might not be completely necessary but it will allow more flexibility for our pages outside of the need for SSR.

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -939,6 +939,7 @@
 
             // null or no RID
             if (resolverId === null || !tuple.data || !tuple.data.RID) {
+                // appLink is relative but the display for
                 var url = tuple.reference.contextualize.detailed.appLink;
                 // remove query parameters
                 var lastIndex = url.lastIndexOf("?") > 0 ? url.lastIndexOf("?") : url.length
--- a/common/utils.js
+++ b/common/utils.js
@@ -939,14 +939,14 @@
 
             // null or no RID
             if (resolverId === null || !tuple.data || !tuple.data.RID) {
-                // appLink is relative but the display for
+                // appLink is relative but the display for permalink should be complete
                 var url = tuple.reference.contextualize.detailed.appLink;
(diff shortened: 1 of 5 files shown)


common/utils.js
```diff
@@ -464,7 +464,8 @@
                 appPath = ContextUtils.getValueFromContext(appContextMapping, context);
             }
 
-            var url = chaiseBaseURL() + appPath + "/#" + location.catalog + "/" + location.path;
+            // slice to remove trailing '/' instead of modifying appPath
```
jrchudy, line 467: We previously wrote in a comment that `chaiseDeploymentPath()` will return a leading and trailing `/` character. 

I could update appPath to remove the leading `/` to be consistent with this change but it seems like this is the only case where it's an issue so I modified the string specifically for this case instead.

RFSH:
Based on [google's documentation](https://developers.google.com/search/docs/data-types/dataset) the urls defined in `url` and `description` JSON-LD that we're attaching to chaise record page must be using absolute path. DBAs define these attributes in the [`google-dataset` annotation](https://github.com/informatics-isi-edu/ermrestjs/blob/master/docs/user-docs/annotation.md#tag-2021-google-dataset) and mainly rely on applink logic for generating links to chaise apps. Therefore in this case applink must produce absolute paths and I don't think we should merge this with master.

Based on the discussion that we had, the only reason that we wanted to make the links relative was for the SSR experiment. Given that we're already deploying a different chaise for this purpose, this branch should be merged with the SSR experiment instead.

Issue #1573, opened 2018-06-20
recordset horizontal scroll bar


hongsudt:
People seem to struggle with finding the horizontal scroll bar on the recordset since the horizontal scroll-bar is not always visible.. 

What are our options to address this problem?

jrchudy:
If you have users that see this on linux or windows, we can explore more potential issues but I can't find a simple solution for those cases right now.

karlcz:
To be clear, the issue is not scrollbar showing or not.  It's that the
horizontal scrollbar is offscreen at the bottom of a scrollable
vertical region that is taller than the viewport.  It's very confusing
that the two scrollbars do not stay bound to the viewport and remain
on screen at all times.

jrchudy:
We can do something similar to what was done for `recordedit` which has a scrollbar on the top of the container as well as on the bottom of the container.

kpherwan:
Query -
Does this problem need to be solved for all tables i.e. recordset page, facet(view more), record page  OR exclusively for the recordset table?

Few alternatives - 

1. Keep a fixed max height of the table (possible 60% of viewport height) so the horizontal toolbar would be more easily visible (you don't need to vertically scroll over all the entries to find the horizontal scrollbar)

Pros - vertical scrolling is also reduced over the page i.e. user does not need to scroll through all the rows before moving on to the next table in case of record page or to move on to the pagination controls (endless scrolling reduced)

Cons - We now have two scrollbars next to each other and a slightly busy user interaction.

This is available to play around with here - https://dev.rebuildingakidney.org/~kpherwan/chaise/recordset/#2/Protocol:Protocol

<img width="1438" alt="Screen Shot 2021-06-23 at 6 39 15 PM" src="https://user-images.githubusercontent.com/78573171/123189527-d02dc780-d452-11eb-8913-6e718ede5220.png">


2. As Josh mentioned above, we can do something similar to what was done for recordedit which has a scrollbar on the top of the container as well as on the bottom of the container. This will definitely work for recordset (need to check for facet and record tables)

Pros - Makes it a consistent behavior with recordedit, the scrollbar on top would be latched

Cons - Two scrollbars together are uncommon and the problem of vertical scrolling remains i.e. if we decide to see 100 records, we have to scroll through all of them before we hit the pagination controls or the next table in case of record page)

[...]

PR #2104, opened 2021-07-29
Horizontal scrollbar

kpherwan:
Issue - https://github.com/informatics-isi-edu/chaise/issues/1573

```
/**
         * Some of the tables can be very long and the horizontal scroll only sits at the very bottom by default
         * A fixed horizontal scroll is added here that sticks to the top as we scroll vertically and horizontally
         * Scenarios -
         * 1. Record
         * 2. Recordset
         * 3. Record facer
*/
```

Test link - https://dev.rebuildingakidney.org/~kpherwan/chaise/recordset/#2/Protocol:Protocol@sort(RMT::desc::,RID)

### DIFF
--- a/common/styles/scss/app.scss
+++ b/common/styles/scss/app.scss
@@ -724,19 +724,19 @@ html {
     }
 }
 
-.topScrollWrapper {
+.top-scroll-wrapper, .top-scroll-record-wrapper {
     overflow-x: auto;
     overflow-y: hidden;
     height: 15px;
-    position: fixed;
-    margin-top: -15px;
+    position: sticky;
     width: -moz-available;
     width: -webkit-fill-available;
-    margin-right: 15px;
-    margin-left: 20px;
+    z-index: 2;
+    top: 0;
 }
 
-.topScroll {
+.top-scroll, .top-scroll-record {
+    width: 2000px;
     height: inherit;
     overflow-x: scroll;
 }
--- a/common/styles/scss/app.scss
+++ b/common/styles/scss/app.scss
@@ -727,7 +727,7 @@ html {
 .top-scroll-wrapper {
     overflow-x: auto;
     overflow-y: hidden;
-    height: 15px;
+    height: 0;
     position: sticky;
     width: -moz-available;
     width: -webkit-fill-available;
--- a/common/styles/scss/app.scss
+++ b/common/styles/scss/app.scss
@@ -724,7 +724,7 @@ html {
     }
 }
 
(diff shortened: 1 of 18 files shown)


RFSH:
Per our previous conversation, the previous condition wasn't ideal,

- It was waiting for the initial load and if the initial load was slow, scrollbar would show up and wasn't doing anything.
- Since we're not doing any data request in the result page of multi-create/edit, the scrollbar event wasn't getting attached.

So I changed the condition to be similar to what we have in the template. To make sure we're attaching the event handler as soon as the directive DOM is ready.

Please make sure you've properly tested this in different tables and if it works, we should be able to merge.

P.S. I don't know why a lot test cases were failing in your other commits, let's wait for the latest build and see the result. If a lot test cases failed we have to investigate.

kpherwan:
> Per our previous conversation, the previous condition wasn't ideal,
> 
> * It was waiting for the initial load and if the initial load was slow, scrollbar would show up and wasn't doing anything.
> * Since we're not doing any data request in the result page of multi-create/edit, the scrollbar event wasn't getting attached.
> 
> ~So I changed the condition to be similar to what we have in the template. To make sure we're attaching the event handler as soon as the directive DOM is ready.~
> 
> We actually don't need to have any condition and we can just attach the event handler in the body of `registerTableCallbacks` function. I added a comment in the code to explain this, and also refactored the code a bit.
> 
> Please make sure you've properly tested this in different tables and if it works, we should be able to merge. Also is there any way that we can make the scrolling smoother?



Sorry I accidentally closed the PR, reopened it now.

The refactor made is working perfectly, thank you so much. As discussed, I have made few changes to make the scrolling smoother and tested it on Chrome, Firefox, Safari. Let me know if it still persists.

common/styles/scss/app.scss
```diff
@@ -723,3 +723,20 @@ html {
         line-height: 1.2;
     }
 }
+
+.top-scroll-wrapper {
```
RFSH, line 727: Given that app.css is included in the static pages that have navbar, we have to make sure our rules are properly scoped and also use specific names if we can.

In this case, let's move these two new rules that you added to be in inside the `.chaise-body` wrapper above. Also let's just change the class names that you added to have `chaise-` and also more specific to table, so

- `top-scroll-wrapper` -> `chaise-table-top-scroll-wrapper`
- `top-scroll` -> `chaise-table-top-scroll`

kpherwan: Makes sense, made the necessary changes

RFSH (approved):

PR #2107, opened 2021-08-05
883 - suppress table count if annotation exists

kpherwan:
This PR suppresses the requests for table count if `hideRowCount` is set to true in the annotation `tag:misd.isi.edu,2015:display`

related issue: https://github.com/informatics-isi-edu/ermrestjs/issues/883


test/e2e/specs/default-config/recordedit/add-x-forms.spec.js
```diff
@@ -213,6 +213,7 @@ describe('Record Add', function() {
                     return chaisePage.recordEditPage.getSelectAllPopupBtn(colName).click();
                 }).then(function () {
                     // wait for modal rows to load
+                    browser.wait(EC.visibilityOf(chaisePage.recordEditPage.getRecordSetTable()), browser.params.defaultTimeout);
```
RFSH, line 216: In your other PR you did this by adding a different wait condition
```js
browser.wait(EC.visibilityOf(chaisePage.recordEditPage.getCreateNewButton()), browser.params.defaultTimeout);
```
Which is more accurate? (there's a git conflict because of this change)

RFSH:
I realized that it was easier for me to do the testing myself (instead of explaining where you need to change), so I pushed the test cases for this simple change to this PR. 

Now that we merged your other PR, there's a merge conflict that I think it's better if you resolve. When Github Action is done and you resolved the conflict, we should be able to merge.

RFSH (approved):

PR #2111, opened 2021-08-19
New login

jrchudy:
This PR adds a new login app that mimics the auto close and auto reload after logging in to globus (or other login providers). We are adding some new functionality that uses a new chaise config property, `termsAndConditionsConfig`.

2 new functions were added to `authen.js`, `logoutWithoutRedirect` and `refreshLogin`. `logoutWithoutRedirect` will log the user out without redirecting the page to a post logout page. This is done in the background while the new login app is showing content (only if the user doesn't have the defined group).

`refreshLogin` will log the user in within the same window and change the page to the link provided during login. This link is the same as as the first time the post login flow shows, so the page refreshes then either closes or shows the instructions again (if the user didn't join the group).

_pending chaise-config.md change..._


RFSH (approved):
Overall the changes look good to me. As I mentioned given that we have a deadline coming up I don't think I should review in depth. We don't have the time to go back and forth.

Issue #2118, opened 2021-09-14
Override "My Profile" modal with link


jrchudy:
Allow a way to configure a link that when defined will override the "My Profile" link in the “user name” dropdown in the top right corner. 

Specifically for CFDE, the configuration property needs to be able to use handlebars templating to use the `client.id` value from the session object returned by webauthn. We need to create a link to a record app page keyed by this id value. 

We talked about adding this as either an annotation change or chaise-config. We want this to apply to the user object associated with the whole deployment, so it would need to be something defined at the catalog level. This might also be defined for the whole deployment (multiple catalogs), so I think chaise-config property is what we should do.

We have an existing property called `profileURL` that replaces the whole dropdown with a link to the defined url. I think this removes the ability to logout from the dropdown so doesn't handle our intended purpose. 

Maybe we call this property `myProfileTemplateLink`

RFSH:
The following came up in our group meeting,

- We should make sure we're properly handling error cases and they are consistent with navbar behavior. For example what if `url_pattern` returns an empty string? do we ignore the whole entry? or just act as if `url_pattern` wasn't defined? What if an unknown type is used? etc
- In the future we might want to replace the current profile modal with a record page in modal. If we decided to do that, the current syntax should be able to handle it (we don't need to plan for anything, as long as we can expand later we should be good)

jrchudy:
@hongsudt @karlcz I updated the main body of the issue to be consistent with our discussion from chaise standup. I outlined the inconsistencies that we noticed with `navbarMenu` in this issue and opened issue #2128 to update `navbarMenu` in the future.

jrchudy:
Based on type, determine if entry is valid and remove from array if `menuOption` is invalid.

```
{
  type: "my_profile"
}, { 
  nameMarkdownPattern: "My Profile", 
  urlPattern: "/chaise/record/#registry/CFDE:user_profile/id={{#encode $session.id}}{{/encode}}", 
  type: "url" 
}, {
  type: "logout"
}
```

For the discussion between "children" vs "urlPattern":
```
switch (option.type) {
  case "menu":
    accepts "children"
  case "url":
    accepts "urlPattern"
  case "logout":
    ignore "children", "urlPattern"
  case "my_profile":
    ignore "children", "urlPattern"
  default:
    do nothing
}
```

jrchudy:
Closing this issue now that the changes have been merged into master.

Issue #911, opened 2017-01-05
Cookie not set in RecordEdit when adding multiple entities


shinyichen:
This is the scenario. Click '+' button in RecordSet to open RecordEdit. Create more than one entity and submit. Come back to RecordSet and nothing happens (RecordSet is suppose to have updated the table with the new record). However, this works when creating only 1 entity.

Here is the reason for the bug. This bug was introduced at resolving conflict for `recordedit/form.controller.js`. https://github.com/informatics-isi-edu/chaise/commit/be54746982aad0e8c8fcd3af5417572b4c4054f2. 

Cookie is only set when 1 entity is added, but should be actually be set regardless of number of entities added.

PR #2129, opened 2021-10-06
Login menu config

jrchudy:
This PR should be merged along with [PR #911 in ermrestJS repo](https://github.com/informatics-isi-edu/ermrestjs/pull/911). The changes in this PR are based around adding the chaise-config property `loggedInMenu` to allow for configuration and customization of the login dropdown menu. A brief explanation is below, more info regarding this can be found in the attached issue (#2118 ):
 - Added functionality to the login dropdown to look for the chaise-config property and show the defined `loggedInMenu.menuOptions` if valid. 
    - If `menuOptions` is not defined, the old options ("My Profile" and "Log Out") will show. 
 - New type property is being applied to menuOptions, making this differ from navbarMenu
    - valid each type with simple validations
 - new `subMenu` directive defined to avoid cluttering `navbarMenu` directive with conditional logic
 - created `MenuUtils` factory to move functionality that `navbar`, `navbarMenu`, `login`, and `subMenu` share in common.
 - test cases for login menu to verify invalid options are hidden, my profile and logout can have their names customized, and enable acls are working
 - change each `*.app.js` to call `ERMrest.setClientSession()` to align with changes in the above mentioned ermrestJS PR

_Pending documentation..._


common/login.js
```diff
@@ -6,68 +6,214 @@
...
+                                            } else {
+                                                // acls could be defined with nothing in it, or with only show or only enable
+                                                if (child.acls.show === undefined) child.acls.show = parentAcls.show;
+                                                if (child.acls.enable === undefined) child.acls.enable = parentAcls.enable;
+                                            }
+
+                                            q.push(child);
+                                        });
+                                    }
+                                    option.isValid = MenuUtils.isOptionValid(option);
```
RFSH, line 106: We're going to have to rewrite this in React so you don't need to change this, but I feel like both code and template would have been simpler/more efficient if you were just removing invalid elements. Instead of just setting a boolean that then template removes.

I think we're doing the same thing in navbar.

common/utils.js
```diff
@@ -1190,6 +1191,343 @@
...
+                } else {
+                    $window.location = menuObject.url;
+                }
+            };
+        }
+
+        function renderName(option) {
+            // new syntax will always use nameMarkdownPattern
+            if (option.nameMarkdownPattern) {
+                return $sce.trustAsHtml(ERMrest.renderHandlebarsTemplate(option.nameMarkdownPattern, {inline: true}));
```
RFSH, line 1471: You need to do `renderHandlebarsTemplate` and then `renderMarkdown` (since it's both pattern and markdown).

common/login.js
jrchudy: In navbar, we are not removing the option. We are using ng-if to decide if certain properties are set and show that `<li>` if the condition is met. 

The way we are using a queue to modify each menu option, we only have access to a reference to the menuOption from the array or from the children array. Because of this, we can't simply "delete" that entry if it is invalid, so I decided on a similar approach which is to mark the item as valid/invalid and hide it if it is invalid.

common/login.js
```diff
@@ -6,68 +6,214 @@
...
+                                    menuConfig.acls = {
+                                        "show": ["*"],
+                                        "enable": ["*"]
+                                    };
+                                }
+                                scope.overrideProfile = false;
+                                scope.overrideLogout = false;
+
+                                // iterate over menuOptions and check to see if profile and logout need to be replaced
+                                var q = [menuConfig];
```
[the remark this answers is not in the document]
jrchudy: When I was testing the code, it seemed like the top level was mostly ignored. I initially kept the top level checks for inheriting acls but I can see how it's a bit confusing.

PR #2131, opened 2021-10-11
Allow submission of empty set when facet popup loaded with selections made

jrchudy:
This PR addresses an issue with the facet "Show More" popup not allowing the user to submit a selection of 0 rows when the popup loaded with more than 0 rows. More details can be found in the attached issue.

### DIFF
--- a/common/modal.js
+++ b/common/modal.js
@@ -384,7 +384,12 @@
             switch (params.selectMode) {
                 case modalBox.multiSelectMode:
                     if (params.onSelectedRowsChanged) {
-                        return params.onSelectedRowsChanged(getMultiSelectionResult());
+                        var res = params.onSelectedRowsChanged(getMultiSelectionResult());
+
+                        // if it returns false, then we should disable submit button
+                        // since we cannot apply the change (url limit issue)
+                        vm.disableSubmit = (res === false);
+                        return res;
                     }
                     break;
                 default:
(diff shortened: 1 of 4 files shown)


common/modal.js
```diff
@@ -309,6 +309,7 @@
         function SearchPopupController(ConfigUtils, DataUtils, params, Session, logService, modalBox, recordsetDisplayModes, $rootScope, $timeout, $uibModalInstance) {
         var vm = this;
 
+        vm.rowsOnLoad = params.selectedRows.length > 0;
```
RFSH, line 312: Can you please add a comment explaining why this variable exists? At least I'm sure that I won't remember it by just looking at its name.

RFSH (approved):

common/modal.js
```diff
@@ -383,12 +384,7 @@
             switch (params.selectMode) {
                 case modalBox.multiSelectMode:
                     if (params.onSelectedRowsChanged) {
-                        var res = params.onSelectedRowsChanged(getMultiSelectionResult());
-
-                        // if it returns false, then we should disable submit button
-                        // since we cannot apply the change (url limit issue)
-                        vm.disableSubmit = (res === false);
-                        return res;
+                        return params.onSelectedRowsChanged(getMultiSelectionResult());
```
RFSH, line 387: As we discussed, you should revert this to what it was since we want the button to be disabled when users reach URL length limitation while changing the selected rows in facet popup.

PR #2133, opened 2021-10-15
defaults for saved query name and description

jrchudy:
This PR adds default values into the name and description columns during saved query creation. These values will populate the form when it is opened and can be replaced by other values by the user.

This also fixes an issue with markdown inputs not displaying correctly in the saved query popup that hasn't been encountered yet.

### DIFF
--- a/common/table.js
+++ b/common/table.js
@@ -1570,6 +1570,9 @@
                      *      <facet displayname> (<option1>, <option2>, <option3>, ...)
                      *   3. listing the facet displayname and number of selections if over either of the numFacetChoices or facetTextLength thresholds
                      *      <facet displayname> (6 choices)
+                     * If the name after appending all facet names together is over the nameLength threshold,
+                     * a further shortened syntax is used for the whole name:
+                     *     <reference displayname> with x facets: <facet1 displayname>, <facet2 displayname>, <facet3 displayname>, ...
                      *
                      * `description` begins with the reference displayname and "with" also. The rest of the
                      * description is created by iterating over each facet and the selections made in the facets.
--- a/common/table.js
+++ b/common/table.js
@@ -1534,10 +1534,19 @@
                         return model.column.name == "description"
                     })[0].inputType == "longtext"
 
-                    function facetOptionsToString (options, isRange) {
+                    function facetOptionsToString (options) {
                         var str = "";
                         options.forEach(function (option, idx) {
-                            str += isRange ? (option.min + " to " + option.max) : (option === null ? " _No value_" : " " + option);
+                            var name = option.displayname.value;
+                            if (name === null) {
+                                str += " _No value_";
+                            } else if (name === "<i>All records with value </i>") {
+                                str += " _All records with value_"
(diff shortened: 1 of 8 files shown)


RFSH (changes requested):
If our goal here is to use what users see in the chiclets, then you shouldn't use the raw stable facet. The stable facet is going to give you the "raw values", so it won't apply any formatting and in case of entity modes it will be the raw value instead of rowname.

You should do similar things that we did for chiclets. You can use `vm.reference.facetColumns[i].displayname` to get the displayname of a facet, and `vm.facetModels[i].appliedFilters` to get the list of filters that are applied to that facet. You can see how this is done [here](https://github.com/informatics-isi-edu/chaise/blob/master/common/templates/recordset.html#L94,L118)

common/table.js
```diff
@@ -1527,10 +1527,85 @@
...
+                    var description = nameDescriptionPrefix + ":\n";
+
+                    var modelsWFilters = scope.vm.facetModels.filter(function (fm, idx) {
+                        fm.displayname = scope.vm.reference.facetColumns[idx].displayname.value;
+                        fm.preferredMode = scope.vm.reference.facetColumns[idx].preferredMode;
+                        return (fm.appliedFilters.length > 0);
+                    });
+
+                    if (scope.vm.search) {
+                        name += " " + scope.vm.search;
```
RFSH, line 1573: I think you need a separator between search and the rest.

common/table.js
```diff
@@ -1527,10 +1527,85 @@
...
+                            if (name === null) {
+                                str += " _No value_";
+                            } else if (name === "<i>All records with value </i>") {
+                                str += " _All records with value_"
+                            } else if (name === '') {
+                                str += " _Empty_"
+                            } else {
+                                str += " " + name;
+                            }
+                            if (idx+1 != options.length) str += ", "
```
RFSH, line 1550: This `", "` is causing two whitespaces after comma, is that intentional?

common/table.js
```diff
@@ -1527,10 +1527,85 @@
...
+                        });
+                        return str;
+                    }
+
+                    // thresholds for when to use a modified simpler syntax for the name
+                    var facetChoicesThreshold = 5; // default to 5 choices
+                    var facetTextLengthThreshold = 60; // default to 50 characters;
+                    var nameLengthThreshold = 200; // default to 200 characters
+
+                    var nameDescriptionPrefix = scope.vm.reference.table.name + " with";
```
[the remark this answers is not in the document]
jrchudy: fixed

common/table.js
jrchudy: fixed

jrchudy:
I added chaise config properties for the thresholds, added an explanation in the code for the default name/description, and added chaise config documentation. We can decide on the chaise config property names with hongsuda in chaise meeting.

jrchudy:
Closing this pull request since #2134 also includes changes for saved query functionality and is based off of this branch.

PR #2145, opened 2021-12-04
Change makefile targets to fix Github Actions issues

RFSH:
Currently for testing we use a bunch of npm dependencies that are not needed while building chaise. That's why we listed them as part of `devDepencendies` and this allows us to do `npm install --production` during `make install`, while we have to do `npm install` before testing.

I'm not sure when this happened, but it seems like the behavior of `npm install --production` is not the same as before. If you do `npm install` and then `npm install --production` it will remove all the dev dependencies. That's why the test cases in Github Actions started failing since last week.

So,
- Introduced a `deps-test` Makefile target to install dependencies after doing `make install` in Github Actions.
- Updated `test` Makefile targets to not install anything by default and only run tests. Previously we used to update webdriver as part of `test` target. This was causing issues in Github Actions as we have to use the privileged user for changing the folder, but the `CI` environment variable is not available for the privileged user in Github.
- Introduced a `update-webdriver` that can be used before running test cases.
- Updated documentation based on the changes (I need to update the docs and remove the parts that mention we're updating webdriver automatically. I didn't do it so it doesn't affect the currently running test build)
- Introduced other install targets based on whether it should also copy the chaise-config, or install dependencies.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -30,6 +30,7 @@ jobs:
       - name: Setup the system
         run: |
           node --version
+          npm --version
           sudo timedatectl set-timezone America/Los_Angeles
           export TZ=America/Los_Angeles
           export PATH="$(systemd-path user-binaries):$PATH"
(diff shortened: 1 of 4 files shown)


RFSH:
@jrchudy The PR is ready to be reviewed. I made the changes based on our discussions and cleaned up the documentation/help target to reflect these changes.

PR #2146, opened 2021-12-07
Disable add record and pre filter popups

jrchudy:
Changes in this PR will disable the add button in record app for related tables when the key information for the relationship on the main record is not defined. This PR also adds a "not null" filter to the request made for pure and binary association references so the rows returned in the popup will not have a null value for the key material used as part of the association.

Added test cases for each of the above feature changes.


common/templates/record.html
```diff
@@ -15,9 +15,10 @@
...
                     toggle-display-mode='toggleInlineDisplayMode({dataModel:col.tableModel})' can-edit="canEditRelated({ref:col.tableModel.reference}) && !showError($index)"
-                    can-create="canCreateRelated({ref:col.tableModel.reference})" show-toggle-display-btn="allowInlineTableMarkdown($index)"
+                    can-create="canCreateRelated({ref:col.tableModel.reference})" can-create-disabled="canCreateRelatedDisabled({ref:col.tableModel.reference})" show-toggle-display-btn="allowInlineTableMarkdown($index)"
                     add-related-record="addRelatedRecord({tableModel:col.tableModel})" to-recordset="toRecordSet({ref:col.tableModel.reference})"
-                    displayname="col.column.displayname" base-table-name="col.baseTableName">
+                    displayname="col.column.displayname" base-table-name="col.baseTableName"
+                    keyset="col.tableModel.reference.derivedAssociationReference ? col.tableModel.reference.derivedAssociationReference.origFKR.key.colset.columns : col.tableModel.reference.origFKR.key.colset.columns">
```
RFSH, line 21: You don't need to check for `derivedAssociationReference`. `origFKR` will be available in `col.tableModel.reference` in both one-hop related and p&b.

It won't be available in the free-form cases but given that we're not showing the "add" button in those cases, it shouldn't matter.

recordedit/form.controller.js
```diff
@@ -310,7 +310,18 @@
 
             params.displayMode = vm.editMode ? recordsetDisplayModes.foreignKeyPopupEdit : recordsetDisplayModes.foreignKeyPopupCreate;
 
-            params.reference = column.filteredRef(submissionRow, vm.recordEditModel.foreignKeyData[rowIndex]).contextualize.compactSelect;
+            var andFilters = [];
+            // loop through all columns that make up the key information for the association with the leaf table and create non-null filters
+            column.foreignKey.key.colset.columns.forEach(function (col) {
```
RFSH, line 315: I don't know why I assumed this PR only includes the related table changes and missed this. If we're doing this for foreign key popup, you should also change the `inputSwitch` implementation to do the same for foreign key popups.

PR #2152, opened 2022-01-19
Pb batch unlink

jrchudy:
This PR adds a button to record app for pure and binary associations to remove many related rows at once. The flow of behavior is as follows:
 1. User clicks "Remove records".
    - A multi select modal with currently associated rows is shown.
 2. User selects rows to remove and clicks "Remove".
    - confirm delete dialog may show with "remove" text used instead of "delete"
 3. The selected rows are deleted and a "Batch Remove Summary" dialog is shown.
    - information about how many successful and failed rows is provided
 4. Upon Summary dialog close, the remove popup is still shown with the rows updated to reflect rows that were removed
    - If rows failed, those rows will remain selected
 
Test cases for this feature were also added to verify batch unlink can occur correctly without any special row permissions. Other tests added only allow 1 specific row to be removed, this test verifies the error case where if some rows fail in a request, all rows fail. Then the test goes on to verify the summary content and that the rows remain selected after the failed delete request.


common/table.js
```diff
@@ -2255,6 +2255,12 @@
                     );
                 }
 
+                scope.showAddRecord = function () {
```
RFSH, line 2258: I think you were meant to use this for the `ng-if` of the "create" button in `tableHeader.html`, but it's not used any where. Instead it seems like you're just doing it with css for unlink popup. 

So I think you should just remove the CSS rule and make sure this function is used in `tableHeader.html`

common/templates/ellipsis.html
```diff
@@ -26,7 +26,8 @@
         <button ng-if="edit" type="button" class="edit-action-button chaise-btn chaise-btn-tertiary chaise-btn-link icon-btn" ng-click="edit()" target="_blank" tooltip-placement="bottom" uib-tooltip="Edit">
             <span class="chaise-btn-icon glyphicon glyphicon-pencil"></span>
         </button>
-        <button ng-if="unlink" type="button" class="delete-action-button chaise-btn chaise-btn-tertiary chaise-btn-link icon-btn" ng-click="unlink()" tooltip-placement="bottom" uib-tooltip="Unlink">
+        <!-- TODO: should this be "config.deletable && unlink && associationRef.canDelete" || "unlink" -->
```
RFSH, line 29: I think we're defining `scope.unlink` if these conditions that you mentioned here are met. Right?

record/record.controller.js
```diff
@@ -450,6 +456,150 @@
...
+
+                    // Show modal popup summarizing total # of deletions succeeded and failed
+                    response.clickOkToDismiss = true;
+                    var reloadModalRows = function () {
+                        if (response.failedTupleData.length > 0) {
+                            // iterate over the set of successful ids and find them in selected rows, then remove them
+                            response.successTupleData.forEach(function (data) {
+                                // data is an object of key/value pairs for each piece of key information
+                                // { keycol1: val, keycol2: val2, ... }
+                                // TODO: what to do with multiple keys?
```
RFSH, line 514: You should handle composite keys. You just need to change your `findIndex` to call a `every` (or `some`) on all the key values. Something like this

```js
searchPopupTableModel.selectedRows.findIndex(function (tuple) {
  return Object.keys(data).every(function (key) {
    return data[key] == tuple[key];
  });
});
```

And to be honest given that you're passing the `selectedRows` to the function, you could just return the actual given `Tuple` object without any modification. Then you could either do object comparison or rely on `Tuple.uniqueId` API for comparison. That being said, what you're passing from ermrestjs to chaise is good enough for now.

record/record.controller.js
```diff
@@ -450,6 +456,150 @@
...
+                }
+
+                if (!CONFIRM_DELETE) {
+                    return leafReference.deleteBatchAssociationTuples(tableModel.parentTuple, tuples).then(deleteResponse).catch(deleteError);
+                }
+
+                // log the opening of the confirm delete modal
+                logService.logClientAction({
+                    action: logService.getActionString(logService.logActions.UNLINK_INTEND),
+                    stack: logService.getStackObject()
```
RFSH, line 555: You should use the `logStack` that you created above. `logService.getStackObject()` will return the stack object attached to the `$rootScope` which doesn't include the related entity data.

common/templates/ellipsis.html
jrchudy: I meant to review this statement and check if that's what it was doing.

RFSH (approved):

Issue #2162, opened 2022-03-24
Viewer app: Check dynamic ACLs for delete annotation feature


RFSH:
As it was described in [this issue](https://github.com/informatics-isi-edu/deriva-webapps/issues/125), the delete annotation feature is not checking the dynamic ACLs and offers the "delete" button as long as the user has static `delete` access to the table.

We should ensure the viewer app uses the proper ERMrest APIs to fetch the dynamic ACL and disables the button if users don't have the required access.

PR #2166, opened 2022-04-01
Proper dynamic ACL support in viewer app annotation list

RFSH:
When I added support for dynamic ACLs in chaise apps, I didn't properly do so for the annotation list. This PR will ensure that we're properly showing "edit" and "delete" buttons in the annotation list based on dynamic ACLs.

While testing this I encountered a case where user can "delete" an annotation but cannot "edit" is. The current UI doesn't properly support that since the "delete" button is only available in the "edit form".

issue: #2162

### DIFF
--- a/chaise-config-sample.js
+++ b/chaise-config-sample.js
@@ -16,7 +16,6 @@ var chaiseConfig = {
         enable: ["*"] // [] <- disable
     },
     maxColumns: 6,
-    defaultAnnotationColor: 'red',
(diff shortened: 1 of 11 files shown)


RFSH:
The latest commit should address the issue that I mentioned in the description. We decided to add a "delete" button to the annotation list. Since we now have four icons in the list, I fixed the spacing to work the same as toolbars in the related section of the record page (the buttons stay on the upper right corner, and the text will wrap around).

Issue #2161, opened 2022-03-10
Configure the display state of filter panel in popups


jrchudy:
Chaise currently loads all search popups with the filter panel hidden completely or collapsed by default. We want to give data modelers control over the original state when the filter panel is available but collapsed on load. 

Different places that we might show faceting:
 - Recordset
   - facet panel open by default
   - context = `compact`
 - Recordset - show saved queries
   - no facet panel
   - context = `compact/select`
 - Facets "Show More" (in every place we have faceting)
   - no facet panel
   - context = `compact/select`
 - Record - add pure & binary
   - facet panel closed by default
   - context = `compact/select`
 - Record - unlink pure & binary
   - facet panel closed by default
   - context = `compact/select`
 - Recordedit - foreign key popup
   - facet panel closed by default
   - context = `compact/select`

The following are the proposed configuration changes.

#### 0. Chaise Config property (force open)
Allow the data modeler to define this property in the chaise config which could be applied to the whole deployment or an individual catalog.

This property allows us to maintain the functionality of the heuristics and not change the default when only closed panels want to be forced open:
```
"forceFacetPanelOpen": true 
```

NOTE: `false`, `null`, or `undefined` meaning don't change anything or keep the current heuristics

#### 1. Chaise Config property
Allow the data modeler to define this property in the chaise config which could be applied to the whole deployment or an individual catalog.

For configuring open or closed everywhere in the application that filtering is allowed (all of the above except facets "Show More"):
```
"facetPanelOpen": true | false | null
```
NOTE: `null` (or `undefined`) meaning don't change anything or keep the current current heuristics

The following are all the contexts that are used for displaying tables of data, `compact`, `compact/brief`, `compact/brief/inline`, and `compact/select`. But only `compact` and `compact/select` contexts allow for faceting in chaise. Alternative solution to configure (override) individual contexts:
```
"facetPanelOpen": ["compact/select"] | ["*"] | []
```

NOTES: 
 - Defining "compact" as a context wouldn't be necessary since recordset has the facet panel open by default and this property will open it for the defined contexts. 
 - "*" means the same thing "compact/select". 
 - `[]` would have the facet panel closed by default for both `compact` and and `compact/select`. 
 - Inheriting the value from `compact` for `compact/select` should not be done since that goes against our current heuristics.

#### 2. Part of other annotations
[...]

PR #2168, opened 2022-04-11
Compact select association

jrchudy:
This PR adds a chaise config property to allow for control over the initial display of the facet panel in search popups on modal load. The property includes 2 arrays, open and closed, of contexts related to compact and its sub-contexts.

To determine the state of the facet panel, chaise will get the value from ermrestJS which calculates the value based on the context of the reference and the defined `facetPanelDisplay` value.

Other changes include using more specific "contexts" when contextualizing a reference in places in chaise. The only reference that hasn't been changed was the reference used for the `Show More` popup in faceting.

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -41,10 +41,7 @@
           "showWriterEmptyRelatedOnLoad": null,
           "savedQueryConfig": null,
           "loggedInMenu": {},
-          "facetPanelDisplay": {
-              "open": ["compact"],
-              "closed": ["compact/select"]
-          },
+          "facetPanelDisplay": {},
           "shareCiteAcls": {
               "show": ["*"],
               "enable": ["*"]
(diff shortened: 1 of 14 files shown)


common/utils.js
```diff
@@ -41,6 +41,9 @@
           "showWriterEmptyRelatedOnLoad": null,
           "savedQueryConfig": null,
           "loggedInMenu": {},
+          "facetPanelDisplay": {
+              "open": []
```
RFSH, line 45: This should be default that we want, not just empty open list.

RFSH (approved):

jrchudy:
This PR resolves issue #2161 . The changes we made are summarized in [this comment](https://github.com/informatics-isi-edu/chaise/issues/2161#issuecomment-1068593346).

PR #2170, opened 2022-04-15
Decorate session attributes

jrchudy:
Update the attributes array in the session object when it is returned in a response from webauthn instead of in each place the attributes array is being consumed. We were doing this in 2 separate places, once in chaise for the profile dialog and again in ermrestJS for the templating environment.

Change the profile modal dialog to only show the properties of the client we want to show to users instead of any properties that might be part of the client. This is to prevent `identities` from showing and later `extensions` when the changes are available from webauthn.


RFSH (approved):
It looks much nicer now. Instead of doing this in multiple places, we're now decorating the session object from source and everywhere else is just using it.

PR #2171, opened 2022-04-18
upgrade moment and protractor + fix test cases

RFSH:
The old version of both dependencies had some vulnerability issues.


RFSH:
When we merged the `$session` changes we forgot to update chaise test cases to use the new format. So as part of this PR I also updated those.

Issue #2061, opened 2021-03-08
Color picker displays null values as "black"


ljpearlman:
The color picker appears to be displaying null values as "black". This is most easily demonstrated on a column defined as non-null with no default value. If you try to create a record, the value you'll see displayed for that column is "black". If you then try to save that new record without changing the color, you'll get a "null value in [column] violates non-null constraint" error.

PR #2173, opened 2022-04-20
Handle empty input for colors like any other input

RFSH:
Whether the input value is required or not (based on the `nullok` flag in the model), Chaise allows users to "clear" the input. And the "required" check is only done when users attempt to submit the form.

This was not the case for the color inputs. If a color typed column had `nullok=false`, Chaise wouldn't allow users to clear the input. As a result, the input would show the "black color" on load, while the actual saved value is still `null`. This would lead to a confusing UX, as it is described in [#2061](https://github.com/informatics-isi-edu/chaise/issues/2061). 

So in this PR, I just made sure to allow empty values in the color input and then let Angular handle the "required" check.

### DIFF
--- a/common/inputs.js
+++ b/common/inputs.js
@@ -317,7 +317,6 @@
             require: '?ngModel',
             scope: {
                 toggleCallback: "=?",
-                isRequired: "="
             },
             link: function (scope, elem, attrs, ngModel) {
                 if (!ngModel) return;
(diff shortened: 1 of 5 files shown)


RFSH:
When I first added the test case for the required colors, I saw a terminal error like the one we saw in float and foreign keys (the one that is happening because of an unhandled rejection). 

It took me a while to finally figure this out. It still doesn't make sense to me, but after debugging it, I realized that the test case right after the new one that I added is failing. I'm referring to the test case that ensures invalid color values are ignored. For some reason the `getAttribute("value")` was throwing an error. Doing this test in a wait statement (instead of directly testing it) solved the issue.

@jrchudy The changes and also the modified test case are very simple, but I assigned you as a reviewer so you also run the test cases locally and see if you're seeing any weird behavior.

jrchudy (approved):

PR #2182, opened 2022-06-01
Navbar > HTML structure

ravish-kg:
1. Implemented scroll behavior to all navbar menus and submenus.
2. Fixed spacing issues in the navbar.
3. Fixed Responsiveness issue.
4. Aligning right and left based on available screen width is fixed.


src/assets/scss/_navbar.scss
```diff
@@ -268,6 +269,10 @@ nav.navbar {
         // overflow-y: auto;
         width: max-content;
         max-width: 450px;
+        
+        // setting max height for submenus
+        max-height: 50vh;
```
RFSH, line 274: This is not acceptable and has to be dynamic. Make sure you're replicating the existing behavior which is setting the height based on the available space. The scrollbar should only appear when there's a need for it.

This solution is not dynamic and doesn't work for very long submenus.

If you're using mac, go to "System preferences > General" and then make sure "Show scroll bars" is set to "Always". That way you can see that your solution is always showing the placeholder for scrollbar, which is not desirable.

src/assets/scss/_navbar.scss
```diff
@@ -328,16 +333,18 @@ nav.navbar {
         top: 0;
         right: 100%;
         left: auto;
-        margin-top: 0;
-        margin-right: 0.125rem;
+        // margin-top: 0;
+        // margin-right: 0.125rem;
+        margin: -1px;
```
RFSH, line 338: What the reason for this? Why `-1` and not zero?  Add a comment explaining why you've set it to `-1`. And please don't just comment the rules that you think should be removed and remove them instead.

src/assets/scss/_navbar.scss
```diff
@@ -328,16 +333,18 @@ nav.navbar {
...
 
     .dropend>.dropdown-menu[data-bs-popper] {
         top: 0;
         right: auto;
         left: 100%;
-        margin-top: 0;
-        margin-left: 0.125rem;
+        // margin-top: 0;
+        // margin-left: 0.125rem;
+        margin: -1px;
```
RFSH, line 347: The same comment about removing code and commenting the reason applies here.

src/components/login-dropdown.tsx
```diff
@@ -26,6 +26,43 @@ const ChaiseLoginDropdown = ({
 }: ChaiseLoginDropdownProps): JSX.Element => {
   const dropdownWrapper = useRef<any>(null); // TODO: type the useRef wrapped element
 
+  const [fromTop, setFromTop] = useState<number>();
+  const [fromLeft, setFromLeft] = useState<number>();
+  const [dropEnd, setDropEnd] = useState<boolean>(true);
+
+  const alignDropDown = (event) => {
```
RFSH, line 33: Add a jsdoc style comment to this function explaining what it does and its logic.

RFSH (changes requested):
Before creating a PR please make sure you've tested every scenario. None of the subtasks are fully done and you can see the inconsistency with master in all of them. As we mentioned our goal is to replicate the same behavior or in some cases improve it. We're not trying to just implement the bare minimum here. 

Other than the changes that I mentioned directly in the code,
- The "Log in" text must be to the right of "Go to RID" input. I'm assuming you rearrange it to fix the responsive behavior, but this change is not acceptable.
- The responsiveness is still broken
  - The "Go to RID" and "log in" text go beyond the navbar background. While in master, we're showing proper scrollbars.
  - The submenus must be displayed as a list and not as popups. Or if this is not doable with new bootstrap, you need to propose a workable solution. Currently the submenus are displayed on top of each other which is not acceptable.

src/assets/scss/_navbar.scss
ravish-kg: Fixed the issue

src/components/login-dropdown.tsx
ravish-kg: Added comments above these statements

src/components/login-dropdown.tsx
```diff
@@ -25,6 +26,71 @@ const ChaiseLoginDropdown = ({
...
+    const winHeight = windowRef.innerHeight;
+    const padding = 15;
+
+    if (subMenuRef && subMenuRef.current) {
+      // This logic will run when there is no resize event, but called on click of menu item
+      const y = subMenuRef.current.getBoundingClientRect().y;
+      const available = winHeight - y;
+
+      subMenuRef.current.style.maxHeight = available - padding + 'px';
+    } else {
```
RFSH, line 63: Why do you need this else case here?

RFSH (changes requested):
Focus on addressing the comments and finishing what you've already implemented. After that's done, you also need to fix the responsiveness like I mentioned before.

src/components/login-dropdown.tsx
```diff
@@ -25,6 +26,71 @@ const ChaiseLoginDropdown = ({
...
+
+  // state variable to define left, top, dropEnd or dropStart positions...
+  const [fromTop, setFromTop] = useState<number>();
+  const [fromLeft, setFromLeft] = useState<number>();
+  const [dropEnd, setDropEnd] = useState<boolean>(true);
+
+  useLayoutEffect(() => {
+    
+    // when resize event is called, it will call debounce function with 500ms timeout
+    const debouncedFunc = debounce(setHeight, 500);
```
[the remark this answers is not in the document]
ravish-kg: Since it is resize event which can trigger lot of events at the same. So, delaying the function call for 500ms

src/components/login-dropdown.tsx
ravish-kg: When multiple submenus are open on the window, and resize event happens, it should calculate height of all submenus & update.

src/assets/scss/_navbar.scss
```diff
@@ -268,6 +271,9 @@ nav.navbar {
         // overflow-y: auto;
         width: max-content;
         max-width: 450px;
+        
+        // setting max height for submenus
+        overflow-y: scroll;
```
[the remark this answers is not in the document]
ravish-kg: Addressed this issue in the latest commit

src/components/login-dropdown.tsx
```diff
@@ -12,6 +12,7 @@ import {
   onLinkClick, renderName
 } from '@isrd-isi-edu/chaise/src/utils/menu-utils';
 import { windowRef } from '@isrd-isi-edu/chaise/src/utils/window-ref';
+import { debounce } from '../utils/ui-utils';
```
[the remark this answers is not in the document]
ravish-kg: Addressed this issue in the latest commit

src/components/login-dropdown.tsx
```diff
@@ -25,6 +26,71 @@ const ChaiseLoginDropdown = ({
...
+    // when resize event is called, it will call debounce function with 500ms timeout
+    const debouncedFunc = debounce(setHeight, 500);
+    
+    // First time it will be called diretly
+    setHeight();
+
+    window.addEventListener('resize', debouncedFunc);
+  })
+
+  const setHeight = () => {
```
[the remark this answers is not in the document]
ravish-kg: Added comments, please let me know if anything is missing

src/components/login-dropdown.tsx
RFSH: If the “else”  is handling all the submenus, then what’s the point of the “if” case? and if it’s general and for all the submenus, it should go to the navbar.tsx file and the selector should be specific to only navbar submenus and not all the dropdown menus

RFSH:
I don't understand. Do you even test your changes?

- Like I repeatedly mentioned, the "Log in" text must be to the right of "Go to RID" input.
- While you fixed the scrollbar issue of the collapsed menu. It's still broken for the submenus. Take a look at how master behaves:
  - The submenus should have the same background (not white)
  - There shouldn't be an extra scrollbar for the submenus
  - Submenus should be indented 

   I understand that since bootstrap is not using list elements these are behaving differently. But this doesn't mean you can ignore the broken UI. I expect you to come up with a proper solution to fix this broken UI.
- Saying that you cannot find a cleaner solution for the scroll is not acceptable. At least try to clean up the existing one instead of just copy/pasting it. The current solution of changing the height of every open dropdown menu in the login-dropdown.tsx component is not acceptable.

src/components/login-dropdown.tsx
```diff
@@ -17,15 +17,99 @@ interface ChaiseLoginDropdownProps {
   menu: MenuOption[],
   openProfileCb?: MouseEventHandler,
   parentDropdown: any // TODO: useRef wrapper type
+  alignRight: boolean
```
RFSH, line 20: You don't need to worry about the existing props, but please add a comment explaining what this prop is

```
/**
 * Explain its purpose and how its used
 */ 
alignRight: boolean
```

RFSH (changes requested):
Thank you for the recent changes. It looks much better now and based on my testing it works as expected (and in some cases better than master).

I asked for a small documentation change. And please make sure to modify the `chaise-dropdown-header` rule like I mentioned in the slack message.

RFSH (approved):

jrchudy (approved):

PR #2184, opened 2022-06-02
resizable-filter-panel

anmol-singh-usc:
Created the resizable component to implement the resizable filter component for the recordset app

Component Usage:
    This component is used to implement a horizontal draggable and resizable component/div

Props:

    /**
     * a function that returns the left-pane component in the resizable layout
     * pass the leftRef as a ref to the outermost element of your left-pane component
     */
    left: ((ref: React.RefObject<HTMLDivElement>) => JSX.Element),

    /**
     * the right-pane component in the resizable layout or a function that returns the right-pane component
     */
    right: JSX.Element | (() => JSX.Element),

    /**
     * use for applying custom styles on the component
     */
    className?: string,

    /**
     * the minimum width limit on the left-pane component
     * defaults to 250px
     */
    minWidth?: number,

    /**
     * the maximum width limit on the left-pane component
     * defaults to 450px
     */
    maxWidth?: number,

    /**
     * the initial width for the left-pane component
     * defaults to 270px
     */
    initialWidth?: number,

    /**
     * set this flag in case you pass maxWidth in vw units instead of px units
     * defaults to false
     */
    convertMinWidth?: boolean,

    /**
     * set this flag in case you pass minWidth in vw units instead of px units
     * defaults to false
     */
    convertMaxWidth?: boolean,

    /**
     * set this flag in case you pass initialWidth in vw units instead of px units
     * defaults to false
     */
    convertInitialWidth?: boolean

Note:
    To resize other components based on the resized value left-pane component add an event listener and listen to the 
    'resizable-width-change' event.

    For usage refer to the recordset.tsx component


jrchudy:
Regarding the functionality, there are 2 things I noticed:
 - the min and max width of the facet panel is inconsistent with master
   - The min-width is being defined as `170px` on the element with class `side-panel-resizable`, but the resizable component isn't allowing me to squeeze the content below `250px`.
   - The max-width is being defined as `40vw` on the element with class `side-panel-resizable`. When testing locally with 2 browsers side by side (both are the same size), on master I can stretch the facet panel content to that `40vw` which gives me a width of `~568px`. It seems the resizable component isn't allowing me to go above `450px`.
     - Another test I did was to reduce my windows to an even smaller size. With the smaller size, master limited me to `379px` width. I can see the same happening in your react branch.
     - It seems like the react component itself is imposing a max/min width for itself and not honoring the styles we applied to the element with class `side-panel-resizable`
 - The divider should display the same as master. Currently it looks like the react component is adding a `div` with class `divider` that is showing that grey line. We should instead use `fa-ellipsis-v` from font awesome. 
    - Make sure the icon is centered with respect to the "bottom-panel-container"

jrchudy (changes requested):
Please see my comment above for details about requested changes.

src/assets/scss/_resizable.scss
```diff
@@ -0,0 +1,17 @@
+.divider-hitbox {
+    cursor: col-resize;
+    align-self: stretch;
+    display: flex;
+    align-items: center;
+    padding: 0 1.5rem 0 0.4rem;
+    top: 50%;
+    color: #ccc;
```
jrchudy, line 8: `color: #bbb;` is defined after this style and should be the one we use. Let's remove `color: #ccc`

src/assets/scss/_resizable.scss
```diff
@@ -0,0 +1,17 @@
...
+    font-size: 2em;
+    color: #bbb;
+    left: 2px;
+}
+
+.divider {
```
jrchudy, line 14: This class seems to be removed now, no need to keep the rule in the CSS.

jrchudy (changes requested):
I see why you initially wrote the min/max/initial as expecting a value that will be appended with `px` since you are doing calculations to set the width. I think how you are doing it is fine but we should talk to Aref about it as part of the final review.

Looking at the implementation, looks like the initial width is still set to `270px`. We should have this initialized to `21vw`. Another thing I noticed is that some styles are present in `app.scss` that you are also overriding with your component. Please check how you are setting the initial, min, and max width and determine whether the styles in `app.scss` should be removed, or if your implementation should be adjusting to use those styles.

Looking at your changes for `_recordset.scss`, it's hard to tell what you actually changed and what was just formatting changes. 

One other UX issue I'm still seeing is how the "Refine search" text and "Hide panel" button adjust their positioning when the panel is almost fully collapsed. We want "Refine search" to remain on one line and go above "Hide panel". On master, instead of using flex, we attached `float: left` to "Refine search" and `float: right;` to "Hide panel".

src/assets/scss/_resizable.scss
anmol-singh-usc: fixed

src/assets/scss/_resizable.scss
anmol-singh-usc: fixed

src/components/resizable.tsx
```diff
@@ -0,0 +1,164 @@
...
+    setLeftWidth: (value: number) => void
+};
+
+const LeftPane = ({ children, leftWidth, setLeftWidth }: LeftPaneProps): JSX.Element => {
+    const leftRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+        if (leftRef?.current) {
+            if (!leftWidth) {
+                setLeftWidth(leftRef.current.clientWidth);
```
RFSH, line 47: Is there any way that we can avoid passing this setter function? Can't we assume `leftWidth` will always have a default value?

src/components/resizable.tsx
```diff
@@ -0,0 +1,164 @@
...
+    };
+
+    const onMouseUp = () => {
+        setDragging(false);
+    };
+
+    const onMove = (clientX: number) => {
+        if (dragging && leftWidth && separatorXPosition) {
+            const newLeftWidth = leftWidth + clientX - separatorXPosition;
+            setSeparatorXPosition(clientX);
```
RFSH, line 122: I'm noticing some jittering while resizing the panel. Could the reason be the fact that you're setting two state variables and you're not in control of their batching?

In general, I think it might be better to merge these two state variables into an object.

src/components/resizable.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+Note:
+    To resize other components based on the resized value left-pane component add an event listener and listen to the 
+    'resizable-width-change' event.
+
+    For usage refer to the recordset.tsx component
+*/
+
+
+const vwToPx = (value: number) => {
```
RFSH, line 26: Please add a jsdoc style comment to the function and explain what it does

```ts
/**
 * <description goes here>
 * @param {number} value <explain the value>
 */ 
const vwToPx = (value: number) => {
```

src/utils/ui-utils.ts
```diff
@@ -276,3 +277,20 @@ export function copyToClipboard(text: string) {
     });
   }
 }
+
+
+export function fireCustomEvent(eventName = 'myEvent', targetElement = 'body', detail = {}, bubbles = true, cancelable = true, composed = false) {
```
RFSH, line 282: Add jsdoc to this function as well.

src/assets/scss/_recordset.scss
```diff
@@ -1,155 +1,197 @@
 @import "helpers";
 @import "variables";
 
-.recordset-container {
-  .top-panel-container {
-    .top-flex-panel .top-left-panel div.panel-header {
-      bottom: 10px;
+.chaise-body {
```
RFSH, line 4: It's hard for me to figure out what exactly you've changed. If you're relying on component to pass the default width values, why are you setting them in css as well?

By the way, just like @jrchudy mentioned, the same set of rules can be found in [app.scss](https://github.com/informatics-isi-edu/chaise/blob/react/src/assets/scss/app.scss#L443). The idea is that the one in app.scss is supposed to define the "default "values. And then in here we customized these rules to make it look better in case of recordset.

If the changes in here was supposed to be for the "default" values, you should change the app.scss one instead.

src/components/resizable.tsx
anmol-singh-usc: @RFSH should I move this function into ui-utils ?

RFSH: Yeah that's a good idea. Make sure to name it something better. Maybe `convertViewWidthToPixel` or `convertVWToPixel`.

src/components/resizable.tsx
anmol-singh-usc: done

src/components/resizable.tsx
anmol-singh-usc: fixed. Bunching all these states together into a single state did the trick.

src/components/resizable.tsx
anmol-singh-usc: done

src/utils/ui-utils.ts
anmol-singh-usc: done

src/assets/scss/_recordset.scss
anmol-singh-usc: the difference is that I an setting the value of flex-basis as unset in the _recordset.scss file as compared to it being flex-basis: $left-panel-width-lg; in app.scss

anmol-singh-usc: this is necessary as giving the element a flex-basis value interferes with the resizable filter's working

RFSH: OK I just wanted to confirm that the rules in `app.scss` are still necessary.

src/components/resizable.tsx
```diff
@@ -0,0 +1,146 @@
+import '@isrd-isi-edu/chaise/src/assets/scss/_resizable.scss';
+import { useRef, useEffect, useState } from 'react';
+import { fireCustomEvent, convertVWToPixel } from '@isrd-isi-edu/chaise/src/utils/ui-utils';
+
+/* 
+Component Usage:
+    This component is used to implement a horizontal dragable and resizable component/div
+
+Props:
```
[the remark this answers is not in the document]
anmol-singh-usc: https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments - consider installing this vscode plugin

src/components/recordset.tsx
```diff
@@ -401,28 +432,24 @@ const RecordsetInner = ({
...
-          >
-            <div className='side-panel-container'>
-              <Faceting reference={reference} />
-            </div>
-          </div>
-        }
-        <div className='main-container dynamic-padding' ref={mainContainer}>
+      <SplitView
+        left={filterSection}
+        right={<div className='main-container dynamic-padding' ref={mainContainer}>
```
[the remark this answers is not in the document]
anmol-singh-usc: done

src/components/resizable.tsx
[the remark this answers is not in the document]
anmol-singh-usc: done

src/components/recordset.tsx
```diff
@@ -295,6 +309,23 @@ const RecordsetInner = ({
     )
   }
 
+  const filterSection = (leftRef: React.RefObject<HTMLDivElement>) => (
```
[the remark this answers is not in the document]
anmol-singh-usc: done

src/components/resizable.tsx
```diff
@@ -0,0 +1,146 @@
...
+    left - the left-pane component in the resizable layout
+    right - the right-pane component in the resizable layout
+    className - to apply custom styles on the component
+    minWidth - the minimum width limit on the left-pane component | defaults to 250px
+    maxWidth - the maximum width limit on the left-pane component | defaults to 450px
+    initialWidth - the initial width for the left-pane component | defaults to 270px
+    convertMaxWidth - use this flag in case you pass maxWidth in vw units instead of px units
```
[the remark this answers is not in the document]
anmol-singh-usc: done

RFSH (approved):
Thank you for the changes. Everything looks good to me. Please wait for @jrchudy's approval before merging the PR.

jrchudy (approved):

PR #2186, opened 2022-06-06
input-switch

anmol-singh-usc:
An input component used for inputing the start and end values for a range. 

This component will be used in the facet-range picker component.

Types supported: 
- int
- float
- numeric
- date
- timestamp

Validations:
- from < to
- int type can only input valid integers based on the following regex - /^\-?\d+$/
- float type can only input valid floats based on the following regex - /^\-?(\d+)?((\.)?\d+)?$/
- numeric type can only input valid floats based on the following regex - /^\-?(\d+)?((\.)?\d+)?$/
- date type should be a valid date (asserted by momentJS)
- datetime-local should be a valid datetime object (asserted by momentJS)

In case any validation fails the one the corresponding error messages are displayed:
- Please enter both From and To values
- From value cannot be greater than the To value
- Please enter an integer value
- Please enter a decimal value
- Please enter a date value in YYYY-MM-DD format
- Please enter a time value in 24-hr YYYY-MM-DDTHH:MM format

### DIFF
--- a/src/assets/scss/_range-input.scss
+++ b/src/assets/scss/_range-input.scss
@@ -21,7 +21,19 @@
         display: flex;
         flex-direction: row;
         align-items: center;
+        position: relative;
     }
+
+    .range-input-datetime {
+        flex-direction: column;
+
+        #range-to-val-date,
+        #range-from-val-date {
+            margin-bottom: 5px;
+        }
+    }
+
+
 
     .range-input-clear {
         position: absolute;
@@ -32,18 +44,15 @@
     .range-input {
         margin-right: 5px;
         flex-shrink: 0;
-        max-width: 200px;
+        max-width: 180px;
 
-        &.numeric-width {
+        &.numeric-width,
+        &.date-width {
             min-width: 140px;
         }
 
-        &.date-width {
-            min-width: 160px;
-        }
-
         &.time-width {
-            min-width: 200px;
+            min-width: 160px;
         }
     }
 
--- a/src/assets/scss/_range-input.scss
+++ b/src/assets/scss/_range-input.scss
@@ -74,7 +74,8 @@
         margin-top: 5px;
     }
 
-    .range-input-error {
-        color: #a94442;
-    }
+}
+
+.range-input-error {
+    color: #a94442;
 }
--- a/src/assets/scss/_range-input.scss
+++ b/src/assets/scss/_range-input.scss
@@ -34,7 +34,7 @@
             }
 
             .range-input-clear {
-                top: 2px;
+                top: 3px;
             }
         }
 
--- a/src/assets/scss/_range-input.scss
+++ b/src/assets/scss/_range-input.scss
@@ -1,3 +1,5 @@
+@import "variables";
+
 .range-input-container {
     display: flex;
     flex-direction: row;
@@ -25,49 +27,6 @@
         label {
             font-weight: 700;
         }
-
-        .chaise-input-control {
-
-            &.has-feedback {
-                height: unset;
-                min-height: auto;
-            }
-
-            .range-input-clear {
-                top: 3px;
-            }
-        }
-
-        .range-input-field {
-            display: flex;
-            flex-direction: row;
-            align-items: center;
-            position: relative;
-        }
-
-        .range-input-datetime {
-            flex-direction: column;
-
-            .chaise-input-control:first-child {
-                margin-bottom: 5px;
-            }
-        }
-
-        #range-from-val-date,
-        #range-to-val-date,
-        #range-from-val-time,
-        #range-from-val-time,
-        #range-from-val,
-        #range-to-val {
-            padding-right: 0;
-            height: auto;
-        }
-
-        .range-input-clear {
-            position: absolute;
-            right: 6px;
-            cursor: pointer;
-        }
     }
 
     .range-input-submit-btn {
@@ -77,5 +36,5 @@
 }
 
 .range-input-error {
-    color: #a94442;
+    color: $danger-color;
 }
--- a/src/assets/scss/_range-input.scss
+++ b/src/assets/scss/_range-input.scss
@@ -12,55 +12,69 @@
         flex-direction: column;
         margin-top: 5px;
 
-        label {
-            font-weight: 700;
-        }
-    }
-
-    .range-input-field {
-        display: flex;
-        flex-direction: row;
-        align-items: center;
-        position: relative;
-    }
-
-    .range-input-datetime {
-        flex-direction: column;
-
-        #range-to-val-date,
-        #range-from-val-date {
-            margin-bottom: 5px;
-        }
-    }
-
-
-
-    .range-input-clear {
-        position: absolute;
-        right: 6px;
-        cursor: pointer;
-    }
-
-    .range-input {
         margin-right: 5px;
         flex-shrink: 0;
-        max-width: 180px;
+        max-width: 160px;
 
         &.numeric-width,
-        &.date-width {
+        &.date-width,
+        &.time-width {
             min-width: 140px;
         }
 
-        &.time-width {
-            min-width: 160px;
+        label {
+            font-weight: 700;
+        }
+
+        .chaise-input-control {
+
+            &.has-feedback {
+                height: unset;
+                min-height: auto;
+            }
+
+            .range-input-clear {
+                top: 2px;
+            }
+        }
+
+        .range-input-field {
+            display: flex;
+            flex-direction: row;
+            align-items: center;
+            position: relative;
+        }
+
+        .range-input-datetime {
+            flex-direction: column;
+
+            .chaise-input-control:first-child {
+                margin-bottom: 5px;
+            }
+        }
+
+        #range-from-val-date,
+        #range-to-val-date,
+        #range-from-val-time,
+        #range-from-val-time,
+        #range-from-val,
+        #range-to-val {
+            padding-right: 0;
+            height: auto;
+        }
+
+        .range-input-clear {
+            position: absolute;
+            right: 6px;
+            cursor: pointer;
         }
     }
 
     .range-input-submit-btn {
         margin-top: 5px;
     }
-}
 
-.range-input-error {
-    color: #a94442;
+    .range-input-error {
+        color: #a94442;
+    }
 }
--- a/src/assets/scss/_range-input.scss
+++ b/src/assets/scss/_range-input.scss
@@ -1,0 +1,57 @@
+.range-input-container {
+    display: flex;
+    flex-direction: row;
+    align-items: end;
+    width: 100%;
+    flex-wrap: wrap;
+    margin-top: -5px;
+
+    .range-input {
+        width: 40%;
+        display: flex;
+        flex-direction: column;
+        margin-top: 5px;
+
+        label {
+            font-weight: 700;
+        }
+    }
+
+    .range-input-field {
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+    }
+
+    .range-input-clear {
+        position: absolute;
+        right: 6px;
+        cursor: pointer;
+    }
+
+    .range-input {
+        margin-right: 5px;
+        flex-shrink: 0;
+        max-width: 200px;
+
+        &.numeric-width {
+            min-width: 140px;
+        }
+
+        &.date-width {
+            min-width: 160px;
+        }
+
+        &.time-width {
+            min-width: 200px;
+        }
+    }
+
+    .range-input-submit-btn {
+        margin-top: 5px;
+    }
+}
+
+.range-input-error {
+    color: #a94442;
+}
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -183,6 +183,11 @@ html {
     display: none;
   }
 
+  .remove-input-btn {
+    cursor: pointer;
+    pointer-events: all;
+  }
+
   .disabled-form .btn.btn-inverted[disabled]:hover {
     background-color: white !important;
     border-color: $border-color  !important;
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -184,6 +184,7 @@ html {
   }
 
   .remove-input-btn {
+    /*the button used to clear an input field - clear-input-btn.tsx*/
     cursor: pointer;
     pointer-events: all;
   }
--- a/src/components/input-switch.tsx
+++ b/src/components/input-switch.tsx
@@ -22,23 +22,21 @@ type DateTimePickerProps = {
      */
     timeRef?: React.RefObject<HTMLInputElement>,
     /**
+     * flag for showing clear button on date field
+     */
+    showClearDateBtn: boolean,
+    /**
+     * flag for showing clear button on time field
+     */
+    showClearnTimeBtn: boolean,
+    /**
      * the handler function called on input change
      */
     handleChange: (() => void)
 };
 
-const DateTimePicker = ({ classes, id, value, handleChange, dateRef, timeRef }: DateTimePickerProps): JSX.Element => {
-
-    // const changeHandler = () => {
-    //     const dateVal = dateRef?.current?.value;
-    //     let timeVal = timeRef?.current?.value;
-
-    //     console.log({ dateVal, timeVal });
-
-    //     if (dateVal && !timeVal) timeVal = '00:00';
-
-    //     handleChange(dateVal || timeVal ? `${dateVal || ''}T${timeVal}` : '');
-    // }
+const DateTimePicker = ({ classes, id, value, handleChange, dateRef,
+    timeRef, showClearDateBtn, showClearnTimeBtn }: DateTimePickerProps): JSX.Element => {
 
     const handleTimeClear = () => {
         if (timeRef?.current) timeRef.current.value = '';
@@ -58,7 +56,7 @@ const DateTimePicker = ({ classes, id, value, handleChange, dateRef, timeRef }: 
                 <ClearInputBtn
                     btnClassName='range-input-clear'
                     clickCallback={handleDateClear}
-                    show={Boolean(dateRef?.current?.value)}
+                    show={showClearDateBtn}
                 />
             </div>
             <div className='chaise-input-control has-feedback'>
@@ -67,7 +65,7 @@ const DateTimePicker = ({ classes, id, value, handleChange, dateRef, timeRef }: 
                 <ClearInputBtn
                     btnClassName='range-input-clear'
                     clickCallback={handleTimeClear}
-                    show={Boolean(timeRef?.current?.value)}
+                    show={showClearnTimeBtn}
                 />
             </div>
         </div>
@@ -105,17 +103,25 @@ type InputSwitchProps = {
     id: string,
     /** 
      * the default date value being used in case of date and timestamp types
-    */
+     */
     value: string,
-    /** 
+    /**
+     * flag for showing clear button on input field
+     */
+    showClearBtn: boolean,
+    /**
+     * flag for showing clear button on time field in case type is timestamp
+     */
+    showClearnTimeBtn: boolean,
+    /**
      * the handler function called on input change
-    */
-    handleChange: (() => void)
+     */
+    handleChange: (() => void),
+
 };
 
-const InputSwitch = ({ placeholder = 'Enter', classes = '', id, reference, timeRef, type, value, handleChange }: InputSwitchProps): JSX.Element => {
-
-    // const changeHandler = () => handleChange(reference?.current?.value || '');
+const InputSwitch = ({ placeholder = 'Enter', classes = '', id, reference, timeRef,
+    type, value, handleChange, showClearBtn, showClearnTimeBtn }: InputSwitchProps): JSX.Element => {
 
     const clearInput = () => {
         if (reference?.current) reference.current.value = '';
@@ -124,7 +130,8 @@ const InputSwitch = ({ placeholder = 'Enter', classes = '', id, reference, timeR
 
     return (
         type === 'timestamp' ? (
-            <DateTimePicker classes={classes} value={value} id={id} dateRef={reference} timeRef={timeRef} handleChange={handleChange} />
+            <DateTimePicker classes={classes} value={value} id={id} dateRef={reference}
+                timeRef={timeRef} showClearDateBtn={showClearBtn} showClearnTimeBtn={showClearnTimeBtn} handleChange={handleChange} />
         ) : (
             <div className='chaise-input-control has-feedback'>
                 {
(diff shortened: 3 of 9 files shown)


jrchudy (changes requested):
We don't want to make either minor max required since we allow for only one value to be set and we do a less than or greater than query instead.

Can you removed the required attribute and validation for having both inputs defined and can you go through and get the code working based on the types. See my comment in facet-range-picker.tsx

Other updates from our discussion in chaise meeting:
 - don't stretch the time/date as the panel stretches
 - use the `clear-input` directive for the x button
 - change the order of the calendar and clock with the x button

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+ */
+const errorMsgMap: {
+    [key: string]: string;
+} = {
+    'empty': 'Please enter either From or To value',
+    'range': 'From value cannot be greater than the To value',
+    'int': 'Please enter an integer value',
+    'float': 'Please enter a decimal value',
+    'date': 'Please enter a date value in YYYY-MM-DD format',
+    'timestamp': 'Please enter a time value in YYYY-MM-DD HH:MM format'
```
jrchudy, line 26: Is this for both the date and time error of timestamp inputs or just the time input? If it's just the time input, let's rename this to `time`.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+};
+
+type DateTimePickerProps = {
```
jrchudy, line 29: Can you add jsdoc comments for each prop being added? Explain what the property is for and any restrictions it might have.

similar to how Ravish had done it here:
https://github.com/informatics-isi-edu/chaise/blob/react/src/components/facet-header.tsx#L12-L37

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+    if (inputType.indexOf('float') > -1) return 'float';
+
+    if (inputType.indexOf('date') > -1) return 'date';
+
+    if (inputType.indexOf('timestamp') > -1) return 'timestamp';
+
+    return 'Invalid_Type';
+}
+
+const RangeInputHOC = ({ inputType, classes }: RangeInputHOCProps) => {
```
jrchudy, line 156: Just from googling, I'm guessing the abbreviation here means "Higher Order Component"?

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+    const toTimeRef = useRef<HTMLInputElement>(null);
+
+    const type = getType(inputType);
+
+    const [error, setError] = useState<string | null>(null);
+
+    const [disableSubmit, setDisableSubmit] = useState<boolean>(type === 'int' || type === 'float');
+
+    const disableSubmitBtn = () => {
```
jrchudy, line 173: Is it possible to combine disable/enable into one function call?

something like:
```
toggleSubmitBtn(condition) => {
  setDisableSubmit(condition)
}
```

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**checks if both inputs are empty and disables the submit button */
+    const handleChange = (value: string) => {
+        const formatedValues = type === 'timestamp' ? formatTimeValues()
+            : { fromVal: fromRef?.current?.value || '', toVal: toRef?.current?.value || '' };
+
+        console.log(formatedValues);
+
+        if (!formatedValues.fromVal && !formatedValues.toVal) disableSubmitBtn();
```
jrchudy, line 205: After changing the submitBtn function above, this can change to:
```
toggleSubmitBtn(!formatedValues.fromVal && !formatedValues.toVal)
```

So in this case, if the condition is true, we want to `setDisableSubmit` to true.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+            toVal: toDateVal || toTimeVal ? `${toDateVal}T${toTimeVal}` : '',
+        };
+    }
+
+    /**checks if both inputs are empty and disables the submit button */
+    const handleChange = (value: string) => {
+        const formatedValues = type === 'timestamp' ? formatTimeValues()
+            : { fromVal: fromRef?.current?.value || '', toVal: toRef?.current?.value || '' };
+
+        console.log(formatedValues);
```
jrchudy, line 203: Please remove this when debugging is done.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+        if (!formatedValues.fromVal && !formatedValues.toVal) disableSubmitBtn();
+        else enableSubmitBtn();
+
+        const validatedResult = validateValue(value);
+
+        if (!validatedResult) setError(errorMsgMap[type]);
+
+        if (validatedResult && !error) setError(null);
```
jrchudy, line 212: One of my comments in slack was if the value was invalid before (error is defined) and it is now valid, we want to hide the error. This check should change to make sure no value is invalid for either `to` or `from` before hiding the error.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+import { useState, useRef } from 'react';
+import { windowRef } from '@isrd-isi-edu/chaise/src/utils/window-ref';
+import { ClearInputBtn } from '@isrd-isi-edu/chaise/src/components/clear-input-btn';
+
+
+const INTEGER_REGEXP = /^\-?\d+$/;
+
+const FLOAT_REGEXP = /^\-?(\d+)?((\.)?\d+)?$/;
+
+const TIME_FORMAT = 'YYYY-MM-DDTHH:mm';
```
jrchudy, line 11: We should call this `TIMESTAMP_FPRMAT`

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+        return rangeCheck(fromVal, toVal) ? 'valid' : 'range';
+    }
+
+    const handleSubmit = () => {
+        const formatedValues = type === 'timestamp' ? formatTimeValues()
+            : { fromVal: fromRef?.current?.value || '', toVal: toRef?.current?.value || '' };
+
+        const validatedResult = validateValues(formatedValues.fromVal, formatedValues.toVal);
+
+        console.log('validation result: ', validatedResult);
```
jrchudy, line 261: Please remove this when debugging is done.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+        if (validatedResult === 'valid') {
+            /* clear out any previous errors when a new submission is validated */
+            setError(null);
+
+            /* eslint-disable  @typescript-eslint/no-non-null-assertion */
+            console.log('values sent to server', formatedValues);
```
jrchudy, line 268: Please remove this when debugging is done.

jrchudy (changes requested):
The changes and functionality look really good. I left a few comments in the PR about cleaning up some of the console logs, renaming some variables, and combining the enable/disable functions.

Other than that, I left 3 comments in our conversation in slack that I will copy here:
 - the integer input shows an error after typing 123.1, after removing the “.1” the error still shows. We should update the displayed validation error (hide it) if the value is now “valid” instead of waiting until they click the submit button. 
      - NOTE: make sure if the error shows and applies to both inputs, if the error is fixed for one of the inputs, make sure the error still shows if it still applies because of the other input
 - to/from error should be hidden as soon as one of the inputs is changed again
 - calendar/clock are slightly lower than the “x”

src/components/range-inputs.tsx
anmol-singh-usc: yes RangeInputHOC is the component which renders the 2 range inputs

src/components/range-inputs.tsx
anmol-singh-usc: no this is the type in case we want a timestamp so we render both date n time fields. In case we only want a date field we send the type as date.

src/components/range-inputs.tsx
anmol-singh-usc: done

src/components/range-inputs.tsx
anmol-singh-usc: once you have tested the functionality I'll remove all the logs

src/components/range-inputs.tsx
anmol-singh-usc: `      
const validatedResult = validateValue(formatedValues.fromVal) && validateValue(formatedValues.toVal);

if (!validatedResult) setError(errorMsgMap[type])
else setError(null);
`

changed the validation logic check to this so your issue should be resolved now

src/components/range-inputs.tsx
anmol-singh-usc: done

src/components/range-inputs.tsx
anmol-singh-usc: once you have tested the functionality I'll remove all the logs

src/components/range-inputs.tsx
anmol-singh-usc: once you have tested the functionality I'll remove all the logs

src/components/range-inputs.tsx
```diff
@@ -196,20 +233,18 @@ const RangeInputHOC = ({ inputType, classes }: RangeInputHOCProps) => {
...
-    const handleChange = (value: string) => {
+    const handleChange = () => {
         const formatedValues = type === 'timestamp' ? formatTimeValues()
             : { fromVal: fromRef?.current?.value || '', toVal: toRef?.current?.value || '' };
 
         console.log(formatedValues);
 
-        if (!formatedValues.fromVal && !formatedValues.toVal) disableSubmitBtn();
-        else enableSubmitBtn();
+        toggleSubmitBtn(!formatedValues.fromVal && !formatedValues.toVal);
```
jrchudy, line 244: When one of the values is invalid, we want the submit button to be disabled. I think the condition needs to be updated to check `!validatedResult` also which mean this will have to be called after calling `validateValue` for both values.

jrchudy (changes requested):
Some more comments I made include:
 - changing the validation error messages since the format in our messages might not match the browser format
 - When no value is shown in the input, the clear button should no be shown
   - in the case of date and time, the order of the buttons should change so that when the "x" is hidden, the calendar/clock don't have a lot of whitespace to the right of them
 - submit button should be disabled if one of the values is invalid
 
 
I realized the "input switch" is the `RangeInput` component. We want this component to be for reusable inputs in `recordedit` app as well so maybe the name should be made more general instead of specific for the inputs in the range picker, we called it `InputSwitch` in angularJS. This means the component also needs to be exported, so it should probably have it's own `.tsx` file as well.

src/components/range-inputs.tsx
jrchudy: After renaming `RangeInput` to `InputSwitch`, this can be called `RangeInputs`

src/components/range-inputs.tsx
```diff
@@ -8,7 +8,7 @@ const INTEGER_REGEXP = /^\-?\d+$/;
 
 const FLOAT_REGEXP = /^\-?(\d+)?((\.)?\d+)?$/;
 
-const TIME_FORMAT = 'YYYY-MM-DDTHH:mm';
+const TIMESTAMP_FPRMAT = 'YYYY-MM-DDTHH:mm';
```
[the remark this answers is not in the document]
anmol-singh-usc: https://github.com/informatics-isi-edu/chaise/pull/2186/commits/535b3cc845ae1df38f9c42742e1fb0425341705e

src/components/range-inputs.tsx
anmol-singh-usc:
```
const validatedResult = validateValue(formatedValues.fromVal) && validateValue(formatedValues.toVal);

const areBothFieldsEmpty = !formatedValues.fromVal && !formatedValues.toVal;

toggleSubmitBtn(areBothFieldsEmpty || !validatedResult);
```

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,349 @@
...
+ * an object for mapping the error types to error messages
+ */
+const errorMsgMap: {
+    [key: string]: string;
+} = {
+    'empty': 'Please enter either From or To value',
+    'range': 'From value cannot be greater than the To value',
+    'int': 'Please enter an integer value',
+    'float': 'Please enter a decimal value',
+    'date': 'Please enter a date value in YYYY-MM-DD format',
```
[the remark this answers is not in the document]
anmol-singh-usc: have moved the input component logic into a separate file called input-switch. 

clear btn now enabled only when field has a valid value

cannot not change the order of the calendar icon/clock icon with the cross icon. Also cannot change the width of the field so some with space will be there on the right side (let's discuss on this)

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+    return (
+        <div className='range-input-datetime'>
+            <div className='chaise-input-control has-feedback'>
+                <input id={`${id}-date`} className={classes} type='date' ref={dateRef}
+                    placeholder='YYYY-MM-DD' min='1970-01-01' max='2999-12-31' step='1' defaultValue={value} onChange={handleChange} />
+                <ClearInputBtn
+                    btnClassName='range-input-clear'
+                    clickCallback={handleDateClear}
+                    show
```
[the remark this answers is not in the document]
anmol-singh-usc: done

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,349 @@
...
+                {
+                    type === 'int' || type === 'float' ? <input id={id} type='number' placeholder={placeholder} className={classes}
+                        ref={reference} onChange={handleChange} />
+                        : <input id={id} type='date' className={classes} ref={reference} step='1'
+                            defaultValue={value} pattern='\d{4}-\d{2}-\d{2}' min='1970-01-01' max='2999-12-31' onChange={handleChange} />
+                }
+                <ClearInputBtn
+                    btnClassName='range-input-clear'
+                    clickCallback={clearInput}
+                    show
```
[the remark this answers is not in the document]
anmol-singh-usc: done

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+const DATE_FORMAT = 'YYYY-MM-DD';
+
+/**
+ * an object for mapping the error types to error messages
+ */
+const errorMsgMap: {
+    [key: string]: string;
+} = {
+    'empty': 'Please enter either From or To value',
```
jrchudy, line 20: I don't think this message is necessary since submission is disabled when no value is present.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,211 @@
...
+    'range': 'From value cannot be greater than the To value',
+    'int': 'Please enter an integer value',
+    'float': 'Please enter a decimal value',
+    'date': 'Please enter a date value',
```
jrchudy, line 24: Should be:
`Please enter a valid date value`

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,211 @@
...
+    'timestamp': 'Please enter a valid date and time value'
+};
+
+type RangeInputProps = {
```
jrchudy, line 28: `RangeInputsProps`

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,211 @@
...
+    if (inputType.indexOf('float') > -1) return 'float';
+
+    if (inputType.indexOf('date') > -1) return 'date';
+
+    if (inputType.indexOf('timestamp') > -1) return 'timestamp';
+
+    return 'Invalid_Type';
+}
+
+const RangeInput = ({ inputType, classes }: RangeInputProps) => {
```
jrchudy, line 56: This is a container for both to and from inputs, the name should be plural.
`RangeInputs`

jrchudy (commented):
A couple name changes and the delayed clear button from my comment below.

From my comment in slack:
> if I open my integer facet and put a value into the “from” input, I see the “x” show right away. While that value is there, as I type a value into the “to” input, I don’t see the “x” show up until I click on the input (or somewhere else on the page). When the “x” does show and I click it, the value clears but the “x” remains until I click somewhere else.

src/components/range-inputs.tsx
anmol-singh-usc: yes I forgot to remove this from the error map even though we don't return this error from the validation function anymore

src/components/range-inputs.tsx
anmol-singh-usc: done

jrchudy (approved):
These changes look good to me. Assigning to @RFSH for review.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,234 @@
...
+        return date.isValid();
+    }
+
+    /**
+     * performs basic range validation : from_value > to_value
+    */
+    const rangeCheck = (fromVal: string, toVal: string): boolean => {
+        if (type === 'int') return parseInt(fromVal) < parseInt(toVal);
+
+        if (type === 'float') return parseInt(fromVal) < parseInt(toVal);
```
RFSH, line 149: This should be `parseFloat`.

src/assets/scss/_range-input.scss
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    .range-input-submit-btn {
+        margin-top: 5px;
+    }
+
+}
+
+.range-input-error {
+    color: #a94442;
```
RFSH, line 80: We already have a `$danger-color` variable defined in `varables.scss`. Import it in this file and then use the same color.

src/assets/scss/app.scss
```diff
@@ -183,6 +183,11 @@ html {
     display: none;
   }
 
+  .remove-input-btn {
```
RFSH, line 186: Comment what this is referring to.

src/assets/scss/_range-input.scss
```diff
@@ -0,0 +1,81 @@
...
+
+        .range-input-datetime {
+            flex-direction: column;
+
+            .chaise-input-control:first-child {
+                margin-bottom: 5px;
+            }
+        }
+
+        #range-from-val-date,
```
RFSH, line 56: Why are you writing rules based on `id`? You should do this with `class` instead.

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+    const classTypeName = (type === 'int' || type === 'float') ? 'numeric-width' : type === 'date' ? 'date-width' : 'time-width';
+
+    return (
+        <div className={classes}>
+            <div className='range-input-container'>
+                <div className={`range-input ${classTypeName}`}>
+                    <label htmlFor={`range-from-val-${type}`}>From:</label>
+                    <InputSwitch
+                        id={`range-from-val-${type}`}
```
RFSH, line 200: What's the purpose of this `id`? Did you add it only for `htmlFor` of the label?

One mistake that we made in AngularJS version was abusing `id` and using them when we knew it won't be unique.  In this case, the `id` value that you gave is not unique since there might be multiple range pickers of the same type on the page.

RFSH (changes requested):
Most probably Josh has already mentioned this to you, but we're using 2 space indents. So please fix the indentation for the changes in this PR.

Apart from `int` and `float`, we also have a `numeric` type that you're not handling. You should treat this type the same as `float`. Other than that, I asked for some small changes. Please let me know you applied these changes so we can go ahead and merge your changes.

-------------

I really don't like how the `InputSwitch` is delegating everything to the parent. This component is supposed to be self-contained and the parent should just get the value from it (InputSwitch itself should handle the clear-btn state for example). I also don't know why you're not using [controlled components](https://reactjs.org/docs/forms.html#controlled-components) like how we did it in "Go to RID" and "Search" inputs.

That being said, we don't want to delay merging these changes any further and things will get more interesting when you get to use this component in recordedit form. And based on what you learned from there, you most probably would need to come back and change the implementation here, so I'm not opposed to pushing it this way for now.

src/components/range-inputs.tsx
anmol-singh-usc: removed ids

src/assets/scss/_range-input.scss
anmol-singh-usc: fixed

src/assets/scss/app.scss
anmol-singh-usc: done

src/assets/scss/_range-input.scss
anmol-singh-usc: done

src/components/range-inputs.tsx
anmol-singh-usc: done

src/components/input-switch.tsx
```diff
@@ -0,0 +1,154 @@
...
+        ) : (
+            <div className='chaise-input-control has-feedback'>
+                {
+                    type === 'int' || type === 'float' ? <input id={id} type='number' placeholder={placeholder} className={classes}
+                        ref={reference} onChange={handleChange} />
+                        : <input id={id} type='date' className={classes} ref={reference} step='1'
+                            defaultValue={value} pattern='\d{4}-\d{2}-\d{2}' min='1970-01-01' max='2999-12-31' onChange={handleChange} />
+                }
+                <ClearInputBtn
+                    btnClassName='range-input-clear'
```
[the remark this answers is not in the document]
anmol-singh-usc: renamed this to input-switch-clear

src/components/input-switch.tsx
```diff
@@ -0,0 +1,154 @@
...
+        handleChange();
+    }
+
+    const handleDateClear = () => {
+        if (dateRef?.current) dateRef.current.value = '';
+        handleChange();
+    }
+
+    return (
+        <div className='range-input-datetime'>
```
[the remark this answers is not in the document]
anmol-singh-usc: yes I agree, fixed these and made them specific to input-switch

src/components/input-switch.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+const InputSwitch = ({ placeholder = 'Enter', classes = '', id, reference, timeRef,
+    type, value, handleChange, showClearBtn, showClearnTimeBtn }: InputSwitchProps): JSX.Element => {
+
+    const clearInput = () => {
+        if (reference?.current) reference.current.value = '';
+        handleChange();
+    }
+
+    return (
```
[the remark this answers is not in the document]
anmol-singh-usc: done

src/components/range-inputs.tsx
```diff
@@ -0,0 +1,234 @@
...
+            from: Boolean(fromRef?.current?.value),
+            to: Boolean(toRef?.current?.value),
+            fromTime: Boolean(fromTimeRef?.current?.value),
+            toTime: Boolean(toTimeRef?.current?.value),
+        });
+
+        const formatedValues = type === 'timestamp' ? formatTimeValues()
+            : { fromVal: fromRef?.current?.value || '', toVal: toRef?.current?.value || '' };
+
+        const validatedResult = validateValue(formatedValues.fromVal) && validateValue(formatedValues.toVal);
```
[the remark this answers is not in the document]
anmol-singh-usc: as per our discussion for now I'm leaving it as is. Will work on moving the validation and handle change logic into the input-switch component after this PR is merged

RFSH (approved):

PR #2190, opened 2022-06-15
Faceting > HTML structure > panel header

ravish-kg:
Changes pushed to panel-header branch are:
 Made sure the header looks the same as master (Old Angular).
 Made sure the header height and font is the same
 Made sure the "chevron" looks the same
 Conditionally added the class when the header has a comment (just like the columns in recordset-table.tsx)
 Made sure ellipsis should work properly:
     -- If the header has comment, always show the comment.
     -- Otherwise, if the ellipsis is displayed, the tooltip should show the header name.

### DIFF
--- a/src/assets/scss/_faceting.scss
+++ b/src/assets/scss/_faceting.scss
@@ -51,10 +51,12 @@
           font-family: "Font Awesome 6 Free";
           font-weight: 900;
           transform: unset;
+          font-size: 18px;
+          height: unset;
         }
 
         &:not(.collapsed)::after {
-          content: '\f107'; // fa-solid fa-chevron-down
+          content: '\f078'; // fa-solid fa-chevron-down
         }
 
         &:focus {
--- a/src/assets/scss/_faceting.scss
+++ b/src/assets/scss/_faceting.scss
@@ -51,7 +51,7 @@
           font-family: "Font Awesome 6 Free";
           font-weight: 900;
           transform: unset;
-          font-size: 18px;
+          font-size: 1.25rem;
           height: unset;
         }
 
--- a/src/assets/scss/_faceting.scss
+++ b/src/assets/scss/_faceting.scss
@@ -36,9 +36,22 @@
         padding: 10px 5px;
         text-transform: none;
         position: relative;
+        // Overriding accordian-button css to align right angle bracket icon to left
+        flex-direction: row-reverse;
+        justify-content: flex-end;
 
         background-color: $table-header-background-color;
         color: $black-color;
+
+        // Overriding accordian-button css to align right angle bracket icon to left
+        &::after {
+          margin: 0px 8px;
+          transform: rotate(-90deg);
+        }
+
+        &:not(.collapsed)::after {
+          transform: rotate(0deg);
+        }
 
         &:focus {
           box-shadow: none;
@@ -65,6 +78,16 @@
       .accordion-body {
         padding: 10px;
       }
+    }
+
+    .facet-header-text {
+      font-size: 1.25rem;
+    }
+
+    .align-center-icon {
+      line-height: 20px;
+      vertical-align: middle;
+      margin: 4px;
     }
   }
 }
--- a/src/assets/scss/_faceting.scss
+++ b/src/assets/scss/_faceting.scss
@@ -46,11 +46,15 @@
         // Overriding accordian-button css to align right angle bracket icon to left
         &::after {
           margin: 0px 8px;
-          transform: rotate(-90deg);
+          background-image: none;
+          content: '\f054'; // fa-solid fa-chevron-right
+          font-family: "Font Awesome 6 Free";
+          font-weight: 900;
+          transform: unset;
         }
 
         &:not(.collapsed)::after {
-          transform: rotate(0deg);
+          content: '\f107'; // fa-solid fa-chevron-down
         }
 
         &:focus {
--- a/src/components/FacetHeader.tsx
+++ b/src/components/FacetHeader.tsx
@@ -1,0 +1,89 @@
+import '@isrd-isi-edu/chaise/src/assets/scss/_faceting.scss';
+
+import { useRef, useState } from 'react';
+
+// Components
+import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
+import Tooltip from 'react-bootstrap/Tooltip';
+import DisplayValue from '@isrd-isi-edu/chaise/src/components/display-value';
+
+type FacetHeaderProps = {
+  key: string;
+  headerText: any;
+  showTooltipIcon: any;
+  tooltipContent: any;
+};
+
+/**
+ * @param key key prop to uniquely identify panel header item
+ * @param headerText headerText prop that holds content to show as header
+ * @param showTooltipIcon prop to enable tooltip for header text
+ * @param tooltipContent text to be shown on hover of header text
+ * @returns FacetHeader Component
+ */
+const FacetHeader = ({
+  key,
+  headerText,
+  showTooltipIcon,
+  tooltipContent,
+}: FacetHeaderProps) => {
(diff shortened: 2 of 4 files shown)


src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+type FacetHeaderProps = {
+  key: string;
+  headerText: any;
+  showTooltipIcon: any;
+  tooltipContent: any;
+};
+
+/**
+ * @param key key prop to uniquely identify panel header item
```
RFSH, line 18: Move these comments to the custom type. Like we mentioned each property in the type should have its own comment in place.

src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import { useRef, useState } from 'react';
+
+// Components
+import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
+import Tooltip from 'react-bootstrap/Tooltip';
+import DisplayValue from '@isrd-isi-edu/chaise/src/components/display-value';
+
+type FacetHeaderProps = {
+  key: string;
```
RFSH, line 11: I think you were asking about this one and I didn't fully understand what you meant before. I was looking at where this `id` is used. We're using this in testing to get the button that we want to click on. Given that I've moved the click target to be on the `Accordion.Header`,  we should also make sure the `id` is attached to the same element.

So you don't need to pass this `key` here and the id should be defined on the `Accordion.Header`.

src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+  headerText: any;
+  showTooltipIcon: any;
```
RFSH, line 13: Use proper `boolean` type

src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+  tooltipContent: any;
```
RFSH, line 14: Use proper `string` type, and let's rename it to `comment` for consistency. It should also be optional.

src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { useRef, useState } from 'react';
+
+// Components
+import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
+import Tooltip from 'react-bootstrap/Tooltip';
+import DisplayValue from '@isrd-isi-edu/chaise/src/components/display-value';
+
+type FacetHeaderProps = {
+  key: string;
+  headerText: any;
```
RFSH, line 12: Use proper `Displayname` type, and let's rename it to `displayname` for consistency.

src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <OverlayTrigger
+      trigger='hover'
+      placement='right'
+      overlay={
+        <Tooltip style={{ maxWidth: '50%', whiteSpace: 'nowrap' }}>
+          {contentRef &&
+          contentRef.current &&
+          isTextOverflow(contentRef.current) ? (
+            <>
+              <DisplayValue value={headerText} />: {tooltipContent}
```
RFSH, line 55: The `:` should be included only if `tooltipContent` is truthy.

src/assets/scss/_faceting.scss
```diff
@@ -36,10 +36,23 @@
...
+        // Overriding accordian-button css to align right angle bracket icon to left
+        flex-direction: row-reverse;
+        justify-content: flex-end;
 
         background-color: $table-header-background-color;
         color: $black-color;
 
+        // Overriding accordian-button css to align right angle bracket icon to left
+        &::after {
+          margin: 0px 8px;
```
RFSH, line 48: Given that the size and color of the icon is not the same as master, instead of just transforming to fix the direction, you should just show the icon that we want instead of the svg that bootstrap uses.

Something like this
```scss
&::after {
    background-image: none;
    content: '\f078'; // fa-solid fa-chevron-down
    font-family: "Font Awesome 6 Free";
    font-weight: 900;
    // TODO change the font-size and position to make sure it looks the same as master
}

&:not(.collapsed)::after {
  content: '\f054'; // fa-solid fa-chevron-right
  // TODO change the font-size and position to make sure it looks the same as master
}
```

src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return false;
+  };
+
+  return (
+    <OverlayTrigger
+      trigger='hover'
+      placement='right'
+      overlay={
+        <Tooltip style={{ maxWidth: '50%', whiteSpace: 'nowrap' }}>
+          {contentRef &&
```
RFSH, line 51: Given that it has too many different conditions now, move this to a function and use proper if statements. This would help with readability of your code.

RFSH (changes requested):

RFSH:
I just saw the filename. Please make sure you're following the guidelines. Filenames must be all lowercase and `-` should be used to separate words. So the new component filename should be `facet-header.tsx`

src/components/FacetHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+}: FacetHeaderProps) => {
+  /**
+   * @contentRef variable to store ref of facet header text
+   * @show state variable to control whether to show tooltip or not
+   */
+  const contentRef = useRef(null);
+  const [show, setShow] = useState(false);
+
+  // Function to check the text overflow.
+  const isTextOverflow = (element) => {
```
RFSH, line 38: Add proper type to the parameter.

src/components/FacetHeader.tsx
ravish-kg: Resolved

src/components/FacetHeader.tsx
ravish-kg: Resolved

src/components/FacetHeader.tsx
ravish-kg: Resolved

src/components/FacetHeader.tsx
ravish-kg: Made a separate function and added the missing condition

src/components/FacetHeader.tsx
ravish-kg: resolved

src/components/FacetHeader.tsx
ravish-kg: Made a separate function.

src/components/facet-header.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useRef, useState } from 'react';
+
+// Components
+import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
+import Tooltip from 'react-bootstrap/Tooltip';
+import DisplayValue from '@isrd-isi-edu/chaise/src/components/display-value';
+import { Displayname } from '@isrd-isi-edu/chaise/src/models/displayname';
+
+type FacetHeaderProps = {
+  displayname: Displayname; // content to be displayed as panel header
```
RFSH, line 12: This is not what I meant. Do it like you did in the other PR in the jsdoc style. The goal is to make sure vscode (or any other tools) can find the comment and show it.

```
  /**
   * comment goes here
   */
  displayname: Displayname;
```

src/components/facet-header.tsx
```diff
@@ -0,0 +1,88 @@
...
+  showTooltipIcon?: boolean; // prop to enable tooltip for displayname
+  comment?: string; // Optional text to be shown on hover of displayname
+};
+
+/**
+ * @returns FacetHeader Component
```
RFSH, line 18: Just add a line explaining what this is.. I know the name is self explanatory, but for completeness let's add it.

RFSH (approved):

PR #2191, opened 2022-06-15
Broken fk picker test in edit mode

jrchudy:
Move modal title setter and add a 100 ms sleep before trying to open the modal again


jrchudy:
Merge into master since it doesn't make the tests worse but we can't verify with any certainty that this fixes the problem since there are _other issues_ with the tests. Notably, the connection to saucelabs seems to be dropped almost every time we have run (and rerun) the test cases for this PR.

PR #2194, opened 2022-06-21
Export > dropdown and functionality

ravish-kg:
This PR has changes related to export functionality:

1. The button should be disabled while data is loading
2. Clicking the button should display the list of templates in a dropdown.
3. There should be a tooltip for the button. Just like the navbar, the tooltip should be hidden when the dropdown is open.
4. Depending on the type of the template, on click of the option,
5. If it's the default CSV option, change the location of browser
6. Otherwise,
    - Open a modal that shows we're waiting for data
    - When the export is done, close the modal and prompt a download
    - Users should be able to cancel the progress
    - Errors must be properly dispatched to the global handler

### DIFF
--- a/src/components/export.tsx
+++ b/src/components/export.tsx
@@ -165,7 +165,7 @@ const Export = ({ reference, disabled }: ExportProps): JSX.Element => {
     <>
         <Dropdown className='export-menu' onClick={() => setShowTooltip(false)}>
           <OverlayTrigger
-            placement='bottom'
+            placement={ConfigService.appSettings.hideNavbar ? 'left' : 'top-end'}
             overlay={<Tooltip>{MESSAGE_MAP.tooltip.export}</Tooltip>}
             onToggle={(next: boolean) => setShowTooltip(next)}
             show={showTooltip}
--- a/src/components/export.tsx
+++ b/src/components/export.tsx
@@ -184,7 +184,7 @@ const Export = ({ reference, disabled, onCancelDownload }: ExportProps): JSX.Ele
     <>
         <Dropdown className='export-menu'>
           <OverlayTrigger
-            placement='left'
+            placement='bottom'
             overlay={<Tooltip>{MESSAGE_MAP.tooltip.export}</Tooltip>}
           >
             <Dropdown.Toggle
--- a/src/components/export.tsx
+++ b/src/components/export.tsx
@@ -67,6 +67,7 @@ const Export = ({ reference, disabled }: ExportProps): JSX.Element => {
         !ConfigService.chaiseConfig.disableDefaultExport
       );
       
+      // Update the list of templates in UI
       options.push(...templates);
   
       setOptions(options);
@@ -74,6 +75,11 @@ const Export = ({ reference, disabled }: ExportProps): JSX.Element => {
 
   }, [reference]);
 
+  /**
+   * Send the request for export
+   * @param option selected export option
+   * @returns Downloads the selected file
+   */
   const startExporting = (option: any) => () => {
     const formatType = option.type;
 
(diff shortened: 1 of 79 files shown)


src/components/export.tsx
```diff
@@ -1,51 +1,221 @@
...
+            <div
+              className='progress-bar progress-bar-striped active'
+              aria-valuenow={100}
+              aria-valuemax={100}
+              style={{ width: '100%' }}
+            ></div>
+          </div>
+        </Modal.Body>
+        <Modal.Footer>
+          <Button className='chaise-btn chaise-btn-primary' variant='outline-primary' onClick={closeModal}>
```
RFSH, line 171: Use the `chaise-btn chaise-btn-secondary` class for this button (the same as AngularJS version).

src/components/export.tsx
```diff
@@ -1,51 +1,221 @@
...
 
-const Export = ({
-  reference,
-  disabled
-}: ExportProps): JSX.Element => {
+const Export = ({ reference, disabled, onCancelDownload }: ExportProps): JSX.Element => {
+  /**
+   * State variable to open progress modal dialog
+   */
+  const [show, setShow] = useState(false);
```
RFSH, line 45: Use a better name... show-what-exactly?

src/components/export.tsx
```diff
@@ -1,51 +1,221 @@
...
-  )
-
-}
+    <>
+        <Dropdown className='export-menu'>
+          <OverlayTrigger
+            placement='bottom'
+            overlay={<Tooltip>{MESSAGE_MAP.tooltip.export}</Tooltip>}
+          >
+            <Dropdown.Toggle
```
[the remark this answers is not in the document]
RFSH: The button should be disabled if there aren't any option.

RFSH (changes requested):
Please address the commends that I added. Also,
- Hide the tooltip when the dropdown is open (the tooltip is blocking the menu, so you should hide it)
- Remove the outline of the dropdown button (the outline that shows up when users focus on it).

src/components/export.tsx
```diff
@@ -1,51 +1,214 @@
...
 
+const Export = ({ reference, disabled }: ExportProps): JSX.Element => {
+  /**
+   * State variable to open progress modal dialog
+   */
+  const [showProgressModal, setShowProgressModal] = useState<boolean>(false);
+  /**
+   * State variable to export options
+   */
+  const [options, setOptions] = useState<Array>([]);
```
RFSH, line 33: Use `any[]` to be consistent with the `any` typeof `selectedOption`. Eventually when we have proper export object type in ermrestjs, we will fix these.

src/components/export.tsx
```diff
@@ -1,51 +1,214 @@
...
+  const startExporting = (option: any) => () => {
+    const formatType = option.type;
+
+    switch(formatType) {
+      case 'DIRECT':
+        location.href = reference.csvDownloadLink;
+        break;
+      case 'BAG':
+      case 'FILE': 
+        setIsDownloading(true);
```
RFSH, line 92: Remember that using multipe set-states in a callback might cause batching issues. And to be honest you don't need this many state variables.. they all are doing the same thing.

Just set the `selectedOption`.  It makes the code easier to read as well.

src/components/export.tsx
```diff
@@ -1,51 +1,214 @@
...
+  const cancelExport = () => {
+    // Cancel download
+    if (!!exporterObj) {
+      exporterObj.cancel();
+    }
   }
 
+  const closeModal = () => {
+    cancelExport();
+    setShowProgressModal(false);
```
RFSH, line 165: Assuming you remove all the unnecessary state variables and only set the `selectedOption`, then in here you would clear it.

This would also fix a bug in your current implementation where the button stays disabled

RFSH (changes requested):
Please be conservative about state variables. `showProgressModal`, `selectedOption`, and `isDownloading` all are doing the same thing. There's no reason to have all these three.

It's also causing the bug that I mentioned in one of my comments.

RFSH (approved):

jrchudy (commented):
The tooltip position is different than what is on master. While looking into this, I realized that we weren't handling the case where the navbar is hidden if `hideNavbar` query parameter is present. I will make changes to support hiding the navbar and conditionally setting the position of the tooltip.

Only other concern I have is that the modal width is different. We are using `size: 'md'` in angular but setting no size in this case. From reading the documentation for react bootstrap modals:
> size - Render a large, extra large or small modal. When not provided, the modal is rendered with medium (default) size.

It looks like bootstrap changed the definition of `modal-md` in the version we are now using. @RFSH and I should talk about this before merging

jrchudy (approved):

PR #2196, opened 2022-06-24
Table header > create/edit

ravish-kg:
This PR contains changes related to create/edit:

Clicking on "Bulk Edit" should change the current location to the recordedit app.
Clicking on the "Create" button,
Should open the recordedit page in a new tab.
After creating a record, focusing on the recordset page should update the page.

### DIFF
--- a/src/assets/scss/_table-header.scss
+++ b/src/assets/scss/_table-header.scss
@@ -28,4 +28,10 @@
     padding-bottom: 2px;
     margin-right: 4px;
   }
+
+  .pull-right {
+    display: flex;
+    justify-content: flex-end;
+    gap: 8px;
+  }
 }
--- a/src/components/recordset.tsx
+++ b/src/components/recordset.tsx
@@ -10,7 +10,7 @@ import Title from '@isrd-isi-edu/chaise/src/components/title';
 import Export from '@isrd-isi-edu/chaise/src/components/export';
 import ChaiseSpinner from '@isrd-isi-edu/chaise/src/components/spinner';
 import RecordsetTable from '@isrd-isi-edu/chaise/src/components/recordset-table';
-import { attachContainerHeightSensors, attachMainContainerPaddingSensor, copyToClipboard } from '@isrd-isi-edu/chaise/src/utils/ui-utils';
+import { attachContainerHeightSensors, attachMainContainerPaddingSensor, copyToClipboard, deleteCookie, getAllCookies, getCookieName } from '@isrd-isi-edu/chaise/src/utils/ui-utils';
 import { RecordsetConfig, RecordsetDisplayMode } from '@isrd-isi-edu/chaise/src/models/recordset';
 import { isObjectAndKeyDefined } from '@isrd-isi-edu/chaise/src/utils/type-utils';
 import { createRedirectLinkFromPath, getRecordsetLink } from '@isrd-isi-edu/chaise/src/utils/uri-utils';
@@ -122,11 +122,6 @@ const RecordsetInner = ({
    * We have to validate the facets first, and then we can show them.
    */
   const [facetColumnsReady, setFacetColumnsReady] = useState(false);
-
-  /**
-   * State variable to hold newly created referrer ids
-   */
-  const [recordRequests, setRecordRequests] = useState<any[]>([]);
 
   const mainContainer = useRef<HTMLDivElement>(null);
 
@@ -242,29 +237,22 @@ const RecordsetInner = ({
   };
 
   const updateRecords = () => {
-    let completed = 1; 
+    let completed = 0; 
+    const allCookies = getAllCookies();
 
-    for (const referrerId in recordRequests) {
-      // TODO: ask
-      // const cookie = $cookies.getObject(referrerId);
-      // if (cookie) {
-        delete recordRequests[referrerId];
-        // $cookies.remove(referrerId);
+    const recordRequests = allCookies.filter(c=> c.trim().startsWith('recordset-'));
+
+    for (const referrerId of recordRequests) {
+        deleteCookie(getCookieName(referrerId));
         completed += 1;
-      // }
     }
 
     if (completed > 0) {
       const cause = completed ? LogReloadCauses.ENTITY_CREATE : LogReloadCauses.ENTITY_UPDATE;
 
-      // TODO: ask
       update(null, null, true, true, true, false, cause);
     }
   };
-
-  const onRecordRequest = (referrerId: string) => {
-    setRecordRequests([...recordRequests, referrerId]);
-  }
 
   //------------------- UI related functions: --------------------//
 
@@ -474,7 +462,7 @@ const RecordsetInner = ({
             </div>
             {renderSelectedFacetFilters()}
             {renderShowFilterPanelBtn()}
-            <TableHeader config={config} onRecordRequest={onRecordRequest} />
+            <TableHeader config={config} />
           </div>
 
         </div>
--- a/src/components/recordset.tsx
+++ b/src/components/recordset.tsx
@@ -123,6 +123,11 @@ const RecordsetInner = ({
    */
   const [facetColumnsReady, setFacetColumnsReady] = useState(false);
 
+  /**
+   * State variable to hold newly created referrer ids
+   */
+  const [recordRequests, setRecordRequests] = useState<any[]>([]);
+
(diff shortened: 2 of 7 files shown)


src/assets/scss/_table-header.scss
```diff
@@ -28,4 +28,10 @@
     padding-bottom: 2px;
     margin-right: 4px;
   }
+
+  .pull-right {
```
RFSH, line 32: pull-right was a bootstrap3 built-in class and that's why we've used it there. Let's name this something more specific. Something like `chaise-table-header-buttons`

src/components/recordset.tsx
```diff
@@ -234,6 +236,27 @@ const RecordsetInner = ({
     });
   };
 
+  /**
+   * On window focus, remove request and update the page
+   */
+  const updateRecords = () => {
```
RFSH, line 242: Let's call it `onWindowFocus` or simply `onFocus` so it's easier to understand how this function is used.

src/components/table-header.tsx
```diff
@@ -105,26 +185,53 @@ const TableHeader = ({ config }: TableHeaderProps): JSX.Element => {
...
+        <div
+          className='pull-right'
+        >
+          {shouldShowCreateButton() && (
+            <OverlayTrigger
+              placement='bottom-end'
+              overlay={
+                <Tooltip>
+                  Create new{' '}
+                  {reference.displayname && reference.displayname.value}
```
RFSH, line 197: Use the `DisplayValue` component here and pass the `displayname` as the "value" prop.

src/components/table-header.tsx
```diff
@@ -105,26 +185,53 @@ const TableHeader = ({ config }: TableHeaderProps): JSX.Element => {
...
+            </OverlayTrigger>
+          )}
+
+          {/* Edit Button */}
+          {shouldShowEditButton() && (
+            <OverlayTrigger
+              placement='bottom-end'
+              overlay={
+                <Tooltip>
+                  {shouldEditButtonDisabled() ? `Editing disabled when items per page > ${RECORDEDIT_MAX_ROWS}` : 'Edit this page for records.'}
```
RFSH, line 217: The tooltip is not showing up when the button is disabled. To test this add `?limit=201` at the end of your recordset url to force the page size.

src/utils/ui-utils.ts
```diff
@@ -316,4 +316,46 @@ export function convertVWToPixel(value: number) {
 
   const result = (x * value) / 100;
   return result;
+}
```
RFSH, line 319: Let's move all these cookie related functions to a separate file and call it `cookie.ts` and move it to `services` folder.

And the file should define a class and these functions that you defined here should be static properties of it. Something like

```
export default class CookieService {

  static deleteCookie(name:string, path = '/', domain?:string) {
   ...
  }
  
 ...
}
```

src/utils/ui-utils.ts
```diff
@@ -316,4 +316,46 @@ export function convertVWToPixel(value: number) {
...
+      ';expires=Thu, 01 Jan 1970 00:00:01 GMT';
+  }
+}
+
+/**
+ * Get cookie helper function
+ * @param name name of the cookie
+ * @returns cookie 
+ */
+export function getCookie(name:string){
```
RFSH, line 341: Add return type. You should also explain what happens when the cookie doesn't exist (returns false, undefined, or etc)

src/utils/ui-utils.ts
```diff
@@ -316,4 +316,46 @@ export function convertVWToPixel(value: number) {
...
+  return document.cookie.split(';').some(c => {
+      return c.trim().startsWith(name + '=');
+  });
+}
+
+/**
+ * Returns cookie list
+ */
+export function getAllCookies() {
```
RFSH, line 350: Add return type

src/utils/ui-utils.ts
```diff
@@ -316,4 +316,46 @@ export function convertVWToPixel(value: number) {
...
+  return document.cookie.split(';');
+}
+
+/**
+ * 
+ * @param cookie accpets cookie in format of <cookieName>=<value>
+ * @returns cookie name
+ */
+export function getCookieName(cookie:string) {
```
RFSH, line 359: Add return type

src/utils/math-utils.ts
```diff
@@ -20,3 +20,15 @@ export function generateUUID(): string {
...
   return Math.floor((1 + Math.random()) * 0x10000).toString(36);
 }
+
+/**
+ * 
+ * @param min minimum number 
+ * @param max maximum number
+ * @returns random number generated
+ */
+export function getRandomInt(min: number, max: number) {
```
RFSH, line 30: Add return type

src/utils/ui-utils.ts
```diff
@@ -316,4 +316,46 @@ export function convertVWToPixel(value: number) {
 
   const result = (x * value) / 100;
   return result;
+}
+
+/**
+ * Delete cookie helper function
```
RFSH, line 322: Explain what happens if the cookie doesn't exist. Also add return type (`void`).

RFSH (changes requested):
Why did you create this branch based on an old commit? Next time, please make sure you're pull/fetching the react branch before creating a new one out of it.

RFSH (approved):
Thank you for the changes. Everything looks good to me now. 

I should mention that the implemented `onFocus` logic assumes that all the cookies are for this page, which is wrong. This is something that I will have to think more about and fix. And should be ignored while reviewing this PR.

src/components/table-header.tsx
```diff
@@ -105,26 +186,57 @@ const TableHeader = ({ config }: TableHeaderProps): JSX.Element => {
...
```
jrchudy, line 220: Tooltip should be: `Edit this page of records.`. Please change `for` to `of`

src/components/table-header.tsx
```diff
@@ -105,26 +186,57 @@ const TableHeader = ({ config }: TableHeaderProps): JSX.Element => {
...
+                </Tooltip>
+              }
+            >
+              <span>
+                <Button
+                  className='chaise-btn chaise-btn-primary'
```
jrchudy, line 226: On master, this button has another class called `chaise-table-header-edit-link`.

src/components/table-header.tsx
```diff
@@ -105,26 +186,57 @@ const TableHeader = ({ config }: TableHeaderProps): JSX.Element => {
...
+              overlay={
+                <Tooltip>
+                  Create new{' '}
+                  <DisplayValue value={reference.displayname} />
+                </Tooltip>
+              }
+            >
+              <span className='chaise-table-header-buttons-span'>
+                <Button
+                  className={`chaise-btn  ${config.displayMode === RecordsetDisplayMode.FULLSCREEN ? 'chaise-btn-primary' : 'chaise-btn-secondary'}`}
```
jrchudy, line 204: On master, this button has another class called `chaise-table-header-create-link`.

jrchudy (commented):
Only a few minor issues. The buttons are missing a class, the tooltip for bulk edit is incorrect, and the spacing between the buttons is inconsistent with master.

src/assets/scss/_table-header.scss
```diff
@@ -28,4 +28,13 @@
...
     margin-right: 4px;
   }
+
+  .chaise-table-header-buttons {
+    display: flex;
+    justify-content: flex-end;
+  }
+
+  .chaise-table-header-buttons-span {
+    margin: 0px 8px;
```
jrchudy, line 38: Let's change this to `margin: 0px 5px;` to be consistent with master

src/components/table-header.tsx
```diff
@@ -105,26 +186,57 @@ const TableHeader = ({ config }: TableHeaderProps): JSX.Element => {
...
+              overlay={
+                <Tooltip>
+                  Create new{' '}
+                  <DisplayValue value={reference.displayname} />
+                </Tooltip>
+              }
+            >
+              <span className='chaise-table-header-buttons-span'>
+                <Button
+                  className={`chaise-btn  ${config.displayMode === RecordsetDisplayMode.FULLSCREEN ? 'chaise-btn-primary' : 'chaise-btn-secondary'} chaise-table-header-edit-link`}
```
jrchudy, line 204: This one should be `chaise-table-header-create-link`

ravish-kg: I have added class to the button element

PR #2204, opened 2022-06-29
Edit delete

ravish-kg:
This PR has changes:

Related to "edit":
-       Clicking on edit should open the recordedit page with the proper query parameter.
-       For now, you don't need to worry about updating the recordset after the edit is done. I need to think more about it.

Related to "delete":
-  Depending on the confirmDelete chaise-config property, clicking the "delete" button should either send the request or show a confirmation modal.
- The confirmation modal should Specify the tuple.displayname of what the user is attempting to delete.
- Have a Cancel and close button like the master. There should be a callback for this function where we can later log the action.
- Have a delete button that will trigger the delete.

**PENDING:**
- For now, you don't need to worry about updating the recordset after the delete is done. I need to think more about it.
- Setting logStack property [here](https://github.com/informatics-isi-edu/chaise/blob/6ab7ac42abf0845645d7e2b2c4ccc8a335f4c64c/src/components/table-row.tsx#L148,L152)

### DIFF
--- a/src/components/delete-confirmation-modal.tsx
+++ b/src/components/delete-confirmation-modal.tsx
@@ -15,19 +15,23 @@ type DeleteConfirmationModalProps = {
    * prop to trigger on cancel
    */
   onCancel: () => void;
+  /**
+   * children passed to the component
+   */
+  children: any
 };
 
 /**
  * returns Modal Component - Component that renders delete comfirmation dialog
  */
-const DeleteConfirmationModal = ({ show, onDeleteConfirmation, onCancel, ...props }: DeleteConfirmationModalProps) => {
+const DeleteConfirmationModal = ({ show, onDeleteConfirmation, onCancel, children }: DeleteConfirmationModalProps) => {
   return (
     <Modal show={show} onHide={onCancel} keyboard={false}>
       <Modal.Header>
         <Modal.Title>Confirm Delete</Modal.Title>
       </Modal.Header>
       <Modal.Body>
-        <div className='modal-text'>{props && props.children}</div>
+        <div className='modal-text'>{children}</div>
       </Modal.Body>
       <Modal.Footer>
         <Button
--- a/src/components/delete-confirmation-modal.tsx
+++ b/src/components/delete-confirmation-modal.tsx
@@ -1,0 +1,52 @@
+import '@isrd-isi-edu/chaise/src/assets/scss/_export.scss';
+
+import { Button, Modal } from 'react-bootstrap';
+
+type DeleteConfirmationModalProps = {
+  /**
+   * prop to show modal
+   */
+  show: boolean;
+  /**
+   * prop to trigger on delete confirmation
+   */
+  onDeleteConfirmation: () => void;
+  /**
+   * prop to trigger on cancel
+   */
+  onCancel: () => void;
+};
+
+/**
+ * returns Modal Component - Component that renders delete comfirmation dialog
+ */
+const DeleteConfirmationModal = ({ show, onDeleteConfirmation, onCancel, ...props }: DeleteConfirmationModalProps) => {
+  return (
+    <Modal show={show} onHide={onCancel} keyboard={false}>
+      <Modal.Header>
+        <Modal.Title>Confirm Delete</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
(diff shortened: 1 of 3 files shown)


src/components/table-row.tsx
```diff
@@ -337,18 +414,28 @@ const TableRow = ({
...
+        </td>
+      }
+      {renderCells()}
+    </tr>
+    <DeleteConfirmationModal 
+      show={showDeleteConfirmationModal} 
+      onDeleteConfirmation={onDeleteConfirmation}
+      onCancel={onCancel}
+    >
+      Are you sure you want to delete <DisplayValue value={tuple.displayname} />?
```
RFSH, line 436: Instead of passing the children, you should just pass the tuple and let the modal display the body.

src/components/table-row.tsx
```diff
@@ -46,7 +52,25 @@ const TableRow = ({
     maxHeightStyle: defaultMaxHeightStyle
   })
 
+  /**
+   * state variable to open and close delete confirmation modal window
+   */
+  const [showDeleteConfirmationModal, setShowDeleteConfirmationModal] = useState<boolean>(false);
```
RFSH, line 58: Instead of a boolean, you should have an object that if it's defined, it will be the props of delete-confirmation modal.

The props would be,
- The tuple that will be deleted.
- The confirm callback. This callback will be different if 'isUnlink' is true, that's why I suggest storing the callback in the state. So you don't have to store an extra `isUnlink` variable.

src/components/table-row.tsx
```diff
@@ -46,7 +52,21 @@ const TableRow = ({
     maxHeightStyle: defaultMaxHeightStyle
   })
 
+  /**
+   * state variable to open and close delete confirmation modal window
+   */
+  const [showDeleteConfirmationModal, setShowDeleteConfirmationModal] = useState<any>({
```
RFSH, line 58: Add proper type for it and also `show` is not needed. You should just set this variable to null if you want to close the modal. And as long as it has a not-null value, you will show the modal.

```ts
const [showDeleteConfirmationModal, setShowDeleteConfirmationModal] = useState<{
  onConfirm: () => void,
  buttonLabel: string
} | null>(null);
```

src/components/delete-confirmation-modal.tsx
```diff
@@ -0,0 +1,65 @@
...
+          onClick={onConfirm}
+        >
+          {buttonLabel}
+        </Button>
+        <Button
+          className='chaise-btn chaise-btn-secondary'
+          variant='outline-primary'
+          onClick={onCancel}
+        >
+          cancel
```
RFSH, line 58: It should be `Cancel`

src/components/delete-confirmation-modal.tsx
```diff
@@ -0,0 +1,65 @@
...
+   */
+  buttonLabel: string;
+};
+
+/**
+ * returns Modal Component - Component that renders delete comfirmation dialog
+ */
+const DeleteConfirmationModal = ({ show, onConfirm, onCancel, tuple, buttonLabel }: DeleteConfirmationModalProps) => {
+  return (
+    <Modal show={show} onHide={onCancel} keyboard={false}>
```
RFSH, line 36: What's the reason for passing the keyboard prop? We want users to be able to cancel the modal by just pressing escape (and it already is working this way too).

Also while it won't make any difference, for consistently let's specify `size='sm'` here.

RFSH (changes requested):
I asked for some minor changes. Everything else looks good to me.

On Tuesday's meeting we should have a discussion about the behavior of this modal:
- Given that we're changing the UI to mention the tuple's displayname, is displaying rowname enough? Should it be qualified by the table name?
- Do we need a special indicator that we're waiting for delete? On master, we're showing the "data is loading" spinner. Should we do the same? Or should we show a spinner in place of the delete/unlink button?

RFSH:
@jrchudy The PR is ready to be reviewed. As we mentioned in the meeting, I had to adjust the confirmation message to include the table name. While doing so, I made some small refactoring changes as well.

As we discussed previously, with the changes in this PR, the page will not load after editing/deleting. And that's something that I'll work on next. I left a TODO for myself when I get to work on this (we need to discuss the expected behavior while waiting for a delete request).

PR #2205, opened 2022-07-07
Migrate navbar/base-config.spec.js

ravish-kg:
This PR has changes related to navbar e2e testing.

Changes:
The spec file test/e2e/specs/all-features/navbar/base-config.spec.js should be passing based on our current implementation of navbar in react.

### DIFF
--- a/test/e2e/specs/all-features/navbar/base-config.spec.js
+++ b/test/e2e/specs/all-features/navbar/base-config.spec.js
@@ -7,7 +7,6 @@ describe('Navbar ', function() {
         chaisePage.navigate(browser.params.url + "/recordset/#" + browser.params.catalogId + "/product-navbar:accommodation");
         
         navbar = element(by.id('mainnav'));
-        // menu = element(by.id('navbar-menu'));
         menu = element(by.css('.navbar-menu-options'));
         browser.executeScript('return chaiseConfig;').then(function(config) {
             chaiseConfig = config;
@@ -44,18 +43,23 @@ describe('Navbar ', function() {
         });
     });
 
-    // TODO: 
-    // it('for the menu, should generate the correct # of list items based on acls to show/hide specific options', function() {
-    //     var nodesInDOM = menu.all(by.tagName('li'));
-    //     // Count the number of nodes that are being shown (top level and submenus)
-    //     //   - Local: config has 13 but 1 is hidden by ACLs
-    //     //   - CI: config has 13 but 7 are hidden based on ACLs
-    //     var counter = (!process.env.CI ? 12 : 6); // counted from chaise config doc rather than having code count
+    it('for the menu, should generate the correct # of list items based on acls to show/hide specific options', function() {
+        const nodesInDOM = menu.all(by.css('a'));
+        const headers = menu.all(by.css('.chaise-dropdown-header'));
+        // Count the number of nodes that are being shown (top level and submenus)
+        //   - Local: config has 13 but 1 is hidden by ACLs
+        //   - CI: config has 13 but 7 are hidden based on ACLs
+        var counter = (!process.env.CI ? 12 : 6); // counted from chaise config doc rather than having code count
+        let totalNodes = 0;
 
-    //     nodesInDOM.count().then(function(count) {
-    //         expect(count).toEqual(counter, "number of nodes present does not match what's defined in chaise-config");
-    //     });
-    // });
+        nodesInDOM.count().then(function(count) {
+            totalNodes += count;
+        });
+        headers.count().then(function(count) {
+            totalNodes += count;
+            expect(totalNodes).toEqual(counter, "number of nodes present does not match what's defined in chaise-config");
+        });
+    });
 
     it('should prefer markdownName over name when both are defined', function () {
         // option #2 has both name and markdownName defined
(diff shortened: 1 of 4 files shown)


test/e2e/specs/all-features/navbar/base-config.spec.js
```diff
@@ -44,52 +44,59 @@ describe('Navbar ', function() {
...
+        let secondMenuItem = menu.all(by.css('div.nav-item'));
+        expect(secondMenuItem.get(0).getText()).toBe("Test Recordsets", "name was used instead of markdownName");
     });
 
     it('should render a markdown pattern using proper HTML', function () {
         // in ci we don't have the same globus groups so the "show" ACL hides the 3rd link ("Records")
         var idx = (!process.env.CI ? 3 : 2);
         // option #4 has only markdownName defined
-        element.all(by.css('#navbar-menu > li.dropdown')).get(idx).element(by.css("a")).getAttribute('innerHTML').then(function (aInnerHTML) {
+        let lastMenuItem = menu.all(by.css('div.nav-item')).get(1);
```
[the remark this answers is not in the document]
jrchudy: Thinking about this case more, since we are specifically trying to test the "2nd dropdown's displayed value", maybe it makes sense to make this selector more specific. The proposed change removes the need for `idx` since the 1st and 3rd links don't have the `nav-item` class.

jrchudy (changes requested):
Can you add comments to the PR about each change to selectors that were made and why? This way we can determine what your thought process was for updating the tests instead of making changes to the navbar migration implementation. (I should have asked this before reviewing since I've done a bit of digging to help understand why some changes were made)

The reason I ask that is so that we can have a discussion about what might be a better solution and why. In my PR review comments for the first test case to test the "correct # of list items...", I suggested that a small change to the navbar code would have been better than writing a 2nd selector. In some cases, if not all, we want to be preserving the way the test was working before.

In my 2nd comment, the test case was changed to select a subset of navbar items instead of all of them. Part of the test is verifying the "total selection" stays the same (still 4 items being selected) and verifying a specific item in the list (same order) has the expected display value. I realize those extra tests "aren't mentioned as an expectation or comment in the code".

In my 3rd comment, I noticed a similar thing as the 2nd one, where a small change to the code (adding a consistent class to all navbar items) was a simpler fix. But after thinking about it more, the change to that test removed the randomness based on the test environment (`idx = !process.env.CI ? 3 : 2`).

To summarize:
 - [x] inline comments to add `as='a'`
 - [x] inline comments to add `chaise-nav-item` class
   - update selectors from `element.all(by.css('#navbar-menu > li.dropdown'))` to `element.all(by.css('.navbar-menu-options > .chaise-nav-item'))`
   - revert changes in tests to this changed selector and use `menuDropdowns` again
 - [x] fix promise chains with click events and `chaisePage.wait` conditions (function in chaisePage calls `browser.wait`)

test/e2e/specs/all-features/navbar/base-config.spec.js
```diff
@@ -103,37 +105,32 @@ describe('Navbar ', function() {
...
-         var closeLink = element(by.css('.modal-close'));
-         var modalContent = element(by.css('.modal-content'));
-         browser.wait(EC.elementToBeClickable(closeLink), browser.params.defaultTimeout);
-         chaisePage.clickButton(closeLink).then(function(){
+        const closeLink = element(by.css('.modal-close'));
+        const modalContent = element(by.css('.profile-popup'));
 
-            browser.wait(EC.not(EC.presenceOf(modalContent)), browser.params.defaultTimeout);
+         chaisePage.clickButton(closeLink).then(function(){
+            chaisePage.waitForElementInverse(modalContent, 2000);
```
jrchudy, line 133: Instead of `2000` use `browser.params.defaultTimeout`. That's a value we set as part of the setting up the test environment for the wait statements.

test/e2e/specs/all-features/navbar/base-config.spec.js
```diff
@@ -103,37 +105,32 @@ describe('Navbar ', function() {
...
+        link.click().then(function() {
+            const profileLink = element(by.css('#profile-link'));
+            return profileLink.click();
+        }).then(function() {
+            // Get profile card modal element using className profile-popup
+            const modalContent = element(by.css('.profile-popup'));
+            chaisePage.waitForElement(modalContent);
+            expect(modalContent.isDisplayed()).toBeTruthy();
+            done();
         });
```
jrchudy, line 125: Please add the `catch` case back. If something in the code fails, this test will break test execution since "done.fail()" isn't being called which signals for the next `it` to run.

```
.catch(function (err) {
  done.fail();
  console.log(err);
});
```

test/e2e/specs/all-features/navbar/base-config.spec.js
```diff
@@ -57,39 +57,41 @@ describe('Navbar ', function() {
...
+            let editMenu = menu.all(by.css('.chaise-nav-item')).get(3);
+            editMenu.click().then(function() {
+                expect(editMenu.all(by.css('.chaise-dropdown-header')).get(0).getText()).toBe("For Mutating Data", "Sub menu header is incorrect or not showing");
                 return editMenu.all(by.css("a.disable-link"));
             }).then(function (options) {
                 disabledSubMenuOptions = options;
-
                 expect(options.length).toBe(4, "some options are not shown properly");
                 expect(disabledSubMenuOptions[0].getText()).toBe("Edit Existing Record", "the wrong link is disabled or none were selected");
             });
```
jrchudy, line 97: Can you add the catch here too? In this case, `done.fail()` should NOT be included since we aren't using it in this test case to "control" the flow of the tests. 
```
.catch(function (err) {
  console.log(err);
});
```

jrchudy (changes requested):
Tests are passing for me still. Only a couple things to change for consistency and error handling.

jrchudy (approved):

PR #2210, opened 2022-07-11
Migrate navbar/no-logo.spec.js and catalog-chaise-config.spec.js

ravish-kg:
This PR contains related to no-logo and catalog-chaise-config end to end test fixes.

1. I have made changes in "should generate the correct # of list items" test case. Since all elements in the dropdown menu are not visible initially,  click function is needed on loginDropdown element.

2. In test case "should have a disabled link named Disabled Link", I have made changes to the selector. This change is needed as we can have multiple nested submenu inside login dropdown (including div, a, etc). This selector will select all first level children under login dropdown. (USAGE: .dropdown-menu > * )

3. Also, chaisePage.clickButton(loginDropdown) function doesn't seem to work, so I have used loginDropdown.click() (As provided by protractor).


test/e2e/specs/all-features-confirmation/navbar/no-logo.spec.js
```diff
@@ -1,4 +1,4 @@
-const { browser } = require('protractor');
+const { browser, element } = require('protractor');
```
jrchudy, line 1: This doesn't actually need to be added since protractor attached browser/element to the global scope of the testing environment.

test/e2e/specs/all-features-confirmation/navbar/no-logo.spec.js
```diff
@@ -140,11 +143,12 @@ describe('Navbar ', function() {
...
 
             it('first level nested link should inherit newTab value from parent when newTab is undefined and navigate in new tab', function (done) {
-                var recordsetsMenu = menu.all(by.tagName('li')).get(1);
+                // Added chaise-nav-item class to get all first level menu item in navbar
+                var recordsetsMenu = menu.all(by.css('.chaise-nav-item')).get(1);
                 expect(recordsetsMenu.getText()).toBe("RecordSets", "Second top level menu option text is incorrect");
 
                 recordsetsMenu.click().then(function () {
-                    var datasetOption = recordsetsMenu.all(by.tagName('li')).get(0);
+                    var datasetOption = recordsetsMenu.all(by.css('.dropdown-item')).get(0);
```
jrchudy, line 151: This works because each dropdown item is just a link, but this class isn't always present on every dropdown item. I think we want to do `by.tagName('a')` here.

test/e2e/specs/all-features-confirmation/navbar/no-logo.spec.js
```diff
@@ -216,13 +205,36 @@ describe('Navbar ', function() {
         });
 
         it('should have a disabled link named "Disabled Link".', function () {
-            var dropdownOptions = element.all(by.css('.chaise-login-menu > li'));
+            var dropdownOptions = element.all(by.css('.username-display > div.dropdown-menu > *'));
```
jrchudy, line 208: This should be a similar selector that we are using for the other dropdown menus. Above I suggested to do:
`recordsetsMenu.all(by.tagName('a'))`

This selector should work the same way if you get the loginDropdown first:
`element(by.css('.username-display > .dropdown-menu')).all(by.tagName(' a`))`

test/e2e/specs/all-features-confirmation/navbar/no-logo.spec.js
```diff
@@ -8,8 +8,8 @@ describe('Navbar ', function() {
         browser.ignoreSynchronization=true;
         browser.get(browser.params.url + "/recordset/#" + browser.params.catalogId + "/product-navbar:accommodation");
         navbar = element(by.id('mainnav'));
-        menu = element(by.id('navbar-menu'));
-        loginMenu = element(by.css('.chaise-login-menu'));
+        menu = element(by.css('.navbar-menu-options'));
+        loginMenu = element(by.css('.login-menu-options'));
```
jrchudy, line 12: This selector was selecting the `ul` element that contained all the login dropdown options. This selector should mimic that still and be changed to `element.all(by.css('.username-display > .dropdown-menu))` so we are selecting that same element.

test/e2e/specs/all-features-confirmation/navbar/no-logo.spec.js
```diff
@@ -216,13 +205,36 @@ describe('Navbar ', function() {
...
         });
+
+        it('should generate the correct # of list items', function(done) {
+            let loginDropdown = element(by.css('.username-display'));
+
+            // Click is needed because profile menu dropdown will not be present in the DOM initially
+            browser.wait(EC.elementToBeClickable(loginDropdown), browser.params.defaultTimeout);
+            loginDropdown.click().then(function() {
+                let menuItems = loginDropdown.all(by.css('div.dropdown-menu')).get(0);
+                var nodesInDOM = menuItems.all(by.css('a'));
```
jrchudy, line 223: If you fix the `loginMenu` selector properly, this change only need to be:
```
var nodesInDOM = loginMenu.all(by.tagName('a'));

jrchudy (changes requested):

test/e2e/specs/all-features-confirmation/navbar/no-logo.spec.js
RFSH: I think I accidentally added `const { browser } = require('protractor');` and as you mentioned should be removed. Most probably when I was trying to use `browser` VS Code detected it and added this import statement. I'm assuming this is also the case for `element` that Ravish added.

jrchudy: The same keeps happening to me as well

PR #2214, opened 2022-07-15
WIP: Record > HTML structure > main section and related section

ravish-kg:
This PR has changes related to record app page:

This includes,

- Page action buttons
  Just the placeholder 
- Title bar
   The title information that is in the format of <reference.displayname>: <tuple.displayname>
   The create/copy/edit/delete buttons (again, you don't need to worry about functionality).
   The positioning of the title and buttons should behave the same way as the master.
   Show/hide side panel button
- Make sure the button position in both states is the same as the master.
- The main section
- A list of accordions where each is the related entity.
  The position of the arrow and header should be the same as the master.
  Included "explore" button on the top right, and placeholder for the rest of the buttons.
- Footer


src/components/record/record-footer.tsx
```diff
@@ -0,0 +1,18 @@
+import '@isrd-isi-edu/chaise/src/assets/scss/_record.scss';
```
RFSH, line 1: We already have a `footer` component. Please make sure you're looking at recordset implementation and use the same components if they already exits.

src/components/record/record-action-buttons.tsx
```diff
@@ -0,0 +1,67 @@
+import '@isrd-isi-edu/chaise/src/assets/scss/_record.scss';
```
RFSH, line 1: I don't see the point of this component. It's not adding any value and you should just move this to be inside the `record.tsx`.

src/components/record/record.tsx
```diff
@@ -28,36 +36,180 @@ const Record = ({
...
+    }
+  }
 
   useEffect(() => {
     // get the data
     readMainEntity();
   }, []);
 
+  useLayoutEffect(() => {
+    calculateAvailableHeight();
```
RFSH, line 69: There's a `attachContainerHeightSensors` function in `ui-utils.ts` that is basically the React version of the same function that we have in AngularJS. So please use this function instead. You don't need to pass any parameters to this function for your usecase. Please take a look at how I did it in recordset page.

src/assets/scss/_record.scss
```diff
@@ -1 +1,144 @@
 // record component rules
+.top-panel-container {
+    padding-top: 20px;
```
RFSH, line 3: Please make sure you've pulled the latest `react` branch into this branch. If you've done this and you mimicked the same page structure as recordset (and master), then you don't need to manually set all these padding and margins. The [default rules in `app.scss`](https://github.com/informatics-isi-edu/chaise/blob/react/src/assets/scss/app.scss#L413) should take care of this.

src/components/record/record-page-action-buttons.tsx
```diff
@@ -0,0 +1,58 @@
...
+          Show empty sections
+        </Button>
+      </ChaiseTooltip>
+      <ChaiseTooltip
+        placement='bottom-start'
+        tooltip='Click here to show an export format'
+      >
+        <Export
+          reference={null}
+          // disabled={isLoading || !page || page.length === 0}
```
RFSH, line 41: When you moved this to `record.tsx`, you should pass proper values to it and make sure it's working properly in context of record as well. We don't have a `isLoading` for now, so let's just set the disabled state based on `page` (and add a TODO saying that it should also be disabled while loading).

src/assets/scss/_record.scss
```diff
@@ -1 +1,144 @@
...
+            }
+    
+            &:not(.collapsed)::after {
+                content: '\f078'; // fa-solid fa-chevron-down
+            }
+        }
+    }
+}
+
+.footer-container {
```
RFSH, line 106: If you're using the same Footer component then this should be removed.

src/assets/scss/_record.scss
```diff
@@ -1 +1,144 @@
...
+
+.top-flex-panel {
+    position: relative;
+    display: flex;
+    flex: 1;
+    -webkit-box-flex: 1;
+}
+
+.chaise-btn-icon {
+    margin-right: 4px;
```
RFSH, line 131: You realize that this will apply to all the instances of this class, right? If you just want to do this for specific buttons in specific container, please add them. Otherwise general rules like this should be in `app.scss`. In this case you should either remove this rule or just make it more specific like I suggested.

src/assets/scss/_record.scss
```diff
@@ -1 +1,144 @@
...
+}
+
+.pull-right {
```
RFSH, line 134: Move this to `app.scss`.

RFSH (changes requested):
In general, please make sure you're mimicking the same structure as master. This includes all the class names and ids (as well as applying more or less the same CSS rules). This would reduce the amount of changes needed to CSS and testing.

I'm not going to go over the whole structure, but as an example the title elements have `entity-subtitle` and `entity-title` ids that are missing from here.

PR #2269, opened 2023-01-14
Foreignkey input support

RFSH:
This PR will introduce a `ForeignkeyField` component. Given that we decided to improve the AngularJS behavior, and I made many changes that would need review, I created this PR.

In summary,
- I had to add some more props to `InputSwitch` for foreignkey inputs. I left a TODO in there, but I think what I have is good for now. The added props are more or less the same as what we have in AngularJS.
- The `domain-filter` logic needs all the raw data of a foreignkey table. We also need to show the row-name to the users, and not the saved raw value.
  - [`foreignKeyData`](https://github.com/informatics-isi-edu/chaise/blob/react-recordedit-fk/src/providers/recordedit.tsx#L45,L51) has been added to `RecordeditContext` to capture the data (it's a `useRef`).
  - In edit mode, we already have access to this data through the main request. So we don't need any further changes during the initial load.
  - In create mode, if a column has prefilled or default values, we should fetch the foreignkey data. 
    - At first, I implemented this logic in the `populateCreateInitialValues` function. But given that this function is synchronous, I created a function called [`getPrefilledDefaultForeignKeyData`](https://github.com/informatics-isi-edu/chaise/blob/react-recordedit-fk/src/providers/recordedit.tsx#L504,L593) in `RecordeditProvider`, so `RecordeditComponent` can call it.
    -  To be able to change the displayed values in the form, I had to pass the `setValue` function of `react-hook-forms`'s `FormContext` to the function mentioned above. It looks a bit ugly, but I think it should be fine and I couldn't find a better way.
    - Unlink AngularJS, I'm making sure to send foreignkey requests only if any input relies on `domain-filter`.
    - Unlike AngularJS, I'm showing an overlay for inputs that rely on `ForeignKeyData`. I added `waitingForForeignKeyData` for the spinner that is displayed while requests are pending.
  - When users pick a new foreignkey, I'm updating the `foreignKeyData` at the same time as the displayed value in the input.

The code is complete and works based on my testing. But I still need to fully test if the changes have affected other columns (apart from integers which I already did). Also, I got some warning during build in dev mode after including the `RecordsetModal` in `ForeignkeyField`. So I had to make a small adjustment to the webpack. While I think my change should be good, I'll need to test it a bit more.


RFSH:
While testing, I found some bugs that I fixed in my latest commit. Summary of changes:

- Moving the `column-cell` to be an outer div of `InputSwitch` broke the height logic. So I reverted that change and moved the spinner inside the `foreignkey-field.tsx`.
- Added proper tooltip and disabled logic to submit button.
- Made sure the default and prefill logic worked in all cases. 
  - In the AngularJS version, we're abusing `submissionRows` by storing the raw foreignkey column values. Since we don't have that in React, we're keeping everything in the form values (by doing `FormContext.setValue`). I found some cases where I didn't properly do this, so I had to make some adjustments.
  - Changed the `populateSubmissionRow` and `copyForm` logic to use the `reference.activeList.allOutBounds` for foreignkey values, as some prefilled foreignkeys might not be part of the visible columns.

PR #2289, opened 2023-04-01
handle date and time inputs manually

RFSH:
When we migrated the recordset app, we used the browser's `date` and `time` inputs for facet range pickers.

But now that we've used this feature more and discussed it, we prefer the old AngularJS method where we used `text` inputs with special masks and validators. The following are some of the issues with `date`/`time` input type:
  - The behavior is not consistent between browsers. It also shows different formats based on user locale (e.g., 24hr time vs. 12hr), which could confuse UX.
Partial values were causing issues in both date and time.
  - We cannot easily support multiple formats for time.
  - We have to test the validation and date/time pickers manually.

Therefore, this will modify date and time inputs into text inputs with special masks and validators (the same as AngularJS) implementation.

Notes and other changes,
  - In AngularJS implementation, we allowed 24hr format and had AM/PM toggle. Since this would leak to confusing UX, I didn't implement AM/PM toggle (this is consistent with range-input behavior as in AngularJS it was just 24hr format).
  - Added a mask to date inputs using [imaskjs](https://imask.js.org/).
  - Changed the DateTimeField component to use DateField instead of repeating the same structure in two places.
  - Changed RangeInputs to fix a side effect where the facet wasn't updating the min/max displayed on the inputs.

**Pending**:
- There's a bug in my implementation where even though I'm showing the errors for date-time-field properly, the react-hoo-form submit is firing the successful callback and submitting the form (the value is empty in this case, so it's not a usage issue).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
         "react-dom": "^18.2.0",
         "react-error-boundary": "^3.1.4",
         "react-hook-form": "^7.36.1",
+        "react-imask": "^6.5.0-alpha.0",
         "react-plotly.js": "^2.5.1",
         "sass": "^1.58.3",
         "sass-loader": "^13.2.0",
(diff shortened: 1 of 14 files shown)


RFSH:
To fix the bug that I mentioned above, it seems like that's the expected behavior. Calling `setError` will not invalidate the whole form. The form validation is only based on the rules that are passed when registering the form, so I'm now setting the input to an invalid value that will trigger the validator on the input.

Issue #2306, opened 2023-04-25
Filtering on the Date Type


svoinea:
We have at https://dev.pdb-dev.org/chaise/recordset/#99/PDB:entry@sort(RCT::desc::,RID) `2` records that have the `Submitter Flag Date` equal to `2023-04-25`.
The filter panel lokks:
<img width="423" alt="Initial" src="https://user-images.githubusercontent.com/1609577/234425817-0636965d-3600-4602-9f41-b19eea8927ab.png">

Is it OK on the `x` coordinate the date `2023-04-24` to appear twice? Also the date `2023-04-25`?
Is it OK on the `y` coordinate to appear `1` twice and `2` twice?

When we are checking the box at the specified range, we are getting:
<img width="422" alt="Range" src="https://user-images.githubusercontent.com/1609577/234425869-88562bb9-8272-4214-820c-4895933dd162.png">

jrchudy:
This is happening for the `yaxis` because plotly is trying to display more "ticks" than it needs to. By turning the "tick" formatting off, I was able to see the yaxis labels as `[0, 0.5, 1, 1.5, 2]`. The formatting was rounding up these half values.

In the old implementation, plotly was automatically setting `tickmode: linear` based on the data type used for the axis.

This is expected in the case of the `xaxis` in very small ranges of data.

jrchudy:
@svoinea Changes to fix this issue have been merged into master and should be available after the next hourly update.

svoinea:
@jrchudy I have tested and looks good.

PR #2307, opened 2023-04-27
Range picker bugs

jrchudy:
This PR fixes issue #2306. 

Adding `tickmode: 'linear'` to the `yaxis` communicates to plotly that the values displayed for `yaxis` should be a linear relation. the other bug in this PR was checking the difference between min and max incorrectly by doing a conversation and diff instead of using  the moment API, `isBefore`.

Another bug was found while trying to fix the other bugs, we were allowing users to "zoom" in the plot using the builtin plotly functionality when the plot couldn't be zoomed anymore (providing a confusing UX).

### DIFF
--- a/src/components/faceting/facet-range-picker.tsx
+++ b/src/components/faceting/facet-range-picker.tsx
@@ -84,7 +84,7 @@ const FacetRangePicker = ({
 
   const [compState, setCompState] = useState<RangePickerState>(() => {
     const defaultPlotLayout: PlotlyLayout = {
-      autosize: true,
+      // autosize: true,
       height: 150,
       margin: {
         l: 40,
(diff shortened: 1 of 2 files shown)


src/components/faceting/facet-range-picker.tsx
```diff
@@ -84,7 +84,7 @@ const FacetRangePicker = ({
 
   const [compState, setCompState] = useState<RangePickerState>(() => {
     const defaultPlotLayout: PlotlyLayout = {
-      autosize: true,
+      // autosize: true,
```
RFSH, line 87: Please comment why you're uncommenting it in the code as well.

src/components/faceting/facet-range-picker.tsx
```diff
@@ -107,7 +108,8 @@ const FacetRangePicker = ({
       yaxis: {
         fixedrange: true,
         zeroline: true,
-        tickformat: ',d'
+        tickformat: ',d',
+        tickmode: 'linear'
```
RFSH, line 112: Please comment why you added this it in the code as well.

PR #2321, opened 2023-06-08
Document how to add chaise urls to GA report

RFSH:
We recently realized that Google Analytics is not reporting URL fragments. So I tried to figure out how to fix that and modified the documentation to include the new information.

As part of this, I had to write a JavaScript script to extract the URL path we care about. That's why instead of directly pushing the code I created this PR so @jrchudy  can review the chagnes.

P.S. I'm not sure if having the code directly in the markdown page is good or bad. GitHub offers a "copy" button so they can just copy and paste. But it's making the document very lengthy.


jrchudy (approved):
Looks good to me for the chaise apps and webapps

Issue #2189, opened 2022-06-15
Optimize position of login popup for multi-monitor setups


RFSH:
Recently I changed where we're positioning the login popup as part of [this commit](https://github.com/informatics-isi-edu/chaise/commit/3ca897a2451595c2142c624fb8d1af12641f7f83).

While this will ensure the popup is displayed in the center of the screen for typical single-monitor setups, it fails to position it in a multi-monitor setup.

This behavior is also not consistent between browsers,

- In Chrome, if you open the popup from your primary monitor, it correctly positions it. But if you do it from the secondary monitors, it will fail to center the popup (it shows the popup on the same monitor).

- In Firefox, as long as you have more than one monitor, it fails to position the popup properly. It also opens the popup on a different monitor than the one you're looking at.

Therefore, we should look into other ways to position the popup in multi-monitor setups. I tried some of the solutions mentioned [here](https://stackoverflow.com/questions/16363474/window-open-on-a-multi-monitor-dual-monitor-system-where-does-window-pop-up) and [here](https://stackoverflow.com/questions/31859280/dynamic-height-in-javascript-popup-window). But all of them are making assumptions about the number of monitors or their orientation.

PoornimaVKrishnan:
Current Implementation :
We are subtracting screen width and popup width and dividing it by two and giving that as the left position to centre the popup. This approach will not work in multi-monitor setups as we need to add primary screen width to the left offset.
const popupWidth = Math.min(800, screen.availWidth);
const popupHeight = Math.min(750, screen.availHeight)
const topOffset = 50;
const popupLeft = (screen.availWidth - popupWidth) / 2;
// top should just have some small offset if there's available space
const popupTop = (topOffset + popupHeight) < screen.availHeight ? topOffset : 0;
const win = window.open('', '_blank', `width=${popupWidth},height=${popupHeight},left=${popupLeft},top=${popupTop}`);
// focus on the opened window
win?.focus(); 
// keep track of the window, so we ensure only one is opened at a time. 
popupWindowRef.current = win;
_logInHelper(win, postLoginCB, 'popUp', null, logAction);

Different Approaches I tried:
1) Approach: Tried calculating popup position based on window resolutions.
     Problems Faced: It was causing problems when it is a single screen and if window is minimised the pop takes up the entire screen area.
Link referred: https://stackoverflow.com/questions/4068373/center-a-popup-window-on-screen

2) Approach:

[...]

PR #2323, opened 2023-06-20
Optimize position of login popup for multi-monitor setups

PoornimaVKrishnan:
This PR is to optimize position of login popup for multi-monitor setups. It centre aligns the popup in chrome and firefox also opens the pop up in the right screen.

The solution for this is to calculate the left and top positions of the popup from the screen resolutions. The solution I referred to: https://stackoverflow.com/questions/4068373/center-a-popup-window-on-screen

### DIFF
--- a/src/providers/authn.tsx
+++ b/src/providers/authn.tsx
@@ -175,15 +175,19 @@ export default function AuthnProvider({ children }: AuthnProviderProps): JSX.Ele
     const topOffset = 50;
     const popupTop = (topOffset + popupHeight) < screen.availHeight ? topOffset : 0;
 
+   
+    const screenWidth = window.screen.availWidth;
+     
     /*
-      * This is to get the screen left offset using availLeft - It represents the left  position
+      * This is to get the screen left offset using screenLeft - It represents the left  position
       * of the available screen space, excluding areas occupied by other system elements .
       * We need to add those so that it  calculates the centre position and aligns the popup in the centre. 
-      * We add Math.max so that it takes care if the calculation comes to negative. This can happen when screen size
-      * is smaller than popup size.
     */
-    const left = Math.max(0, (screen.availWidth - popupWidth) / 2 + screen.availLeft);
-    const openedWindow = window.open('', '_blank', `width=${popupWidth},height=${popupHeight},left=${left},top=${popupTop}`);
+    const leftOffset = Math.floor((screenWidth - popupWidth) / 2) + window.screenLeft ;
+
+    // Open the popup window to the calculated center position
+    const openedWindow = window.open('', '_blank', `width=${popupWidth},height=${popupHeight},left=${leftOffset},top=${popupTop}`);
+    // Focus the window opened
     if (openedWindow && openedWindow.focus) {
       openedWindow.focus();
     }
--- a/src/providers/authn.tsx
+++ b/src/providers/authn.tsx
@@ -172,17 +172,20 @@ export default function AuthnProvider({ children }: AuthnProviderProps): JSX.Ele
 
     const popupWidth = Math.min(800, screen.availWidth);
     const popupHeight = Math.min(750, screen.availHeight);
+    const topOffset = 50;
+    const popupTop = (topOffset + popupHeight) < screen.availHeight ? topOffset : 0;
 
-    const primaryScreen = window.screen;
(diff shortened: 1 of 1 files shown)


src/providers/authn.tsx
```diff
@@ -170,31 +170,29 @@ export default function AuthnProvider({ children }: AuthnProviderProps): JSX.Ele
...
 
-    // left should be in the middle of the screen
-    const popupLeft = (screen.availWidth - popupWidth) / 2;
-    // top should just have some small offset if there's available space
-    const popupTop = (topOffset + popupHeight) < screen.availHeight ? topOffset : 0;
+    const primaryScreen = window.screen;
 
+    // Calculate the center position based on the primary screen dimensions and popup size
+    const centerLeft = Math.max(0, (primaryScreen.availWidth - popupWidth) / 2 + primaryScreen.availLeft);
+    const centerTop = Math.max(0, (primaryScreen.availHeight - popupHeight) / 2 + primaryScreen.availTop);
```
RFSH, line 180: Let's leave this to what it was. We still want the popup to show up closer to the top of the page, so it's more visible.

RFSH (changes requested):
The link that you added to PR is to the question not the answer. So please add the link to the answer.

Also `availLeft` and `availTop` are non-standard. So if we can avoid them, we should. What's the point of having those in the calculation?

You can see that in [here](https://developer.mozilla.org/en-US/docs/Web/API/Screen/availLeft) it's suggesting that we should avoid using this attribute if we can.

PoornimaVKrishnan:
> The link that you added to PR is to the question not the answer. So please add the link to the answer.
> 
> Also `availLeft` and `availTop` are non-standard. So if we can avoid them, we should. What's the point of having those in the calculation?
> 
> You can see that in [here](https://developer.mozilla.org/en-US/docs/Web/API/Screen/availLeft) it's suggesting that we should avoid using this attribute if we can.

The solution I pointed out. It is what I referred. I googled on how else to get the screen left offset. And went through MDN docs and saw availLeft and availTop. We need to add those so that it aligns in the centre. It represents the left position of the available screen space, excluding areas occupied by other system elements.
I have added jsdoc that explains everything, And on the position from top have reverted changes as we need  popup to show up closer to the top of the page.
When going through all the solutions I realised that we need to add a left offset. I avoided using window.screenLeft or window.screenX to solely depend on screen calculations instead of window

src/providers/authn.tsx
PoornimaVKrishnan: reverted

src/providers/authn.tsx
```diff
@@ -170,31 +170,29 @@ export default function AuthnProvider({ children }: AuthnProviderProps): JSX.Ele
...
+    const popupHeight = Math.min(750, screen.availHeight);
 
-    // left should be in the middle of the screen
-    const popupLeft = (screen.availWidth - popupWidth) / 2;
-    // top should just have some small offset if there's available space
-    const popupTop = (topOffset + popupHeight) < screen.availHeight ? topOffset : 0;
+    const primaryScreen = window.screen;
 
+    // Calculate the center position based on the primary screen dimensions and popup size
+    const centerLeft = Math.max(0, (primaryScreen.availWidth - popupWidth) / 2 + primaryScreen.availLeft);
```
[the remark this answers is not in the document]
PoornimaVKrishnan: have added jsdoc

RFSH (changes requested):
We want the popup to be displayed in the middle of the screen (horizontally). The position of the popup should not be dependent on the window's position or size. 

The previous implementation with `availLeft` was doing what we wanted. But the API is non-standard, and therefore we should avoid it.


Your new implementation is not doing what we want. In scenarios where the left of your browser is not stretched to the left, you can see that the popup is not centered anymore, and it's taking the position of your current window into account.

This is a simple task in terms of expected behavior. I don't understand why we have to go back and forth multiple times. Make sure you're testing your code instead of just pushing something and asking for feedback.

RFSH:
Since @PoornimaVKrishnan couldn't find any solution that works consistently, we decided to just close this PR.

Refer to [this comment](https://github.com/informatics-isi-edu/chaise/issues/2189#issuecomment-1605178843) for more information.

PR #2324, opened 2023-06-22
Change behavior of some chaise-config properties (ermrestLocation, defaultCatalog, defaultTables)

### LINKED ISSUE #2317: Avoid hardcoding location of backend services
While currently all `hatrac`, `authn`, and `ermest` services are installed under a predefined location in the same server, this assumption might eventually not be the case anymore.

Just like we have a [`ermrestLocation`](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/chaise-config.md#ermrestlocation) chaise-config property, we might want to include other properties for other services.

As part of this issue, we may want to experiment and see what other changes we could make in chaise/ermrestjs to allow the installation of frontend components in a different server or even locally.

RFSH:
This PR was initially created to change the behavior of `ermrestLocation`. 

We're using `ermestLocation` to find where ermrest is, so allowing it on the `chaise-config` annotation doesn't make sense. We should only honor it when it's part of `chaise-config.js`. After discussing it, we decided to do the same thing with `defaultCatalog`.

We also didn't like how in `defaultTables` we're supposed to encode the catalog numbers and decided to deprecate it and replace it with `defaultTable`. It's more or less the same property. The only difference is that now the catalog number is implied. 

Summary of code changes:

- Added a concept of `CHAISE_CONFIG_STATIC_PROPERTIES` properties. These properties will only be allowed on the `chaise-config.js` file (`ermestLocation` and `defaultCatalog`). 
- Modified `_setChaiseConfig` to ignore the "static" properties from the annotation.
- Added a new `ConfigService.ERMrestLocation` API and used it in different places in chaise that need the ermrest location.
- Modified `authn`'s `serviceURL` to use `ConfigService.ERMrestLocation` similar to how we're doing this for hatrac in ermrestjs.
- Changed usage of `defaultTables` to use `defaulTable` instead.

Closes #2317

### DIFF
--- a/src/utils/uri-utils.ts
+++ b/src/utils/uri-utils.ts
@@ -164,6 +164,18 @@ export function chaiseURItoErmrestURI(location: Location, dontDecodeQueryParams?
 
     // there is no '/' character (only a catalog id) or a trailing '/' after the id
     if (hash.indexOf('/') === -1 || hash.substring(hash.indexOf('/')).length === 1) {
+      /**
+       * if the hash was actually just query parameter, don't even attempt the default logic
+       *
+       * this function is used in every react app as part of config.ts. in some cases (like help app),
+       * the url doesn't have any hash fragment and has only query parameter. therefore we should
+       * throw an error here since the combination of chaise-config properties might accidently
+       * cause side effects.
+       */
+      if (isQueryParameter) {
+        throw new ConfigService.ERMrest.MalformedURIError(catalogMissing);
+      }
+
       // check for default Table
       if (chaiseConfig.defaultTable) {
         const tableConfig = chaiseConfig.defaultTable;
--- a/src/utils/uri-utils.ts
+++ b/src/utils/uri-utils.ts
@@ -183,7 +183,7 @@ export function chaiseURItoErmrestURI(location: Location, dontDecodeQueryParams?
     }
   }
 
-  const baseUri = chaiseConfig.ermrestLocation;
+  const baseUri = ConfigService.ERMrestLocation;
   const path = `/catalog/${catalogId}/entity${hash}`;
 
   return {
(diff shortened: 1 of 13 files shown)


src/utils/uri-utils.ts
```diff
@@ -164,6 +164,18 @@ export function chaiseURItoErmrestURI(location: Location, dontDecodeQueryParams?
...
     if (hash.indexOf('/') === -1 || hash.substring(hash.indexOf('/')).length === 1) {
+      /**
+       * if the hash was actually just query parameter, don't even attempt the default logic
+       *
+       * this function is used in every react app as part of config.ts. in some cases (like help app),
+       * the url doesn't have any hash fragment and has only query parameter. therefore we should
+       * throw an error here since the combination of chaise-config properties might accidently
+       * cause side effects.
+       */
+      if (isQueryParameter) {
```
RFSH, line 175: To get the catalog id, we're calling `chaiseURItoErmrestURI` function. Since in Chaise we wanted to handle urls like `/chaise/recordset?2/schema:table`, this function will assume query parameters are fragments. And then will try to find the catalog id from the query parameter.

So, in the help app case, a URL like `chaise/help/?page=chaise%2Fmarkdown-help` was returning `page=chaise%2Fmarkdown-help` as the catalog number. But since the schema/table names were missing, it would go into the `defaultTables` logic if the chaise-config property was present (it would throw an error otherwise). We didn't catch this before because `defaultTables` definition has the catalog numbers in it, so it would never match these cases. Now that I removed the catalog number from the definition, it would just assume that the catalog is correct and it would go into this path and add the `defaulTable` to the URL.

That's why I added this check here.

RFSH:
@jrchudy I realized my changes broke the `help` app, so I pushed another change to fix this. I added a comment on the line that I changed with more information.

It seems like [saucelabs is having issues](https://status.saucelabs.com/incidents/zd1b297gschc), so I stopped the latest build and will try again later when the issue is resolved. Other than that, the PR is ready to be reviewed/merged.

jrchudy (approved):

PR #2327, opened 2023-07-05
Make related table actions responsive

PoornimaVKrishnan:
This PR focuses on making the related main section and table section buttons responsive based on available space. 

Other changes:
- Add a Bulk edit button to related entities.
- Change the icon of "custom mode"

### DIFF
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -114,7 +114,7 @@
     .related-table-actions {
       float: right;
       display: flex;
-      align-items: baseline;
+      align-items: center;
 
       // make sure the content is not touching the buttons
       padding-left: 2px;
@@ -148,6 +148,10 @@
         top: -15px;
       }
 
+      .dropdown {
+        min-width: 50px;
+      }
+
       .dropdown-item {
         padding: 0;
 
@@ -166,7 +170,20 @@
     .table-header-actions {
       width: 100%;
       justify-content: end;
+      margin-top: -34px;
     }
+
+    .chaise-accordion-header {
+      display: block !important;
+
+      .chaise-accordion-header-buttons {
+        display: flex;
+        margin-top: auto !important;
+        margin-left: 75px;
+        box-sizing: border-box;
+      }
+    }
+
 
     .chaise-table-top-scroll-wrapper {
       z-index: 1;
@@ -183,7 +200,7 @@
     .chaise-accordion-header {
       .chaise-accordion-displayname {
         white-space: nowrap;
-        min-width: 25%
+        max-width: 25%
       }
 
       white-space: unset !important;
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -145,6 +145,13 @@
         border-color: #4674a7;
       }
 
+      .dropdown-toggle-table-mobile {
+        position: absolute;
+        right: 0;
+        top: -15px;
+        margin-right: 2rem;
+      }
+
       .dropdown-item {
         padding: 0;
 
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -11,7 +11,7 @@
           height: variables.$btn-height;
           margin: 0;
 
-          > div {
+          >div {
             display: flex;
           }
 
@@ -33,6 +33,7 @@
             .page-title-container {
               width: 100%;
               display: inline-block;
+
               #page-title {
                 margin: 0px;
               }
@@ -43,15 +44,21 @@
               padding-top: 7px;
               width: 330px;
             }
-          } // end .entity-display-header
-        } // end .title
+          }
+
+          // end .entity-display-header
+        }
+
+        // end .title
 
         .show-toc-btn {
           margin-top: 10px;
           padding-left: 2px;
           padding-bottom: 0px;
         }
-      } // end .top-right-panel
+      }
+
+      // end .top-right-panel
     }
   }
 
@@ -61,7 +68,7 @@
     }
 
     /** bottom left panel **/
-    > .side-panel-resizable {
+    >.side-panel-resizable {
 
       .side-panel-container {
         height: 100%;
@@ -84,7 +91,7 @@
           cursor: pointer;
           font-size: variables.$h4-font-size;
 
-          & > a.empty-toc-heading {
+          &>a.empty-toc-heading {
             opacity: 0.55;
           }
 
@@ -119,44 +126,51 @@
         text-decoration: none;
       }
     }
+
     .related-table-actions {
       display: flex;
       align-items: baseline;
+
       .button-wrapper {
         height: 30px;
         overflow: hidden;
         position: relative;
         text-align: right;
       }
+
       .dropdown-toggle-table {
         color: #4674a7;
         background-color: #fff;
         cursor: pointer;
         border-color: #4674a7;
       }
+
       .dropdown-item {
         padding: 0;
+
         .dropdown-button {
           .dropdown-button-text {
             margin-left: 4px;
           }
         }
       }
-   }
-   
+    }
+
     .table-header-actions {
       width: 100%;
       justify-content: end;
     }
+
     .chaise-accordion-header {
       .chaise-accordion-displayname {
         white-space: nowrap;
         min-width: 25%
       }
+
       white-space: unset !important;
       align-items: center;
     }
-   
+
     /******Related Entities in record **************/
 
     .row-focus {
@@ -198,6 +212,7 @@
     }
 
     .top-right-panel {
+
       //Action button top right corner
       .page-action-btns {
         display: none;
@@ -292,12 +307,12 @@
         border-bottom: none;
       }
 
-      > thead > tr > th,
-      > tbody > tr > th,
-      > tfoot > tr > th,
-      > thead > tr > td,
-      > tbody > tr > td,
-      > tfoot > tr > td {
+      >thead>tr>th,
+      >tbody>tr>th,
+      >tfoot>tr>th,
+      >thead>tr>td,
+      >tbody>tr>td,
+      >tfoot>tr>td {
         border-top: none;
         border-right: 1px solid #bbbbbb;
       }
@@ -342,7 +357,7 @@
 
     //Removes bubbles and represents the vocab as a dot separated list
     .related-section-container {
-      .vocab + .vocab::before {
+      .vocab+.vocab::before {
         content: "\2022" !important;
         padding-right: 6px;
       }
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -131,7 +131,6 @@
       .button-wrapper {
         height: 30px;
         overflow: hidden;
-        position: relative;
         text-align: right;
       }
 
@@ -167,23 +166,14 @@
       }
     }
 
+    .main-section-actions {
+      position: static;
+    }
+
     .table-header-actions {
       width: 100%;
       justify-content: end;
-      margin-top: -34px;
     }
-
-    .chaise-accordion-header {
-      display: block !important;
-
-      .chaise-accordion-header-buttons {
-        display: flex;
-        margin-top: auto !important;
-        margin-left: 75px;
-        box-sizing: border-box;
-      }
-    }
-
 
     .chaise-table-top-scroll-wrapper {
       z-index: 1;
@@ -200,7 +190,6 @@
     .chaise-accordion-header {
       .chaise-accordion-displayname {
         white-space: nowrap;
-        max-width: 25%
       }
 
       white-space: unset !important;
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -167,6 +167,7 @@
     }
 
     .main-section-actions {
+      // We are adding this for the spinner to come above the buttons
       position: static;
     }
 
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -113,6 +113,8 @@
 
     .related-table-actions {
       float: right;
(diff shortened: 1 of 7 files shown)


src/components/record/related-table-actions.tsx
```diff
@@ -496,50 +603,293 @@ const RelatedTableActions = ({
...
+  /*
+   * This hook is to push the buttons to the dropdown if it exceeds the container width.
+   * We are using ResizeSensor to listen to the resize event.
+   */
+  useLayoutEffect(() => {
+    const mainnav = document.querySelector('#navheader') as HTMLElement;
+    if (!buttonRef.current) return;
+
+    const calculateButtons = () => {
+      setIsMobile(window.innerWidth < 600);
```
RFSH, line 763: It doesn't makes sense to have a ResizeSensor and also something special based on the window width. Remember that window width doesn't mean much in this case, as you can resize the side panel (toc) without changing the window width.

So it should be based on the width of your actual container, not the whole window.

src/components/record/related-table-actions.tsx
```diff
@@ -445,48 +532,68 @@ const RelatedTableActions = ({
...
-  return (
-    <>
-      <div className='related-table-actions' ref={container}>
-        {relatedModel.canCreate &&
-          <ChaiseTooltip
-            placement='top'
-            tooltip={renderCreateBtnTooltip()}
-          >
+  const renderButtons = () => {
+    return (
```
RFSH, line 589: As far as I could tell, you're doing the exact same thing as `renderButton`. I see a lot of duplicated code here. Ideally there should be just one function rendering the button. So ideally this function should call your `renderButton` function.

src/components/record/related-table-actions.tsx
```diff
@@ -496,50 +603,293 @@ const RelatedTableActions = ({
...
+              tableHeaderButtonsToAddToDropdown.push(button);
+            }
+          });
+          setButtonsInDropdown(tableHeaderButtonsToAddToDropdown);
+        }
+      }
+    };
+    calculateButtons();
+
+    const mainNavResizeSensor = new ResizeSensor(mainnav as Element, () => {
```
RFSH, line 801: This doesn't make any sense. We want to have the `ResizeSensor` on an element where its height/width change will affect the buttons. So it should be on either the parent of the button containers, or some other elements that affect the container. Let's use `.main-container` for now.

src/components/record/related-table-actions.tsx
```diff
@@ -28,48 +34,73 @@ import { ConfigService } from '@isrd-isi-edu/chaise/src/services/config';
...
 import { getRandomInt } from '@isrd-isi-edu/chaise/src/utils/math-utils';
 import { fireCustomEvent } from '@isrd-isi-edu/chaise/src/utils/ui-utils';
 import Q from 'q';
+import ResizeSensor from 'css-element-queries/src/ResizeSensor';
 
 type RelatedTableActionsProps = {
-  relatedModel: RecordRelatedModel
-}
+  relatedModel: RecordRelatedModel;
+  istableHeader?: boolean;
```
RFSH, line 54: You don't need this new prop. Just use `relatedModel.isInline` to check if this is used in the main section or the related section.

src/components/record/related-table-actions.tsx
RFSH: By doing this, we will avoid issues where we forget to update one of them. For example right now these two functions are returning slightly different results for "Bulk Edit" button.

src/components/record/related-table-actions.tsx
```diff
@@ -496,50 +603,293 @@ const RelatedTableActions = ({
...
+            <Dropdown.Toggle
+              disabled={false}
+              className='chaise-btn chaise-btn-primary dropdown-toggle-table'
+            >
+              <span className='fa fa-angle-double-right'></span>
+            </Dropdown.Toggle>
+            <Dropdown.Menu>
+              {buttonsInDropdown.map((button: any, index) => (
+                <Dropdown.Item
+                  className={''}
```
RFSH, line 860: Add `as={'li'}` here so it doesn't wrap elements inside a link.

```
<Dropdown.Item
  key={`dropdown-button-${index}`}
  as={'li'}
  onClick={(e: any) => handleDropDownClick(e, button)}
/>
```

RFSH: Please make sure there aren't any errors both in dev and production build mode. Because of this, React was complaining about usage of `a` tag inside another `a` tag.

RFSH (changes requested):
Please comment your code. I barely see any comments in your new changes. You need to explain how the logic of pushing buttons to the dropdown works, among other things.


Also the "Bulk Edit" button should be a link (similar to "Explore" button). You can use `usedRef.contextualize.entryEdit.appLink` for its link. And for its tooltip, let's use the following:

```
Edit this page of {currentTable} records related to this {mainTable}.
```

I'll review your PR further after you added comments and addressed my comments.

PoornimaVKrishnan:
Addressed all comments
@RFSH 
Issue still persisting:
During isMobile condition it is pushing all buttons to dropdown but buttonContainer also is having some width. I can alter the width or height or display of buttonContainer because resize sensor is on the parent element of buttonContainer and it wont get triggered if I change display/width/height of that. For now I have added display property but resize sensor is on the parent element of buttonContainer and it wont get triggered if I change width of buttonContainer.

src/components/record/related-table-actions.tsx
```diff
@@ -445,101 +531,331 @@ const RelatedTableActions = ({
...
+  return (
+    <>
+      <div className={containerClassName} ref={containerRef}>
+        {renderButtons()}
+        {buttonsInDropdown.length > 0 && (
+          <Dropdown onToggle={(isOpen: boolean, event: any) => toggleDropdown(isOpen, event)}>
+            <Dropdown.Toggle
+              disabled={false}
+              className={`chaise-btn chaise-btn-primary dropdown-toggle-table ${isMobile
+                && !relatedModel.isInline ? 'dropdown-toggle-table-mobile' : ''}`}
```
RFSH, line 819: Change `dropdown-toggle-table-mobile` to something else to be consistent with the new state variable name.

RFSH (changes requested):
Overall I like your solution. It's easy to understand and we don't need to adjust it if we decided to add more buttons or make any changes to the buttons. Apart from the comments that I made in the code, please apply the following changes. Your changes are a bit buggy in some cases.

1. **Honor the component template**: You should move your functions that you added around so it's consistent with the order that we want. We explained this in the `example.tsx`. So from top to bottom of a component function:

```tsx
const comp = () => {
  // 1. state variables
  
  // 2. useEffect and other hooks
  
  // 3. callbacks

  // 4. functions that are for rendering in the `return (...)` statement below

  return ( 
    ...
  );
}
```

So,
 - Move the `useLayoutEffect` to be after line 106.
 - Move the dropdown callbacks to be after the `useLayoutEffect` among other callbacks.
 - You can leave the `renderButton` and `renderButtons` where they are. 


2. **Make sure the whole row of dropdown items are clickable**: In the inline section, you have to click on the icon or text. But bootstrap shows the background change for the whole row. You most probably can fix this by just adding `width: 100%; text-align:center` to the buttons.

3. **Horizontal scrollbar of related entities shows up above the dropdown when you open the dropdown**: 
Screenshot 2023-07-10 at 11 24 13 AM

4. **Add this tooltip to the dropdown button**: `Click here to view the available actions.` for when we're only showing the dropdown and `Click here to view the extra available actions.` for when there are some buttons showing. In the Chaise meeting show these tooltips to Hongsuda and see if she agrees with these tooltips or not.

5. **The buttons go over the "Displaying ..." text in some cases**. I'm testing in [this page](https://dev.atlas-d2k.org/chaise/record/#2/Common:Collection/RID=17-E9KE) and width is around 900px.
Screenshot 2023-07-10 at 11 31 08 AM

Also the "loading..." that we show on the initial load (and when the page content needs to be updated) is showing up below the buttons.

PoornimaVKrishnan:
@RFSH 

**I have addressed all review comments**.

1) Make sure the whole row of dropdown items are clickable: In the inline section, you have to click on the icon or text. But bootstrap shows the background change for the whole row. You most probably can fix this by just adding width: 100%; text-align:center to the buttons --> If we give, `width: 100%; text-align:center` it is looking as below
Screenshot 2023-07-11 at 12 56 44 AM
So I changed to the following: `width: 100%; text-align:center; padding: 4px 18px`. It looks much nicer as below:
Screenshot 2023-07-11 at 12 57 46 AM
Let me know your thoughts

2) Also the "loading..." that we show on the initial load (and when the page content needs to be updated) is showing up below the buttons.
I didnt understand what the issue is, I compared with the deployed version and loading text looks the same.

PoornimaVKrishnan:
@RFSH 
--> Add this tooltip to the dropdown button: Click here to view the available actions. for when we're only showing the dropdown and Click here to view the extra available actions. for when there are some buttons showing. In the Chaise meeting show these tooltips to Hongsuda and see if she agrees with these tooltips or not

I discussed with Hongsuda and she is inclined to remove the wording 'Click here to' from the tooltip and just keep it as 'View the available actions.' and 'View the extra available actions.'

RFSH (changes requested):
Thank you for the changes. 

Regarding `text-align:center` in one of my comments, I meant `text-align:left` 😄  What you pushed looks nice.
 
While reviewing, I realized it's easier to push some changes instead of going back and forth. The following is the summary of what I pushed:

- Cleaned up some CSS rules, comments, and the component code a bit.
- We already had a `useRef` for the container called `container`. But when you added `containerRef`, you removed the old one. This broke some features. I just had to replace `container.current` with `contaienrRef.current` in multiple places. 
- Used a different class for the bulk edit button. This was breaking some test cases
- In some scenarios, we're not supposed to show the "custom mode" button. You accidentally removed this logic in your changes, which I had to bring back. This was also breaking e2e test cases. 
- Added test case for bulk edit

There's still a bug that you should look at and fix. When there are one or two buttons in the related section header, even if there is enough space, the buttons are collapsed under the dropdown. This is breaking our e2e test cases.

RFSH:
@PoornimaVKrishnan Let's change the `fa-solid fa-grip` that is used for the "Custom mode" button to `fa-solid fa-layer-group`. 

As part of #2340, we will use the `fa-grip` icon for the grab/grip button. So we should change the "Custom mode" icon to something else. I was going to push this directly, but since you've already made changes to that file, you should also make this change to reduce the number of conflicts.

P.S. We initially used the `fa-grip` because we couldn't find a better icon that didn't have a special meaning. Choosing an icon for this button is very difficult because the displayed value can be an iframe, list, or any other element markdown produces. We looked at a few alternatives and chose `fa-layer-group` because it doesn't imply anything special (like a graph, chart, or list). So it's safe to use it as a general-purpose icon. We ultimately want an icon that says "not table" and `fa-layer-group` looks good enough for our purpose. 
P.S.

PoornimaVKrishnan:
@RFSH 

Fixed the issue: 
When there are one or two buttons in the related section header, even if there is enough space, the buttons are collapsed under the dropdown. This is breaking our e2e test cases.

Also changed the custom mode icon

Please re-review this PR

RFSH (changes requested):
- The main section spinner `record-main-spinner-container` is showing up below the buttons. If you look at the rules for `.spinner-container` I'm intentionally not adding z-index for the spinner in this scenario because it didn't need it. But it seems like with your changes the spinner is displayed below the buttons. Most probably the solution is just to add z-index in this spinner as well. But before jumping to solution figure out why this 
behavior has changed (compare it with master).

- To test the race condition or edge cases I usually resize the page slowly to make sure the code is properly switching between adding the `>>` button and removing it. While doing this I found a scenario where we're showing all the buttons as well as `>>` with `Explore` button in it. To reproduce this:
  - Make the page width smaller so `>>` button shows up with `Explore` in it. 
  - Slowly make the page wider and see `Explore` button showing up beside the `>>` button.
  - The buttons stay in this mode until you make the page width smaller

src/assets/scss/_record.scss
```diff
@@ -118,6 +127,84 @@
...
+    * avoids the related action buttons to overlap it.
+    */
+    .entity-value .chaise-table-header {
+      display: inline-block;
+    }
+
+    .chaise-accordion-header {
+      .chaise-accordion-displayname {
+        white-space: nowrap;
+        max-width: 25%
```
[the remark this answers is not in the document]
PoornimaVKrishnan: agreed. removed

PoornimaVKrishnan:
> * The main section spinner `record-main-spinner-container` is showing up below the buttons. If you look at the rules for `.spinner-container` I'm intentionally not adding z-index for the spinner in this scenario because it didn't need it. But it seems like with your changes the spinner is displayed below the buttons. Most probably the solution is just to add z-index in this spinner as well. But before jumping to solution figure out why this
>   behavior has changed (compare it with master).
> * To test the race condition or edge cases I usually resize the page slowly to make sure the code is properly switching between adding the `>>` button and removing it. While doing this I found a scenario where we're showing all the buttons as well as `>>` with `Explore` button in it. To reproduce this:
>   
>   * Make the page width smaller so `>>` button shows up with `Explore` in it.
>   * Slowly make the page wider and see `Explore` button showing up beside the `>>` button.
>   * The buttons stay in this mode until you make the page width smaller

have fixed the spinner issue. The other issue of explore button showing in dropdown as well as outside is not reproducible. Please check it here : https://dev.gudmap.org/~povijaya/chaise/record/#2/Common:Collection/RID=14-4KG6

RFSH (changes requested):
Everything looks good. I just asked to add comments for the latest changes that you did.

P.S. I realized the test case that I added was broken, so I fixed it.

RFSH (approved):

PR #2329, opened 2023-07-07
React saved queries

jrchudy:
This pull request adds the saved query dropdown to recordset app. This dropdown has 2 options which allow for a query to be saved and a saved query to be applied.

Functionality and components added/changed:
 - recordedit component refactored to show in a modal popup when configured
   - the top panel for the modal has multiple differences described in a [comment](https://github.com/informatics-isi-edu/chaise/blob/6555ecb51af60507cd80cda28c16b2123d83ddd9/src/components/recordedit/recordedit.tsx#L486-L496) in the code
 - added `RecordeditDisplayMode` to control functionality and styles for modal popup
 - added `SavedQueryDropdown` component with menu options and functionality when those options are clicked
 - `populateCreateInitialValues` function changed to include parameter for initializing values in create mode that aren't set as "default" for the column and don't come from a prefill query param
 - `populateSubmissionRow` function changed to include a parameter for setting submission values that are not part of the form
 - `SparkMD5` dependency added for creating a query_id using their hash function
 - a `DuplicateSavedQuery` modal is added when a query that already exists is being saved again
 - test cases added to confirm defaults in form, saving a query, applying a query, and trying to save a duplicate query
   - since saved query functionality relies on setting the catalog number in the chaise config property, the catalog annotation is being updated before these tests are run
 - updated documentation to note changes to saved query configuration

To test this, there is a script in ErmrestDataUtils called `createFacetingSavedQuery.js` which will set up a catalog to test this feature. Make sure you add the chaise config property `savedQueryConfig` with your newly created catalog id. The table I set saved query to work on is `faceting:main` in that catalog.

#### Note
There is a bug after closing one of the saved query modals. The tooltip remains showing after the even though the dropdown is closed and the cursor isn't hovering over it anymore. This bug is present for the export dropdown too.


RFSH (commented):
I was testing the feature, and while everything works perfectly, the following are different:

- Regarding the "Show saved search criteria" popup:
  - In AngularJS, we're using a title in the format of `Saved search criteria for <table-displayname>`.
  - We're using `fa-regular fa-square-check` for the apply icon, but this branch is using `fa-solid fa-check`.
  - The "view details" button is not offered in the AngularJS version.
- When you apply a saved query, we're supposed to send a PUT request on load (to update the last used timestamp), which is missing from this branch.

I like how you changed the recordedit to have the modal implementation inside it (instead of doing something similar to `recordset-modal.tsx`). It makes the implementation easier to follow, and we don't need to add an extra submit logic similar to how I had to handle the "selected rows" in the `recordset-modal.tsx`. 

The downside is a slightly more complicated `recordedit.tsx` component. But this looks good, given that we only have one recordedit popup for now. 


P.S. You should remove [these comments](https://github.com/informatics-isi-edu/chaise/blob/master/src/services/config.ts#L435,L468) since you implemented them somewhere else.

RFSH (approved):

PR #2333, opened 2023-07-12
Allow input value to be copied to selected rows instead of all rows

PoornimaVKrishnan:
This PR is focussing on applying change to selected forms instead of all forms. 
Selected forms will have a border to show it is selected.

### DIFF
--- a/src/assets/scss/_input-switch.scss
+++ b/src/assets/scss/_input-switch.scss
@@ -22,6 +22,8 @@
   .chaise-input-control {
     height: auto;
     padding: 0 !important;
+    // Adding this to make the div clickable for apply some feature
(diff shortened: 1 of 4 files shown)


PoornimaVKrishnan:
@RFSH 

- Implemented Select All  and a button Apply and Clear to apply and discard the changes to selected forms. 
- Added overlay to the form to make it clickable
- Moved the action buttons to next line
-  Added indeterminate state for checkbox if not all forms are selected
- On load fist form is active showing it is clicked

RFSH:
I'm going to close this PR since we asked for more UI changes. You can create another PR later when we think the UI is finalized.

PR #2340, opened 2023-07-18
1408 support array column type in recordedit

ladukaraniket:
# Implemented array input fields 
image


src/components/input-switch/input-switch.tsx
```diff
@@ -252,8 +252,10 @@ const InputSwitch = ({
           displayDateTimeLabels={displayDateTimeLabels}
         />;
       case 'integer2':
+      case 'int4':
```
jrchudy, line 255: Instead of adding 2 new conditions, change the line above for the `ArrayField` definition:
`baseArrayType={columnModel?.column.type.baseType.name}`

To remove the `.name` so the component takes in an object instead. Make sure to refactor the array field component after making this change.

src/utils/input-utils.ts
```diff
@@ -317,9 +317,11 @@ export const VALIDATE_VALUE_BY_TYPE: {
 } = {
   'int': integerFieldValidation,
   'integer2': integer2FieldValidation,
+  'int4': integer4FieldValidation,
   'integer4': integer4FieldValidation,
   'integer8': integer8FieldValidation,
   'float': numericFieldValidation,
+  'float4': numericFieldValidation,
```
jrchudy, line 324: same comment about these types if we use the type object instead with getInputType

jrchudy (changes requested):
Functionally this looks good. I left some comments about cleaning up the way the `type` prop is being set for the `input-switch` component.

I will spend more time reviewing the functionality of the code after we make changes to align with the most recent comment I made in the issue:

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,275 @@
...
+  useEffect(() => {
+    let defaultValues = getValues(name);
 
-    // make sure it's an array
-    let validArray = false, arr;
-    try {
-      arr = JSON.parse(value);
-      validArray = Array.isArray(arr);
-    } catch (e) {}
+    defaultValues = typeof defaultValues === 'object' || !defaultValues ? defaultValues : JSON.parse(defaultValues)
```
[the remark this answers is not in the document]
ladukaraniket: In initial stage of implementation the variable defaultValues could take values of the following types - 
array parsed as a string, a array object, or an empty string value.
So to parse this correctly, the following logic was added.

In the current representation, the check for array object is not required and will be removed.

jrchudy (changes requested):
The functionality looks good still. The movement appears seamless. I can edit any array with different types without any issues that I've seen so far (dragging, adding/removing, set new values, using set all and changing all values). Defaults are showing properly and array fields are loading as I expect on edit.

Some UX issues I still found:
 - [ ] Padding on the right side of the `Add` button for array in `set-all` is too small (not the same as other places)
   - image
 - [ ] The date and timestamp buttons are colored in set all (they should not be)
   - image
   - image
 - [ ] Error for timestamp add row is showing under the buttons:
   - image


I found a few warnings/errors that show in the browser console that we should be fixed. If they can't be, or isn't a problem, that should be noted in the code and pull request:
 - [ ] on page load, `Function components cannot be given refs` error:
   > Warning: Function components cannot be given refs. Attempts to access this ref will fail. 
    Did you mean to use React.forwardRef()?
    > 
    > Check the render method of `ArrayField`.
        at InputSwitch (https://dev.isrd.isi.edu/~jchudy/chaise/bundles/common-25dc9f7d.0f2415cd2a8386ce08c1.js:3341:19)
 - [ ] uncontrolled inputs warning on click of "array" (numeric) in defaults schema
   > Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
    at input
 - [ ] duplicate key error for droppable
   > Warning: Each child in a list should have a unique "key" prop.
   >
   > Check the render method of `Droppable`. See https://reactjs.org/link/warning-keys for more information.
    at Droppable (https://dev.isrd.isi.edu/~jchudy/chaise/bundles/vendor-rest-d96105ec.9927d243abe72b48f53b.js:8482:69) 
 - [ ] dnd scroll error:
    > react-beautiful-dnd.esm.js:39 react-beautiful-dndDroppable: unsupported nested scroll container detected.A Droppable can only have one scroll parent (which can be itself)Nested scroll containers are currently not supported.We hope to support nested scroll containers soon: https://github.com/atlassian/react-beautiful-dnd/issues/131
    
As for other issues, we discussed file formatting changes happening when you are saving scss documents. Please try to avoid reformatting each css document when saving and making changes. Specifically `_input-swtich.scss` and `_recordedit.scss` have all of the "blank lines" removed which is probably fine but we don't care to "minify" our written files since we minify them during the build.

Please also run `make lint` before committing your changes and adjust any lint issues that are shown:
  > ✖ 9 problems (9 errors, 0 warnings)
  9 errors and 0 warnings potentially fixable with the `--fix` option.

src/assets/scss/_input-switch.scss
```diff
@@ -1,322 +1,282 @@
 @use 'variables';
```
RFSH, line 1: I don't think you've changed anything in this file, right? I only see code style and indentation change which you should revert. 

Apart from the fact that we should avoid pushing unnecessary changes, this is also changing indentation to 4 spaces which is wrong (we prefer 2).

src/assets/scss/_recordedit.scss
```diff
@@ -1,441 +1,373 @@
 @use 'sass:map';
```
RFSH, line 1: Same comment that I made about `_input-switch.scss` applies to this. Please avoid running formatters on the whole file and only run it on the lines that you've changed. Also make sure your formatter is honoring our style guide (you can use `.prettierrc.js` or `.eslintrc.js` that I've already added to the project).

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,326 @@
...
-const ArrayField = (props : ArrayFieldProps): JSX.Element => {
-  const textAreaRef = useRef<HTMLTextAreaElement>(null);
-  const arrayFieldValidation = (value: string) => {
-    if (!value) return;
+type RowItem = {
+  id: number,
+  value: string
+}
+
+type Options = {
```
RFSH, line 43: This is not used, so please remove it.

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,326 @@
...
+import { formatDatetime, formatFloat, formatInt, getInputType } from '@isrd-isi-edu/chaise/src/utils/input-utils';
+import React, { useEffect, useRef, useState } from 'react';
+import {
+  DragDropContext, Draggable, DraggableProvided, Droppable, DroppableProps, DroppableProvided, DropResult
+} from 'react-beautiful-dnd';
+import { EventType, useFormContext, useFormState, useWatch } from 'react-hook-form';
+
+// since we're using strict mode, react-beautiful-dnd misbehaves due to multiple renders caused by strict mode
+// this is to guard against it
+const StrictModeDroppable = React.memo(function StrictModeDroppable({ children, ...props }: DroppableProps) {
```
RFSH, line 18: Move this outside of this file. It should go to its own file under `src/components/chaise-droppable.tsx` and let's call the component `ChaiseDroppable`.

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,326 @@
...
-import { windowRef } from '@isrd-isi-edu/chaise/src/utils/window-ref';
-import { hasVerticalScrollbar } from '@isrd-isi-edu/chaise/src/utils/input-utils';
+import { formatDatetime, formatFloat, formatInt, getInputType } from '@isrd-isi-edu/chaise/src/utils/input-utils';
+import React, { useEffect, useRef, useState } from 'react';
+import {
+  DragDropContext, Draggable, DraggableProvided, Droppable, DroppableProps, DroppableProvided, DropResult
+} from 'react-beautiful-dnd';
+import { EventType, useFormContext, useFormState, useWatch } from 'react-hook-form';
+
+// since we're using strict mode, react-beautiful-dnd misbehaves due to multiple renders caused by strict mode
```
RFSH, line 16: If you're adding comments to functions, please use the jsdoc style of comments

```
/**
 *
 */
```

Also if you borrowed this code from somewhere else, you should include the link to that source.

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,326 @@
...
+  name?: string,
+  type?: EventType,
+  value?: unknown
+  values?: {
+    [x: string]: any
+  }
+}
+
+const ArrayField = (props: ArrayFieldProps): JSX.Element => {
+  const [itemList, setItemList] = useState<RowItem[]>([])
```
RFSH, line 53: Let's make these lines (53 to 60) easier to read,

- Move the deconstructing assignment of `props` to be first line.
- Move the `useFormContext` to be the second line.
- For each state variable definition, add a jsdoc style of comment explaining what each one is
```
/**
 * Explain what this is
 */
const [itemList, setItemList] = useState<RowItem[]>([])
```

RFSH (changes requested):

src/components/input-switch/date-time-field.tsx
```diff
@@ -80,11 +92,16 @@ const DateTimeField = (props: DateTimeFieldProps): JSX.Element => {
 
     // if only time is missing, use 00:00:00 for it
     let timeValTemp = '';
-    if (!timeVal && !props.requiredInput) {
+    if (!timeVal && !props.requiredInput && !props.requiredInput) {
```
RFSH, line 95: The condition already had a `requiredInput`. Remove the new one that is added.

src/components/input-switch/date-time-field.tsx
```diff
@@ -213,7 +233,7 @@ const DateTimeField = (props: DateTimeFieldProps): JSX.Element => {
               />
             </div>
           </div>
-          {!props.disableInput && props.displayExtraDateTimeButtons && <div className='chaise-btn-group'>
+          {!props.disableInput && props.displayExtraDateTimeButtons && <div className={`chaise-btn-group ${getFieldState(props.name)?.invalid ? 'translateY' : ''}`}>
```
RFSH, line 236: What's the point of this?

src/components/input-switch/date-time-field.tsx
```diff
@@ -205,6 +224,7 @@ const DateTimeField = (props: DateTimeFieldProps): JSX.Element => {
                 className={`${props.timeClasses} input-switch ${showTimeClear() ? 'time-input-show-clear' : ''}`}
                 type='text' disabled={props.disableInput} {...timeField} onChange={handleTimeChange}
                 placeholder={props.placeholder ? props.placeholder : dataFormats.placeholder.time}
+              // value={timeVal} //default time value does not reflect in the input element by setting the form state alone.
```
RFSH, line 227: Completely remove this instead of just commenting.

src/components/chaise-droppable.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { useState, useEffect } from "react";
```
RFSH, line 1: Use `'` for strings instead of `"`.

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,170 @@
...
-import { windowRef } from '@isrd-isi-edu/chaise/src/utils/window-ref';
-import { hasVerticalScrollbar } from '@isrd-isi-edu/chaise/src/utils/input-utils';
+import InputSwitch from '@isrd-isi-edu/chaise/src/components/input-switch/input-switch';
+import { formatDatetime, getInputType } from '@isrd-isi-edu/chaise/src/utils/input-utils';
+import React from 'react';
+import {
+  DragDropContext, Draggable, DraggableProvided, DroppableProvided, DropResult
+} from 'react-beautiful-dnd';
+import { useFieldArray, useFormContext, useFormState, useWatch } from 'react-hook-form';
+import { dataFormats } from '../../utils/constants';
```
RFSH, line 16: Don't use relative paths. This is a ESLint error.

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,170 @@
...
-  /* the type of each element in the array */
+  /**
+   * the type of each element in the array
+   */
   baseArrayType: string,
+  /**
+   * represents the presence of timezone in case of TimeStamp field
+   */
+  hasTimezone?: boolean
+  /**
```
RFSH, line 29: You've already defined this prop

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,170 @@
...
+  const addItem = (value: any) => {
 
-      if (!isValid) {
-        return '`' + val + '` is not a valid ' + getSimpleColumnType(props.baseArrayType) + ' value.';
-      }
+    let valueToAdd: any = {
+      "val": value
+    }
+
+    if (getInputType({ name: baseArrayType }) === `timestamp`) {
```
RFSH, line 58: ESLint error

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,170 @@
...
+      const DATE_TIME_FORMAT = props.hasTimezone ? dataFormats.datetime.return : dataFormats.timestamp;
+      const v = formatDatetime(value, { outputMomentFormat: DATE_TIME_FORMAT })
+
+      valueToAdd["val-date"] = v?.date
```
RFSH, line 62: ESLint error

RFSH (changes requested):
I'm not sure what I'm missing. I deployed this branch and I cannot get arrays to work. 
- If I leave them empty it's sending `[]` value which is wrong. It should just send `null`. 
- Array of text is sending a string instead of array.
- Other arrays are sending values like `{"val": "1234"}` which doesn't make any sense.

By the way, prior to pushing your changes always run `make lint` to make sure you don't have any linter errors. I assumed you've already installed `ESLint` VSCode extension. If you haven't, please do so you can see the linter errors right away.

src/components/input-switch/input-field.tsx
```diff
@@ -57,6 +57,19 @@ export type InputFieldProps = {
    * by default, we are capturing the "enter" key event and stopping it
    */
   allowEnter?: boolean
+  /**
+   * `optional`additional controller rules for the input field.
+   *  Check allowed rules here - https://react-hook-form.com/docs/useform/register#options
+   */
+   additionalControllerRules?: {
```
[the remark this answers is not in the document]
ladukaraniket: I had used this in the previous version of arrayField to pass additional props for Validation.
 I am no longer using this, but I left the overall flow of passing additional props in case there is a need for this in future for any other component.
 
 In case you don't see a possible use case where this might be useful, let me know. I'll remove it.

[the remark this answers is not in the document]
RFSH: Yeah, let's leave it then.

src/components/input-switch/input-field.tsx
```diff
@@ -57,6 +57,23 @@ export type InputFieldProps = {
    * by default, we are capturing the "enter" key event and stopping it
    */
   allowEnter?: boolean
+  /**
+   * true if the current Input field is an element of an InputField of type array
+   */
+  isArrayElement? : boolean
```
RFSH, line 63: Thanks for your adding this prop. But let's call it based on the effect that it does, not where you're using this. If you can come up with a better name than `displayRequireErrorBeforeSubmit` please feel free to change it. Otherwise use this one.

```ts
/**
 * display required error regardless of whether the form has been submitted or not.
 */ 
displayRequireErrorBeforeSubmit?: boolean
```

src/components/input-switch/array-field.tsx
```diff
@@ -1,108 +1,194 @@
...
-              ref={textAreaRef}
-            />
-            <ClearInputBtn
-              btnClassName={`${props.clearClasses} input-switch-clear ${
-                hasVerticalScrollbar(textAreaRef.current) ? 'has-scrollbar-clear' : ''
-              }`}
-              clickCallback={clearInput}
-              show={showClear && !props.disableInput}
+    <>
+      <div className={`array-input-field-container ${name} ${getInputType({ name: baseArrayType })}`}>
```
RFSH, line 140: Adding just the name or the type as class is bad. You should use a more specific name. Something like

```ts
const containerClassName = ['array-input-field-container'];
usedClassNames.push(`${containerClassName[0]}-${name}`);
usedClassNames.push(`${containerClassName[0]}-${getInputType({ name: baseArrayType })}`);

```

So then it would result in something like
```
array-input-field-container array-input-field-container-1-col array-input-field-container-text
```

src/assets/scss/_array-field.scss
```diff
@@ -0,0 +1,135 @@
...
+          padding-left: 3px;
+          & * {
+            position: relative;
+            top: 50%;
+            transform: translateY(-7px);
+          }
+        }
+        .input-box,
+        .input-switch {
+          border: 2px solid transparent;
```
RFSH, line 79: What's the point of this border and box-shadow? It's adding unnecessary padding and when the input is displayed you can see it:

Screenshot 2024-04-22 at 5 41 56 PM

This is not related to this rule but since I added the screenshot to this comment, I'm mentioning it here: Please remove the `height: 100%;` rule on `_recordedit.scss` line 128. With this rule the top border of each cell is not showing up properly when you activate the "select all/some" feature (see how the second form's height is smaller than it should be).

src/assets/scss/_array-field.scss
```diff
@@ -0,0 +1,135 @@
...
+  }
+  .array-input-field-container {
+    ul {
+      padding-left: 0;
+    }
+    .add-element-container {
+      display: flex;
+      flex-direction: row;
+      gap: 5px;
+      [class*='input-switch-container'] {
```
RFSH, line 31: What is this rule trying to do?

RFSH (changes requested):
Thank you for the recent changes. It works now and the array inputs are aligned properly inside the "select all/some" panel. It seems like you missed the empty input case. The code is still sending `[]` when you leave the inputs empty.

I pulled latest master and resolved the conflicts. This change updates a lot of different dependencies, so make sure to do `make deps-test` before deploying/testing. 

The last time I didn't pay too much attention to your tests, but now looking at it, I can't imagine that they would pass for you. The selectors that you used seem invalid to me. You're also directly using selectors. Please spend more time on testing. And this time around make sure they are passing and everything is ready.

src/assets/scss/_array-field.scss
ladukaraniket: It ensures that the input field takes all the available width. This is how it looks with and without this rule.
image
image

src/assets/scss/_array-field.scss
ladukaraniket: I removed those rules since they are no longer required. They were a part of a previous design.

Issue #1978, opened 2020-07-10
Display navbar app while catalog request is loading


RFSH:
In #1936, we explored different ways to improve the navbar performance. To summarize, for displaying the navbar, the app has to send the following category of requests:

1. Javascript and CSS files that `navbar.app.js` is relying on.
2. The catalog document.

With the latest changes, we are now allowing a method to eliminate the need for the first category of request (by prefetching them). We should investigate whether it's possible to load the navbar while the catalog document request is loading. This will completely eliminate the visible "load lag" of the navbar app. We might be able to show the navbar based on `chaise-config.js` property and later update it with the new information from the catalog document.

RFSH:
I spent some time figuring out the navbar slowness issue in our static sites. I ignored the case where we're dynamically fetching the assets. Since we aim to speed up the page, statically including navbar dependencies is a must.

## Showing static content inside navbar

When the navbar finishes loading, it will take up some space. That's why we suggest defining a height for the navbar to reduce the shifting on the page.

We could use a similar method to show a load or static content before the chaise's navbar loads. We can add anything we want inside the "navbar" tag to do so. When the navbar finishes loading, it will replace the content of the navbar. So instead of just `<navbar></navbar>` we could have:

```
<navbar>
  loader goes here
</navbar>
```
I'll document this in our navbar document.

## Chaise improvements

By default, the webpack plugin we're using is adding the script tags with the `defer` attribute. This means the browser will run the scripts only when the page has finished parsing. So even if `defaultCatalog` is missing, the navbar won't appear until all the static content (including images) is loaded.

We could force webpack not to add the `defer` attribute for navbar dependencies. By doing this, the page will wait for navbar dependencies before loading the rest of the page. And as soon as the dependencies are downloaded, it will execute the scripts. For this to work, we must ensure our React app is not blindly looking for `<navbar>` and only runs when the element finally loads. 

With these changes, we can now change the chaise logic to show some load indicators while waiting for the session and catalog request. We could also gradually show the navbar. I created a [`navbar-gradual`](https://github.com/informatics-isi-edu/chaise/tree/navbar-gradual) branch that includes the changes that I mentioned above.

P.S. While we're only focused on the catalog request, we're also fetching the session before showing the navbar. Besides the login/profile button, we need the session for rendering the navbar templates (as you could use `$session` in navbar templates). I'm not sure if skipping the session fetching makes sense, but at least we can show a spinner.

RFSH:
[...]

PR #2342, opened 2023-07-20
Improve Navbar UX in static sites

RFSH:
This PR will improve the Navbar UX in static sites by applying most of the changes I described [here](https://github.com/informatics-isi-edu/chaise/issues/1978#issuecomment-1613632197). The changes are:

- Ensure navbar dependencies are not using `defer` and, therefore, have priority over the static content. 
- Show a spinner while waiting for the session and catalog if the navbar has some preserved space.

### DIFF
--- a/src/services/config.ts
+++ b/src/services/config.ts
@@ -49,7 +49,9 @@ export interface ConfigServiceSettings {
   overrideImagePreviewBehavior?: boolean,
   overrideDownloadClickBehavior?: boolean,
   overrideExternalLinkBehavior?: boolean,
-  openIframeLinksInTab?: boolean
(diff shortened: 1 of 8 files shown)


src/services/config.ts
```diff
@@ -144,6 +144,7 @@ export class ConfigService {
       // the server object that can be used in other places
       ConfigService._server = ERMrest.ermrestFactory.getServer(service, ConfigService._contextHeaderParams);
 
+      console.log('asking for catalog!');
```
jrchudy, line 147: I'm guessing this is not needed

jrchudy (approved):

Issue #2335, opened 2023-07-14
Set all form row is set to view width instead of form container width


jrchudy:
When there are enough forms to cause the recordedit UI to add a scrollbar to go left and right, the set all row initially opens properly. Once you scroll left or right, the set all row scrolls with the forms and eventually moves out of view.

The content in the set all should stay in the same place in the viewport as the forms scroll left and right. Make sure the highlight is applied to every form as well.

<img width="1728" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/3da586f1-6eec-42ed-ae3a-636096383c9b">

Issue #2334, opened 2023-07-13
Markdown bar in Recordedit app wraps incorrectly


jrchudy:
When the user is creating/editing multiple records to the point that the form inputs are at the min-width, the markdown bar for the `longtext-field` should only wrap one time (show buttons across 2 rows max). We should consider setting the min-width to a value that prevents this from happening. 

The current behavior is different based on different zoom levels. Sometimes the content looks fine, then zooming in will force one of the sets of 3 buttons to move to the next line. Zooming in again sometimes fixes this..?!

This can also be seen when the `longtext-field` is disabled, sometimes going to 4 lines.

Issue #2320, opened 2023-06-07
Allow input value to be copied to selected rows instead of all rows


RFSH:
We can set the value for all the forms with the "set all" button in recordedit. In this issue, we want to allow users to set values for specific rows.

This is how it could be done:
- User clicks on the pencil button to open the "set all" UI.
- User selects the value the same way that they currently do.
- We should allow users to select the specific row to which they want the data to be applied. **This requires exploration**. Since we're disabling the inputs, we could allow users to click on each input to select the row that we want the data to be applied to.

**Code details**

[This](https://github.com/informatics-isi-edu/chaise/blob/master/src/components/recordedit/form-container.tsx#L360) is the callback that will set the value of all the inputs. Instead of doing this for all "forms", you would have to selectively do this for the forms that user has selected.

PoornimaVKrishnan:
Action points from the meeting:

1) Implement two Radio Buttons `Select All` and `Select None` and a button `Apply` to apply the changes. Select All will select all the forms and we can deselect from it if we don't want to apply the change to that particular form.  Select None is to deselect all the forms. 

Questions: By default should we have all selected or none selected?

2) Figure out a way to know the form is clickable

3) Experiment `Clear All` button to clear the text from forms

Question: Should we clear text from selected forms or from all forms?

PoornimaVKrishnan:
@RFSH
MOM from todays scrum call:

Move the i icon close to the checkbox label as we are explaining more on how the selection and deselection works
On resize of the textarea the controls on the top(i.e the buttons and checkbox) should also move
Tooltip for checkbox and label should not show once an interaction(i.e a click on checkbox) is made.
Also, in facets we are not making the checkbox label also clickable but in github it is clickable. We have to decide whether we should make it consistent with github or the chaise app
Where should the tooltip be aligned to in all cases(top, bottom, right, left)?
Pending items:

Updating the selection on removing form
Above items from the meeting
A bug which I found: On load one form is selected by default, one deselect of that checkbox is in indeterminate state instead of empty with Select All text
Also please check the tooltip text:

const renderHelpTooltip = () => {
    const splitLine1 = 'You can click on the form to select it and apply changes to it. '+ 
    'You can use the checkbox to select and deselect all records'
    const splitLine2 = 'By default, if there is no previous selection, the first form will be selected. '
    const splitLine4 = 'After the forms are selected, ' + 'you can click Apply button to apply the changes to selected records.'
[...]

PR #2345, opened 2023-08-01
2320 allow input value copied to selected rows

PoornimaVKrishnan:
This PR will modify the existing "select all" feature to essentially allow "select some" as we described in #2320 issue. Internally we decided to call this feature/UI element "multi-form-input". 

The following is the summary of changes,

- Since we're not doing "select all" anymore, we changed the name of the corresponding component from `SelectAllRow` to `MultiFormInputRow`. This change has been reflected in class names and other related properties.

- The existing `form-container.tsx` has three components in it. The file was very hard to read since we had to add more features to the `MultiFormRowInputRow`. So, this PR will break it into three different files. This way, each file has one component in it.

- Both `FormRow` and `MultiFormInputRow` need to keep track of "active forms". Also if we remove a visible form, we have to update this list. We could do this by adding it to the provider state, which would cause unnecessary renders. That's why we're passing some callbacks around for this purpose between the three components. We might want to clean up the code later, so I left a TODO for it.

- The multi-form-input is now sticky and takes only the available visible width. So it won't scroll with the rest of the forms (#2335). To do this, we had to add a `max-wdith` to the `multi-form-input-row` that is updated when the width of the `FormRow` changes (added to `form-row.tsx`).

- We didn't like how the input in the multi-form-row stretches and takes up the whole row. That's why we've imposed a maximum width of 1200px with CSS. This CSS change alone for the textarea case as the `textarea` element doesn't take up the full available width. That's why we had to add a `ResizeSensor` to `multi-form-input-row.tsx`. This solution can be a bit annoying as we're updating the width when the container width changes. So, if a user changed the width intentionally, they would have to change it again.

- https://github.com/informatics-isi-edu/chaise/issues/2334


PoornimaVKrishnan:
@RFSH 
Things pending: 

- the text area in the select-all-row should always be resizable
- cannot reduce the width of text area beyond a certain point (the textarea itself is getting smaller but the white background doesn’t). 

For the first point I tried removing the width 100% for `center-align` class, then its resizable.but the problem is it doesnt span across the entire width of the container.

PoornimaVKrishnan:
@RFSH 
I have fixed all pending bugs. I have added comments as well to know why we have done it in a certain way. Request your review on this

src/components/recordedit/form-container.tsx
```diff
@@ -216,74 +335,204 @@ const FormRow = ({ columnModelIndex }: FormRowProps): JSX.Element => {
...
+    }
+  }, [activeForm]);
+
+  // useEffect to call update only when there is a change in the width of select-all-row to update textarea width
+  useEffect(() => {
+    updateTextareaWidth();
+  }, [width]);
+
+  // useEffect to have a resize sensor to change the direction of upper-row when width of container is less than 400.
+  useEffect(() => {
```
RFSH, line 496: You're just explaining what this does and not why. You should mention you're doing this to make sure buttons are on top of the rest. You also need to explain why 400. Why not 350 or any other values? Is it calculated based on the min-width of textarea or something like that?

I also expected this to work for normal inputs, but it's only working in textarea case.

src/components/recordedit/form-container.tsx
```diff
@@ -216,74 +335,204 @@ const FormRow = ({ columnModelIndex }: FormRowProps): JSX.Element => {
...
   } = useRecordedit();
+  const {
+    watch,
+    reset,
+    getValues,
+    formState: { errors },
+  } = useFormContext();
+  let ref = useRef(null);
+  const [isEmpty, setIsEmpty] = useState(true);
+  const [selected, setSelected] = useState(false);
```
RFSH, line 464: What does this state variable mean? Please comment and also use a better name. Is this supposed to capture if all are selected or not?

src/components/recordedit/form-container.tsx
```diff
@@ -105,20 +158,35 @@ type FormRowProps = {
...
+    activeSelectAll,
+    canUpdateValues,
+    columnPermissionErrors,
+    foreignKeyData,
+    waitingForForeignKeyData,
+    getRecordeditLogStack,
+    getRecordeditLogAction,
   } = useRecordedit();
-
+  const [activeForm, setActiveForm] = useState<number[]>([]);
```
RFSH, line 181: This should be called `activeForms`

src/components/recordedit/form-container.tsx
```diff
@@ -136,22 +205,47 @@ const FormRow = ({ columnModelIndex }: FormRowProps): JSX.Element => {
...
 
     let cachedHeight = -1;
-    const sensor = new ResizeSensor(container.current as Element, (dimension) => {
-      const newHeight = container.current?.getBoundingClientRect().height;
-      if (newHeight === undefined || newHeight === cachedHeight || !container.current) return;
-      cachedHeight = newHeight;
-      const header = document.querySelector<HTMLElement>(`.entity-key.entity-key-${columnModelIndex}`);
-      if (header) {
-        header.style.height = `${cachedHeight}px`;
+    const sensor = new ResizeSensor(
```
RFSH, line 208: Change this block of code back to the style that it was. it was easier to read that way.

RFSH (changes requested):
Let's move `FormRow` and `SelectAllRow` to different files. It's hard to follow the code.I should've done that when I added select all functionality. I'll add more comments after moving the components.

Also please work on commenting your code and using better state variable names. It's hard to understand what your state variables are supposed to capture without reading the code. And when I say comment don't just comment what the code is doing. Add more details and explain why if it's not obvious.

Functional issues that I found:
- Because of the first comment that I added, users can select rows even if the select-all-row is invisible.
- The responsiveness of the `upper-row` should be the same in all input types (it only works on textarea)
- For some reason the z-index of dropdowns is broken now (dropdown shows up below the input).

src/components/recordedit/form-container.tsx
```diff
@@ -52,9 +51,42 @@ const FormContainer = (): JSX.Element => {
...
+    );
+
+    return () => {
+      mainResizeSensor.detach();
+    };
+  }, []);
+
+  /* Callback event for scroll functionality on recordedit-form to set a max-width to select-all-row as the
+  width of the visible area */
+  const handleScroll = () => {
```
RFSH, line 73: Why do you need to call this `onScroll` of the form? Shouldn't the resizeSensor be enough? If that's the case, you should just move the implementation of this function to be inside the resizeSensor.

src/components/recordedit/form-container.tsx
```diff
@@ -52,9 +51,42 @@ const FormContainer = (): JSX.Element => {
...
+
+    if (parentContainer && nonScrollableDiv) {
+      const visibleWidth = parentContainer.offsetWidth; // Width of the visible area
+      nonScrollableDiv.style.maxWidth = visibleWidth + 'px'; // Set the max-width to the visible width
+    }
+  };
+
+  /* This callback is called when we want to delete the form, we are setting the form index and
+  a boolean to know the remove button is clicked */
+  const deleteForm = (formIndex: number, formNumber: number) => {
```
RFSH, line 85: Let's call this `onRemoveForm` or `handleRemoveForm`

src/components/recordedit/form-row.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+// utils
+import { getDisabledInputValue } from '@isrd-isi-edu/chaise/src/utils/input-utils';
+import ResizeSensor from 'css-element-queries/src/ResizeSensor';
+import { isObjectAndKeyDefined } from '@isrd-isi-edu/chaise/src/utils/type-utils';
+import { makeSafeIdAttr } from '@isrd-isi-edu/chaise/src/utils/string-utils';
+
+type FormRowProps = {
+    columnModelIndex: number;
+    removeFormIndex?: number;
```
RFSH, line 20: Add jsdoc style of comments to all the props. I know we're not consistent but we should comment all our types. You can see an example [here](https://github.com/informatics-isi-edu/chaise/blob/master/src/components/tooltip.tsx#L6).

src/components/recordedit/form-row.tsx
```diff
@@ -0,0 +1,285 @@
...
+        activeSelectAll,
+        canUpdateValues,
+        columnPermissionErrors,
+        foreignKeyData,
+        waitingForForeignKeyData,
+        getRecordeditLogStack,
+        getRecordeditLogAction,
+    } = useRecordedit();
+
+    // This state variable is to set the form as active when its selected
```
RFSH, line 45: In your comment mention whether you're storing the `formIndex` or `formNumber`. I'm assuming it's `formNumber`

src/components/recordedit/form-row.tsx
```diff
@@ -0,0 +1,285 @@
...
+                    header.style.height = `${cachedHeight}px`;
+                }
+            }
+        );
+
+        return () => {
+            sensor.detach();
+        };
+    }, []);
+    /*
```
RFSH, line 90: Please use proper style of comments...

RFSH: You should also explain that we're doing it this way because the `removeClicked` comes from the parent and this is basically how we're communicating from parent to this child component.

src/components/recordedit/select-all-row.tsx
```diff
@@ -0,0 +1,405 @@
...
+            setSelectAll(activeForms?.length === forms.length);
+        }
+    }, [activeForms]);
+
+    // useEffect to call update only when there is a change in the width of select-all-row to update textarea width
+    useEffect(() => {
+        updateTextareaWidth();
+    }, [selectAllWidthChanged]);
+
+    /* useEffect to have a resize sensor to change the direction of upper-row when width of container is less than 400.
```
RFSH, line 85: Please use the proper jsdoc style of comment.

src/components/recordedit/select-all-row.tsx
```diff
@@ -0,0 +1,405 @@
...
+                }
+            }
+        );
+
+        return () => {
+            mainResizeSensor.detach();
+        };
+    }, []);
+
+    // useEffect to have a resize sensor to width of textarea to the the parent container width.
```
RFSH, line 111: Your comment doesn't make sense to me. Can you please rewrite this and explain it better? Like I mentioned you should explain but what this function is doing, and why you're doing this.

src/components/recordedit/select-all-row.tsx
```diff
@@ -0,0 +1,405 @@
...
+                <p>{splitLine2}</p>
+                <p>{splitLine3}</p>
+                <p>{splitLine4}</p>
+            </>
+        );
+    };
+
+    return (
+        <div className='select-all-row match-entity-value'>
+            <div className={`centre-align ${columnModel.inputType === 'longtext'
```
RFSH, line 281: You've done the similar check for inputType in multiple places. Let's just do this once and store it in a variable and then use it in these different places. You can call it `isTextarea` or something like that.

src/components/recordedit/select-all-row.tsx
```diff
@@ -0,0 +1,405 @@
...
+                    foreignKeyData.current,
+                    formValue,
+                    SELECT_ALL_INPUT_FORM_VALUE,
+                    clearValue
+                )
+            );
+        });
+    };
+
+    /* This is to set the width of text area as the width of select-all-row. We have to involve javascript as
```
RFSH, line 233: Use jsdoc style of comment. You should also explain why 1800 and 1200. And also have a note that if we want to change this, we should also change the css rules that you added (you should point to the rule that is using the same numbers).

RFSH (changes requested):
Apart from the comments that I made, we should most probably not use "select all" anymore to refer to this feature and component. I don't have any good suggestion and need to think more about it. The only thing that comes to my mind is `multi-form-input-row`.

I mentioned this in my comments multiple times, please make sure you're using this style of comments when you have multiple lines of comment (and for all function comments):
```tsx
/**
 *
 */
```
It's standard, easier to  and modify.

src/components/recordedit/form-container.tsx
PoornimaVKrishnan: we have to set max-width when we scroll horizontally across the container.
Its a common function getting called on, onScroll and resizeSensor

src/assets/scss/_recordedit.scss
```diff
@@ -104,28 +104,151 @@
...
+            justify-content: center;
+            align-items: center;
+            gap: 5px;
+            max-width: 100%;
+          }
 
-          .select-all-text, .select-all-input, .select-all-buttons {
-            float: left;
+          .centre-align-textarea {
+            width: auto;
```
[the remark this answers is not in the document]
PoornimaVKrishnan: This is not a child class for centre-align, it is a separate class for textarea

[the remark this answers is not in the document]
RFSH: I know. `.centre-align-textarea` is a more specific class that you're adding to only some `.centre-align` elements. That's why I wrote it as `&.centre-align-textarea`.

In scss just because you're writing a rule inside another one doesn't mean that it will always go another level deep. It just means start from there. `&` in here refers to `.centre-align` elements and by appending `.centre-align-textarea` to it, we're basically saying all the `.centre-align.centre-align-textarea` elements.

If I you were to just say `.centre-align-textarea` or `& .centre-align-textarea` then it would be different and not what we want. 

Also let's use `center` instead of `centre`.

src/components/recordedit/form-container.tsx
RFSH: Ok so let's change its name to be what it actually does instead of just `handleScroll`.

src/assets/scss/maps/_z-index-map.scss
```diff
@@ -62,4 +62,6 @@ $z-index-map: (
   'modal-error-backdrop-on-search-popup': 1061,
   'modal-error-on-search-popup': 1066,
 
+  // higher than dropdown
+  'multi-form-input-row': 2
```
RFSH, line 66: Please move it up there. The values in this map are sorted for readability purposes.

src/assets/scss/maps/_color-map.scss
```diff
@@ -36,4 +36,5 @@ $color-map: (
   'vocab-background': rgba(220, 220, 220, 0.68),
   'warning': #f0ad4e,
   'white': #ffffff,
+  'how-to-multi-form': #000
```
RFSH, line 39: The values here are ordered alphabetically based on their key, so please put it in the right place.

RFSH (changes requested):
It seems like you only changed `select-all` to `multi-form-input` but that's not enough. You need to make sure any mention of select all (selectAll, etc) in the code is changed.

The instances that I could think of:
- the SelectAllRow component name
- Variables with selectAll name in them.

src/components/recordedit/multi-form-input.tsx
```diff
@@ -0,0 +1,424 @@
...
+            }
+        }
+    };
+    // -------------------------- render logic ---------------------- //
+
+    const columnModel = columnModels[columnModelIndex];
+    const colName = columnModel.column.name;
+    const inputName = `${SELECT_ALL_INPUT_FORM_VALUE}-${colName}`;
+
+    const btnClass = `${makeSafeIdAttr(
```
RFSH, line 272: Revert this back to be just one line. The formatter that you'be used have broken this into three for no reason.

RFSH:
Now that all the changes in this PR are done, we should write test cases for this feature.

## Available documents

The following are our test-related docs. You don't need to read all of these initially, but I wanted to have them in one place so you can easily find them and return them. In this case, I already created the skeleton, so you don't need to worry about many details, and you can jump into writing the test cases.
- I recommend following our step-by-step guide on how to run test cases that is [here](https://github.com/informatics-isi-edu/chaise/wiki/How-to-install-Chaise-ERMrestJS-from-you-local-machine#running-ermrestjs-and-chaise-tests)
- [This document](https://github.com/informatics-isi-edu/chaise/blob/master/docs/dev-docs/e2e-test.md) is our main doc for e2e test cases. The following are the sections that might be useful for you:
  - [File structure](https://github.com/informatics-isi-edu/chaise/blob/master/docs/dev-docs/e2e-test.md#file-structure): You can see how the whole "e2e" folder looks like at a glance.
  - [Debugging](https://github.com/informatics-isi-edu/chaise/blob/master/docs/dev-docs/e2e-test.md#debugging): Goes over how you can use the Chrome inspector client for easier debugging.
- In [here](https://github.com/informatics-isi-edu/chaise/blob/master/docs/dev-docs/e2e-test-writing.md), we go over different topics related to how to write test cases.

## Files of interest

As I mentioned, I already pushed a skeleton for this. The following are the files that you should know about :

- [multi-form-input.spec.js](https://github.com/informatics-isi-edu/chaise/blob/2320-allow-input-value-copied-to-selected-rows/test/e2e/specs/default-config/recordedit/multi-form-input.spec.js): The test spec. 
- [multi-from-input.conf.js](https://github.com/informatics-isi-edu/chaise/blob/2320-allow-input-value-copied-to-selected-rows/test/e2e/specs/default-config/recordedit/multi-form-input.conf.js): This file allows us to run the individual spec. You don't need to change this file. To run this spec, you need to pass the location of this file to protractor. So, assuming you're running this command from the chaise root folder:`npx protractor test/e2e/specs/default-config/recordedit/multi-form-input.conf.js`
- [data_setup/schema/recordedit/multi-form-input.json](https://github.com/informatics-isi-edu/chaise/blob/2320-allow-input-value-copied-to-selected-rows/test/e2e/data_setup/schema/recordedit/multi-form-input.json): In this file you need to add the definition for the table that you're going to use in the spec. Don't worry about this for now. We can finalize this together once we have the list of test cases.

## How to proceed

I recommend following these steps:

1. Follow the instructions [here](https://github.com/informatics-isi-edu/chaise/wiki/How-to-install-Chaise-ERMrestJS-from-you-local-machine#running-ermrestjs-and-chaise-tests) on how to setup the env variables and run chaise test cases.
    - As I mentioned in our call, you need to create a secondary dummy user. When you've done that, share the cookies of your users with me so I can extend their expire date (so you don't have to keep changing them).
    - Instead of `make test` just run `npx protractor test/e2e/specs/default-config/recordedit/multi-form-input.conf.js`
    - Skip running ermrestjs test cases.

2. Go over the [multi-form-input.spec.js](https://github.com/informatics-isi-edu/chaise/blob/2320-allow-input-value-copied-to-selected-rows/test/e2e/specs/default-config/recordedit/multi-form-input.spec.js) file and modify this with `describe` and `it` blocks that describe all the test cases that you will eventually implement. When you're done with this, push it and let me know so we can finalize it together.
  - Feel free to change any of the existing ones that I've added inside the `create mode`. I think the `edit mode` is done and you don't need to modify that. We should mostly test in "create mode" and the "edit mode" is just to make sure the functionality works and we don't need to go over every scenario.

4. After finalizing the skeleton, you can start implementing test cases. I recommend running test cases frequently so you can find issues faster instead of implementing the whole thing and then running it.
    - Let me know if you need help implementing a test case. I can help you, and we most probably have done similar tests in other places that I can point them to you. We also have defined some helper functions in the [`utils` folder](https://github.com/informatics-isi-edu/chaise/tree/master/test/e2e/utils), which might be useful to you. Namely `record-helpers.js`, `recordedit-helpers.js`, and `recordset-helpers.js`.
    - Avoid using selectors directly in your test cases and instead add them to [chaise.page.js](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/utils/chaise.page.js). You should also take a look at the existing selectors as we might already have a getter for it in this file.

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,732 @@
...
+      });
+
+      it('it should not be offered in single mode.', () => {
+        let toggleBtn = recordEditPage.getColumnMultiFormButton('markdown_col');
+        expect(toggleBtn.isPresent()).toBeFalsy("toggle btn is present");
+      });
+
+      it('it should be displayed as soon as users added a new form.', (done) => {
+        chaisePage.clickButton(cloneFormSubmitButton).then(() => {
+          return recordEditPage.getColumnMultiFormButton('markdown_col')
```
RFSH, line 113: you don't need to create a promise here. just capture it in a variable and do the `isPresent`.. similar to the previous it.

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,732 @@
...
+        expect(input.isEnabled()).toBeFalsy("col " + 'id' + " was not disabled.");
+        let toggleBtn = recordEditPage.getColumnMultiFormButton('id');
+        expect(toggleBtn.isPresent()).toBeFalsy("toggle btn is present");
+      });
+
+
+      it('by default only the first form should be selected.', (done) => {
+        let toggleBtn = recordEditPage.getColumnMultiFormButton('markdown_col');
+        chaisePage.clickButton(toggleBtn)
+          .then(() => recordEditPage.getInputSwitchContainer('markdown_col', 1))
```
RFSH, line 131: You don't need these two promises. similar to the next `it`

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,732 @@
...
+      });
+
+
+      it('the form should work as a toggle when selected', (done) => {
+        const inputSwitch = recordEditPage.getInputSwitchContainer('markdown_col', 3)
+        const parentElement = recordEditPage.getParentElement(inputSwitch);
+        chaisePage.clickButton(parentElement).then(() => {
+          expect(parentElement.getAttribute("class")).toContain('entity-active', 'Form is not selected');
+        }).then(() => {
+          return chaisePage.clickButton(parentElement).then(() => {
```
RFSH, line 158: Chain this promise with the previous ones instead of creating a new promise chain.

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,732 @@
...
+        })
+      });
+      it('previous selection should remain after closing and opening again', (done) => {
+        const inputSwitch = recordEditPage.getInputSwitchContainer('markdown_col', 3)
+        const parentElement = recordEditPage.getParentElement(inputSwitch);
+        chaisePage.clickButton(parentElement).then(() => {
+          expect(parentElement.getAttribute("class")).toContain('entity-active', 'Form is not selected');
+        }).then(() => {
+          let toggleBtn = recordEditPage.getColumnMultiFormButton('markdown_col');
+          chaisePage.clickButton(toggleBtn)
```
RFSH, line 171: Chain this promise with the previous ones instead of creating a new promise chain.

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,732 @@
...
+            })
+        }).catch(chaisePage.catchTestError(done));
+      })
+
+      it('the form should not be clickable', (done) => {
+        let toggleBtn = recordEditPage.getColumnMultiFormButton('markdown_col');
+        const inputSwitch = recordEditPage.getInputSwitchContainer('markdown_col', 1)
+        const parentElement = recordEditPage.getParentElement(inputSwitch);
+        chaisePage.clickButton(toggleBtn).then(() => {
+          return chaisePage.clickButton(parentElement).then(() => {
```
RFSH, line 185: Chain this promise with the previous ones instead of creating a new promise chain.

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,732 @@
...
+          })
+        });
+      });
+      describe('checkbox functionality', () => {
+        it('on load the label should reflect what is selected.', (done) => {
+          chaisePage.clickButton(recordEditPage.getColumnMultiFormButton('markdown_col'))
+            .then(() => {
+              return recordEditPage.getInputSwitchContainer('markdown_col', 1);
+            })
+            .then((inputSwitch) => recordEditPage.getParentElement(inputSwitch))
```
RFSH, line 197: Unnecessary promises

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,732 @@
...
+            .then((parentElement) => {
+              expect(parentElement.getAttribute("class")).toContain('entity-active', 'Form is not selected');
+            })
+            .then(() => recordEditPage.getAllElementsWithClass('.form-header.entity-value'))
+            .then((count) => {
+              return recordEditPage.getCheckboxLabel().getText()
```
RFSH, line 203: Chain the promise

RFSH (changes requested):
I assumed you already cleaned up the unnecessary promise chains and usage of promise chain instead of another. But I see a lot of this. I won't review this any further until you fixed all of these smaller changes that you already know.

test/e2e/specs/default-config/recordedit/multi-form-input.spec.js
```diff
@@ -0,0 +1,739 @@
...
+const recordEditHelpers = require('../../../utils/recordedit-helpers.js');
+
+
+const testParams = {
+  schema_table: 'multi-form-input:main',
+  max_input_rows: 200,
+  apply_tests: {
+    not_ci: true,
+    types: [
+      {
```
RFSH, line 12: So the following is how we're doing the apply tests:

1. Open the recordedit page and clone more forms.
2. Select all the forms and apply a value to all of them.
3. Deselect some forms and see that we can apply to some.
4. Deselect more forms and see that we can clear some.
5. Manually change some values in some forms.

Therefore,
- Instead of hardcoding that you're cloning 4 more forms. Just add a variable called `number_of_forms` where it's value would be "5".
- `apply_to_all_value` makes sense.
- For the "apply to some" you need two variables. One is the value and the other one would be forms that you are deselecting.
- For the "clear some" you need the forms that you're deselecting.

So the whole thing would be

```
apply_tests: {
  number_of_forms: number,
  types: [
    {
      type: string,
      column_name: string,
      apply_to_all_value: string,
      apply_to_some: {
       value: string,
       deselected_forms: number[]
      },
      clear_some_deselected_forms: number[]
      manual_test: {
        value: string,
        form: number
      }
    }
  ],
  submission: ...
}
```

We have 5 forms, 
1. apply all will change all. 
2. with apply some deselect 2, which means changing 3 forms.
3. for clear some deselect another 2, which means clearing value for 1 form.
4. Change the value of one of the forms.

It's ok if you apply some/clear some the same forms. But for the manual value change please use different forms for each type/column.

RFSH (changes requested):
Thank you for refactoring the spec. I think overall you've done a good job of testing the overall features. I left some comments about how we should avoid using protractor for figuring out the expected behavior. The expected behavior should always be static and predetermined. 

And as for the apply tests, I left a comment about how you should add more variables. Overall it looks good, but you shouldn't assume which forms are used for apply some (and clear some).

RFSH (approved):

RFSH:
@jrchudy This PR is ready to be merged. I updated the description to reflect the latest PR state and have more information. 

Changing the `selectFile` function to return a promise solved the `browser.sleep` issue I mentioned. I also fixed the bug that we saw related to max-width of multi-form-input-row. I just had to move the `ResizeSensor` to the proper place.

PR #2347, opened 2023-08-16
Support chaise tooltips

RFSH:
This RP will modify the `DisplayValue` component to handle chaise tooltips. Since we have to do this in JavaScript and outside of React, I'm [directly using bootstrap's tooltips](https://getbootstrap.com/docs/5.3/components/tooltips/#enable-tooltips) for this.

### DIFF
--- a/src/components/display-value.tsx
+++ b/src/components/display-value.tsx
@@ -9,14 +9,29 @@ import { DEFAULT_DISPLAYNAME } from '@isrd-isi-edu/chaise/src/utils/constants';
 import { createChaiseTooltips } from '@isrd-isi-edu/chaise/src/utils/ui-utils';
 
 type DisplayValueProps = {
+  /**
+   * The value that should be displayed.
+   */
   value?: Displayname,
+  /**
+   * whether the component should return a special value for null and empty
+   */
   specialNullEmpty?: boolean,
+  /**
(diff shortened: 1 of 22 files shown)


src/components/display-value.tsx
```diff
@@ -22,6 +23,39 @@ const DisplayValue = ({
...
+          el.classList.add('chaise-icon-for-tooltip');
+        }
+        new Tooltip(el, {
+          title,
+          // @ts-ignore ts doesn't understand that we're actually sanitizing the value.
+          placement: ['auto', 'top', 'bottom', 'left', 'right'].indexOf(placement) !== -1 ? placement : 'bottom'
+        })
+      });
+    }
+  });
```
jrchudy, line 57: Should the dependency array for this be `undefined`? My understanding is that an `undefined` dependency array will fire useEffect on each render but we only need to run this once when the `spanRef` is attached to the page

jrchudy (approved):
The changes look fine for consuming the properties. We should consider using a constant for `data-chaise-tooltip`, `data-chaise-tooltip-placement`, and `data-chaise-tooltip-no-icon` in case this is used in other places or changes.

I also left a comment about the dependency array for the `useEffect` that we should consider so the bootstrapJS tooltip functionality isn't reapplied on every rerender.

PR #2348, opened 2023-08-22
Foreign key dropdown

jrchudy:
This PR adds UI support for showing foreign key inputs in recordedit as typeahead dropdown elements. These changes go alongside [changes in ermrestJS](https://github.com/informatics-isi-edu/ermrestjs/pull/986) that add a new property to multiple annotations for configuring this display mode.

Changes included:
 - Added functionality for foreign key dropdown with search and load next page functionality
 - when foreign key dropdown and boolean dropdown are opened, attach a class to the `recordedit-form` to signal a dropdown is open. This class adds `padding-bottom: 400px` to extend the height of the div to allow for content to overflow down. These 2 dropdowns have a max-height of `395 px`. 
 - Changed a constant value that was spelled incorrectly
 - test cases in default schema for setting default properly
 - test cases for functionality of dropdown
 - include chromedriver in package.json to fix local test environments
 - added logging for each action associated with the dropdown
   - log actions added in [Action List](https://docs.google.com/spreadsheets/d/1cHhPR0AacvuH2o3QavWlJCIIyPajfn93fJzFnc7VA5M/edit#gid=1180557009&range=228:236)
   
To test this, I added a foreign key dropdown input to the defaults schema in `ErmrestDataUtils`.


src/components/input-switch/foreignkey-dropdown-field.tsx
```diff
@@ -0,0 +1,472 @@
...
+    searchRef.read(pageLimit, logObj).then((page: any) => {
+      setCurrentDropdownPage(page);
+
+      if (page.length === 0) {
+        const fakeTuple = {
+          rowName: {
+            value: '<i>No Results</i>',
+            isHTML: true
+          },
+          uniqueId: 'no_results_row'
```
RFSH, line 277: Instead of this, let's just add a special flag to this tuple, similar to how we did this in faceting. Something like `isNoResultRow: true` or a better name

RFSH: Or simply set the `dropdownRows` to empty array and in `renderDropdownOptions` render the special "no results" row if this array is empty. You already have a special handling for it in that function, there's no reason to do something special here.

RFSH (changes requested):
## Functionality

After reverting the submit `preventDefault` change, everything works properly for me. The new test cases pass locally. That being said, we need to "invalidate" the currently fetched rows if users made any changes to the form and the current input has domain filter. The URL to fetch the rows might change if users made a change to the form and `dropdownReference` is still based on the initial data. To test this, we could change on of the fks in the domain-filter spec to be a dropdown. 

## UI

The only UI problems that I could find are related to the spinner:

1. The foreignkeys in the "select-all-row" are not showing the spinner properly. Moving the `.column-cell-spinner-container` rules under `.entity-value` to be under the next line `.entity-value,.match-entity-value` will help a bit. Apart from that we should add a `position: relative` to the input and also adjust the top position.
2. The spinner is not showing up during search. It's only displayed for initial load and fetching the next page. 


## Code

Apart from the comments that I made inline, I think we should avoid duplicating code. I understand why instead of just adding branching logic to existing `ForeignkeyField` you added a new `ForeignkeyDropdownField`, but there are a bunch of duplicated code that we should refactor.

I can think of two ways to tackle this:
 
1. We could just move the three common functions to `recordedit-utils.ts`. Similar to how `populateLinkedData` is moved there.
4. Change existing `foreignkey-field.tsx` to only have the common functionalities and add a new foreignkey-popup-field.tsx`. Similar to how we refactored `input-field.tsx`.

src/assets/scss/_input-switch.scss
```diff
@@ -171,16 +171,114 @@
...
+
+      .chaise-input-group-append > .chaise-search-btn {
+        border-bottom-right-radius: 0;
+      } 
+    }
+
+    // for making the dropdown list scrollable (keeping the buttons sticky)
+    .dropdown-list {
+      // rules specifically for fk-dropdown list of options
+      max-height: 355px;
```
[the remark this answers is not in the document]
jrchudy: This was to make sure the dropdown doesn't get "too long". I want the overall height of the dropdown to be a maximum of `400 px`. This is specifically for the list of options that doesn't include the search box and any padding added above/below the `dropdown-list`.

Otherwise, there is no specific reason.

src/assets/scss/_recordedit.scss
```diff
@@ -86,6 +86,13 @@
       min-width: 250px;
       overflow-x: auto;
 
+      &.dropdown-open {
+        // max height of a dropdown is `395px`
```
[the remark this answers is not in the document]
jrchudy: Refer to the comment in `_input-switch.scss`. The max height of the `.dropdown-menu` is `399px`. The `355px` height is referring to the `.dropdown-list` element only.

src/components/input-switch/foreignkey-dropdown-field.tsx
jrchudy: I think at some point I had a function that was rendering each row and I didn't do anything special for the no results row. That eventually changed to having a conditional that I didn't think about cleaning up.

RFSH (approved):

Issue #2308, opened 2023-05-01
Integrate RUI with Chaise


RFSH:
# Summary
We need to explore how we can integrate Chaise with CCF-RUI.

The main integration is on the recordedit app, where we want to allow users to access the RUI application and get a JSON that we store. I created this issue to summarize our discussions about this.

Useful links:

- GitHub Repo: https://github.com/hubmapconsortium/ccf-ui
- RUI Standalone Version: https://hubmapconsortium.github.io/ccf-ui/rui/
- Instructions for embedding the RUI in a larger application: https://github.com/hubmapconsortium/ccf-ui/blob/main/EMBEDDING.md#embedding-the-ccf-rui

# Details

- As I mentioned, we would like to use this app as part of recordedit. We could implement this the same way that we have the foreign key pickers. Users will see an input that summarizes the selected row, and upon clicking, we will open a modal.

- In the modal, we could show the RUI app in an iframe for simplicity. This way, we don't need to worry about making it work with React, and we just have to ensure that Chaise and RUI can communicate. We also don't need to increase the Chaise bundles for something used in only one deployment.

- RUI allows us to define a callback for getting the submitted registry. We can also provide a `editRegistration` for the edit case. So we don't need any extra changes in the CCF-RUI app and can just use one of their releases. We can download their Github release and deploy that code in ATLAS-D2K. So we don't need to worry about CDN or building the bundles ourselves.

- After RUI sends the JSON to Chaise, we would close the modal and change the displayed value in the input. **What exactly should we show?** We could use the JSON to create a summary and show that.

- Since we don't control the content of JSON and are unsure of how big it could get, we should store it in hatrac. The iframe code will be responsible for fetching the hatrac file in edit mode and providing it to CCF-RUI, and the Chaise code will be responsible for saving the file.

Based on this, I created the [`rui-field`](https://github.com/informatics-isi-edu/chaise/tree/rui-field) branch in Chaise as a proof of concept. While I hard-coded the iframe location in the code, we should develop a more general solution. We could add a new `iframe-input` annotation which would allow us to define the location of the iframe and the type of stored value. This is a very rough draft, but something like this:

```json
{
  "tag:isrd.isi.edu,2023:iframe-input": {
    "url_pattern": "/ccf-rui/index.html",
    "type": "asset"
  }
}
```

The `type` would allow us to say that the iframe will return a file or `json`, or in some other future use case, an `int`. 

_this description is subject to change and is not finalized_

[...]

PR #2351, opened 2023-08-24
Support iframe inputs

RFSH:
This PR will implement the chaise side of changes for the new `input_iframe` property that we discussed [here](https://github.com/informatics-isi-edu/chaise/issues/2308#issuecomment-1670121576). 

Summary of changes:

- Added a new `IframeField` component to `InputSwitch`.
- Added a new `IframeFieldPopup` component for showing the iframe in a modal.
- Changed the following different places in recordedit to handle the column mappings:
  - `UploadProgressModal` and `RecordeditProvider.areFilesValid`: the asset column might be part of another columns `inputIframeProps.columns` (and not visible).
  - `recordedit-utils.ts` functions:
    - `copyOrClearValue`: it should also change the extra columns in the column mapping (used in select-all feature)
    - `populateEditInitialValues`: it will populate values for the columns in the column mapping.
    - `populateSubmissionRow`: it will populate values for the columns in the column mapping.
    - Added a new `populateLinkedData` since it's now used in two places.
- Added test cases for a simple iframe case. This test cases use the `input-iframe-test.html` file that I added to `test/e2e/utils` (the first step before testing is copying this to remote location. similar to how we're doing this for `chaise-config.js` file).

ermrestjs PR: https://github.com/informatics-isi-edu/ermrestjs/pull/988


src/providers/recordedit.tsx
```diff
@@ -575,7 +575,8 @@ export default function RecordeditProvider({
 
           // NOTE: This check was being done in angularJS since the modal was closed with a message if the user aborted uploading
           //   - in ReactJS it's handling the case when the modal is closed and no "exception" is returned (undefined)
-          if (typeof exception !== 'string') {
+          //   - when we abort the upload, we're calling onCancel without any parameters
+          if (isObjectAndNotNull(exception)) {
```
jrchudy, line 579: This is better than what I left this case as. I wasn't sure how to handle this before, it makes more sense that we want to check for object and not null since now we know a string will never be returned.

src/utils/recordedit-utils.ts
```diff
@@ -38,16 +38,23 @@ export function columnToColumnModel(column: any, queryParams?: any): RecordeditC
...
   const logStackPathChild = column.isForeignKey ? LogStackPaths.FOREIGN_KEY : LogStackPaths.COLUMN;
 
+  let isRequired = !column.nullok;
+
   let type;
-  if (column.isAsset) {
+  if (column.isInputIframe) {
+    type = 'iframe';
+    // as long as any of the columns in the mapping are required, it should say its required
+    // isRequired
```
jrchudy, line 47: I think you meant to update `isRequired` here by checking if any of the columns in the mapping are required.

src/utils/recordedit-utils.ts
```diff
@@ -507,9 +567,10 @@ export function populateEditInitialValues(
  */
 export function populateSubmissionRow(reference: any, formNumber: number, formData: any, initialValues?: any[]) {
   const submissionRow: any = {};
-  const setSubmission = (col: any, skipEmpty?: boolean) => {
+  const setSubmission = (col: any, skipEmpty?: boolean, includeDisabled?: boolean) => {
     let v = formData[formNumber + '-' + col.name];
 
+    // TODO col.isDisabled is wrong. it's always returning false
```
jrchudy, line 573: Is this still an issue?

jrchudy (approved):
Changes are good for this. I will run the e2e tests once the ermrestJS changes are updated and the comment I left is here is addressed.

I left a comment about `isRequired` in `recordedit-utils` in `columnToColumnModel` since I wasn't sure if you meant to implement something for columns that need to be set for the input iframe to be valid.

src/utils/recordedit-utils.ts
RFSH: Yeah, `col` in this case is an ermrestjs `ReferenceColumn` object and `isDisabled` is not available on this object. The `populateSubmissionRow` function in `recordCreate.js`is also doing the similar check. Instead of just removing `col.isDisabled` I left this TODO so I can think about it and change it later outside of this PR.

PR #2353, opened 2023-09-07
Patch webdriver-manager to point to the proper chromeDriver location

RFSH:
A few weeks ago, we realized that the included `webdriver-manager` cannot install Chrome 115+. After investigating it, we realized that Chrome has decided to [change the location of newer versions](https://chromedriver.chromium.org/downloads/version-selection). To fix this, we need to update two functions in `webdriver-manager` to download it from the new location and unzip it. 

At first, I tried creating a repository and changing the `protractor.configuration.js` code to point to this new repository. But eventually, I found [`patch-package`](https://github.com/ds300/patch-package) from one of the comments in the original issue about this bug. This tool allows us to directly apply a patch on top of already installed node modules. So, in this PR, I'm now using this package to patch webdriver-manager. I think this method is way cleaner and easier to maintain/understand as everything is in one place.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -76,21 +76,6 @@ jobs:
           sudo bash ./test/ubuntu-travis-setup.sh
           sudo a2enmod webauthn
           sudo service apache2 restart
-      - name: Install hatrac
-        run: |
-          git clone https://github.com/informatics-isi-edu/hatrac.git
-          cd hatrac
-          sudo python3 ./setup.py install
-          sudo useradd -m -r hatrac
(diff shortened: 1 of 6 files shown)


RFSH:
I accidentally directly pushed these changes to master in this [commit](https://github.com/informatics-isi-edu/chaise/commit/722ea9889a0ae05d6405b03ab8e2d6c82c0212ce). Given that these changes are only going to affect the development mode (and only related to testing), I don’t think it’s worth reverting it right away. 

I'll revert or apply changes if we found issues with these changes. For now, I'm going to close this PR.

Issue #2349, opened 2023-08-22
Consistent spacing around text in buttons


jrchudy:
We've noticed that our button content renders incorrectly. There are 3 cases that should be looked into: 1. the view details button in the action column of recordset table; 2. the create, copy, and edit buttons in record app; and 3. other buttons render differently in different browsers with different zoom levels. This was originally found in Firefox + Linux (zoom level unknown).

## 1. This issue can be seen in every browser when the row in recordset table is a single line:
This issue looks simple to resolve, we added the style `padding: 5px 0 0 0;` to the action cell in recordset table. This rule should be adjusted to only apply when the content is >1 line.
 <img width="165" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/047110fd-37c1-49ab-ab68-e372a6e6c20e">
 
 
## 2. This issue can be seen in every browser regardless of OS is the spacing of `create`, `copy`, and `edit` in the record app:
Some investigation showed that this was happening specifically because of the use of `<a>` instead of `<button>`. This is intentional since we want users to be able to right click and open a context menu. There are other places (like `bulk edit` in related table actions header) that use `<a>` and don't have this "awkward" shifting of the text.
<img width="304" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/b31b5a0d-3c44-4e21-8885-fd311916d7ee">

#### Chrome
Reproduced broken behavior at 75%, 80%, and 90% zoom:
 - 75% - <img width="75" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/952e42b7-73d4-4d5f-be0b-db6258e13b91">
 - 80% - <img width="81" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/e5f391b1-ad14-4a18-9a4e-e428f4b4032e">
 - 90%- <img width="92" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/4ef4f82f-1750-4d4b-b339-2713da43372c">

Not reproduced at 67%, 100%, 110%, and 125% zoom:
 - 67% - <img width="70" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/97b395e7-5f74-4c9e-9adc-771008da888a">
 - 100% - <img width="105" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/9a4e7e4d-e1d1-4c8f-9696-76cd092498a6">
[...]

PR #2354, opened 2023-09-08
Fixing spacing around chaise buttons

PoornimaVKrishnan:
This PR will address [the issue](https://github.com/informatics-isi-edu/chaise/issues/2349)

Summary of changes:

The aim of this PR change is to fix the inconsistency in button spacing issue.

With this PR we have center aligned the buttons using display: 'inline-flex'.

Also for view details action button a height is given and padding is removed to fix the alignment
The deployed version is [here](https://dev.gudmap.org/~povijaya/chaise/record/#2/Common:Collection/RID=14-4KG6)

### DIFF
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -43,11 +43,6 @@
               font-size: 1rem;
               padding-top: 7px;
               width: 330px;
-
-              // overriding the title-button class
-              & > * {
-                display: inline-flex;
-              }
             }
           }
 
--- a/src/assets/scss/_record.scss
+++ b/src/assets/scss/_record.scss
@@ -43,6 +43,11 @@
               font-size: 1rem;
               padding-top: 7px;
               width: 330px;
+
+              // overriding the title-button class
+              & > * {
+                display: inline-flex;
+              }
             }
           }
 
--- a/src/assets/scss/_recordset-table.scss
+++ b/src/assets/scss/_recordset-table.scss
@@ -158,10 +158,9 @@
   }
 
   .action-btns {
+    // removing the padding since the table cell already has a padding of 5px
     text-align: center;
     font-size: 1.2rem;
-    // since we're increasing the size of icons, we should decrease the padding
-    padding: 5px 0 0 0;
 
     .chaise-btn {
       padding: 0;
@@ -172,6 +171,8 @@
 
     .chaise-view-details {
       font-size: 1.1em;
+      // giving a height to make it center aligned
+      height: 22px;
     }
 
     .disabled {
--- a/src/assets/scss/_recordset-table.scss
+++ b/src/assets/scss/_recordset-table.scss
@@ -157,10 +157,11 @@
     }
   }
 
+  // Changing the padding for the action button to align with the rest of the columns in the table
   .action-btns {
     text-align: center;
     font-size: 1.2rem;
-    padding: 5px 0 0 0;
+    padding: 3px 0 0 0;
     .chaise-btn {
       padding: 0;
       margin: -3px;
@@ -168,8 +169,9 @@
       min-height: 25px; // make sure the button has default height
     }
 
+    // reducing the font-size to make the alignment of view-details better with other action buttons
     .chaise-view-details {
-      font-size: 1.1em;
+      font-size: 1em;
     }
 
     .disabled {
--- a/src/assets/scss/_recordset-table.scss
+++ b/src/assets/scss/_recordset-table.scss
@@ -157,11 +157,12 @@
     }
   }
 
-  // Changing the padding for the action button to align with the rest of the columns in the table
+  // Changing the padding and adding line-height for the action button to align with the rest of the columns in the table
   .action-btns {
     text-align: center;
     font-size: 1.2rem;
     padding: 3px 0 0 0;
+    line-height: 1.5;
     .chaise-btn {
       padding: 0;
       margin: -3px;
--- a/src/assets/scss/_recordset-table.scss
+++ b/src/assets/scss/_recordset-table.scss
@@ -159,32 +159,41 @@
 
   // Changing the padding and adding line-height for the action button to align with the rest of the columns in the table
   .action-btns {
-    $_space-around-buttons: 3px;
+    // the whole row min height is 30px, and buttons are 22px.. so we have to have 4px above and below
+    $_space-around-buttons: 4px;
 
     text-align: center;
     font-size: 1.2rem;
     padding: $_space-around-buttons 0 0 0;
     line-height: 1.5;
 
+    // make sure items stay in the center
+    .action-btns-inner-container {
+      display: flex;
+      align-content: center;
+      justify-content: space-around;
+    }
+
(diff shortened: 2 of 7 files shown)


src/assets/scss/helpers.scss
```diff
@@ -89,6 +88,10 @@
   -moz-user-select: none;
   -ms-user-select: none;
   white-space: nowrap;
+  // ficxing button spacing issue. Fix- center aligning the buttons
+  display: inline-flex !important;
```
RFSH, line 92: You mentioned that you're using `!important` here to override the displays that we've defined in some buttons. 

I was thinking about it, and Instead of `!important` let's just remove the display definition from those other places. We should keep the rules in here as general as possible and should allow other places to be able to override them.

src/assets/scss/helpers.scss
```diff
@@ -89,6 +88,10 @@
   -moz-user-select: none;
   -ms-user-select: none;
   white-space: nowrap;
+  // ficxing button spacing issue. Fix- center aligning the buttons
```
RFSH, line 91: Fix the typo

src/assets/scss/helpers.scss
```diff
@@ -89,6 +88,10 @@
...
+  display: inline-flex !important;
+  align-items: center;
+  justify-content: center;
```
jrchudy, line 94: Add `line-height: normal`. Add a comment explaining what `normal` does and why it works here. Or try other values and explain why that might be better.

Adding `line-height` here will allow for the `<span>` tags to inherit the value and use it for the text part of the button.

jrchudy (changes requested):
I tested these changes myself in chrome on window and still see the issue. I added a comment about adding `line-height` to the `chaise-btn` styles so the children of the button (usually `<span>` tags) will use this property instead of what bootstrap is defining.

I'm unsure of which value line-height should be but using `normal` seemed to be the best so it is consistent for our different button sizes too.

RFSH (changes requested):
Related to positing the "action buttons" in recordset table, ideally we want the buttons to be aligned with the first line of the other cells. We also want to preserve space as much as we can, so we still want to reduce the bottom padding and you should not remove the rule in line 164. 

- In Chrome the alignment looks better but I still think it's a few pixel off from the middle. 
Chrome

- In Firefox the added height is actually making the view button to be way above the rest. 
Firefox

For chrome, I think if you use `padding: 3px 0 0 0;` in line 164 instead of 5, the alignment would be much better.

Make sure you're testing your changes on Chrome, Firefox, and Safari. This might be related to line-height that Josh mentioned. Maybe different browsers are computing different line-heights?

PoornimaVKrishnan:
@RFSH I have addressed all comments. Please have a look. 
[here](https://dev.gudmap.org/~povijaya/chaise-2/record/#2/Common:Collection/RID=14-4KG6)
I tested on chrome and firefox in different zooming levels

RFSH:
@PoornimaVKrishnan Please be more attentive. I'm assuming you were only focused on the create/copy/edit/delete button. But the "Export" and "Permalink" buttons in recordset now show up in two lines.

Also did you look at my comment? I asked you to try to fix the alignment of the action buttons but you only reverted the changes that you did. Like I said try `3px` padding and see if that is better than what we have or not. In our current version the action buttons are a few pixel lower than they should be.

PoornimaVKrishnan:
> @PoornimaVKrishnan Please be more attentive. I'm assuming you were only focused on the create/copy/edit/delete button. But the "Export" and "Permalink" buttons in recordset now show up in two lines.
> 
> Also did you look at my comment? I asked you to try to fix the alignment of the action buttons but you only reverted the changes that you did. Like I said try `3px` padding and see if that is better than what we have or not. In our current version the action buttons are a few pixel lower than they should be.

I fixed the issue. Added comments. `3px` looked better

src/assets/scss/_record.scss
```diff
@@ -43,6 +43,11 @@
               font-size: 1rem;
               padding-top: 7px;
               width: 330px;
+
+              // overriding the title-button class
+              & > * {
+                display: inline-flex;
+              }
             }
```
PoornimaVKrishnan, line 51: added this to override title-button class, giving the change there was causing the export and permlink button to come in two lines

src/assets/scss/_recordset-table.scss
```diff
@@ -157,21 +157,21 @@
...
   }
 
+  // Changing the padding for the action button to align with the rest of the columns in the table
   .action-btns {
     text-align: center;
     font-size: 1.2rem;
-    // since we're increasing the size of icons, we should decrease the padding
-    padding: 5px 0 0 0;
-
+    padding: 3px 0 0 0;
```
PoornimaVKrishnan, line 164: Decided to go with `3px` as the alignment looked fine with other columns

src/assets/scss/_recordset-table.scss
```diff
@@ -157,21 +157,21 @@
...
       margin: -3px;
       height: auto; //override the default button height
       min-height: 25px; // make sure the button has default height
     }
 
+    // reducing the font-size to make the alignment of view-details better with other action buttons
     .chaise-view-details {
-      font-size: 1.1em;
+      font-size: 1em;
     }
```
PoornimaVKrishnan, line 175: edit pencil icon was few pixels off from view-details. reducing the font-size looks better in all zoom levels and browsers

src/assets/scss/_recordset-table.scss
```diff
@@ -157,21 +157,22 @@
...
 
+  // Changing the padding and adding line-height for the action button to align with the rest of the columns in the table
   .action-btns {
     text-align: center;
     font-size: 1.2rem;
-    // since we're increasing the size of icons, we should decrease the padding
-    padding: 5px 0 0 0;
-
+    padding: 3px 0 0 0;
+    line-height: 1.5;
```
PoornimaVKrishnan, line 165: added line-height as it was not aligning in windows when i tested in saucelabs. this is the closest alignment i could get for the action buttons

RFSH:
@PoornimaVKrishnan  The main change in this PR is addition of `display: inline-flex` to buttons. We should not override this rule to try to solve other issues. Because of that change, the "Export" and "Permalink" button labels in recordset were misaligned. So I pushed a fix for this. To fix this we just needed to add `display: inline-block` to the `.export-menu.dropdown`.

Also while the action buttons spacing were a lot better than what we had before, the alignment was still ugly in the one-line case. This was more obvious on the facet and foreign key popups. So I also pushed some changes related to that.

@jrchudy To me this PR is ready to be merged. Can you please review it again?

RFSH:
After discussing this with @hongsudt, we decided to reduce the size of action buttons. This was done because in the single select mode the button was taking up the whole available height and everything was very close to each other.

While doing this change I also fixed the display of checkboxes on Firefox. Firefox doesn't like the way we were attaching `:before` and `:after` to input, so I had to move it to `label` instead. Now all browsers will display the same checkbox.

jrchudy (approved):

PR #2359, opened 2023-09-20
Improve UX with bulk edit in record app

jrchudy:
This pull request changes how bulk edit is displayed for related tables in record app when there are no rows in the related table.

Added test cases to the record page test case that was already testing empty related tables instead of adding to "related table" spec.


src/components/record/related-table-actions.tsx
```diff
@@ -764,11 +769,13 @@ const RelatedTableActions = ({
             }
           >
             <a
-            className={`chaise-btn bulk-edit-link ${
-              tertiary ? 'chaise-btn-tertiary dropdown-button' : 'chaise-btn-secondary'
-            }`}
+              className={`chaise-btn bulk-edit-link
```
RFSH, line 772: Do we want to change the tooltip when the button is disabled (Similar to how we're changing the "Add"/"link" button tooltips when they are disabled?

RFSH (approved):

PR #2367, opened 2023-10-20
Change order of assets in webpack config

RFSH:
We recently realized that bootstrap rules are overwriting chaise rules in deriva-webapps. Upon looking at it, I realized that the chaise CSS files were added before the bootstrap ones. 

I always assumed that webpack is going to add CSS files based on the `import` statements that we have. But it seems like the `priority` we've set for the `cacheGroups` is dictating the order. Based on webpack's documentation, the `priority` is to ensure which rules should be prioritized when a given `test` matches multiple. That's why I had the default `node_modules` as the lowest priority.

So, to fix this, I rearranged the priorities and ensured the default `vendor-rest` excludes other rules. This way, the `priority` is just for dictating the order of added asset files and not tie-breaking the rules.

While this fixes the CSS issues, it's also rearranging the JS files, so we have to test all apps to make sure nothing is broken before merging this.

### DIFF
--- a/webpack/app.config.js
+++ b/webpack/app.config.js
@@ -209,35 +209,40 @@ const getWebPackConfig = (appConfigs, mode, env, options) => {
         maxSize: 500000,
         hidePathInfo: true,
         name: 'common',
+        /**
+         * the noted priority is also changing the order of include statements in the output html.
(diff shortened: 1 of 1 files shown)


RFSH:
The CI test cases passed without any issues. I also ran test cases locally and don't see any issues. So I'm going to merge this PR.

PR #2370, opened 2023-10-30
File upload causing server to hang

jrchudy:
Fixed bug with upload not properly queueing multiple files for upload causing the server to not be able to handle the incoming number of requests when uploading multiple files (5+). Fixed a bug where after aborting, file upload would still complete and submit the recordedit page.

The changes I made for these bugs:
 - changed `erred` and `aborted` from `useState` to `useRef` since they are only ever used inside of other functions which are callbacks from promises.
   - when functions are callbacks inside of promises, the react "state" is not updated for those functions since they get defined in the calling tree once the first function is called
   - using a "ref" is meant to not trigger rerenders and therefore is mutable/accessible during long function calls in promise callbacks
 - Changed `++` to `val + 1` and `--` to `val - 1` since this wasn't being "set" properly in setState functions
 - added a useEffect that watches for `checksumCompleted` count changes and triggers creating upload jobs once all checksums were complete
   - since this is a state variable, it won't update in those functions. filesCt was improperly using a state variable too that was not updating
   - therefore `onChecksumCompleted` would get to `if (checksumCompleted === filesCt)` and both `checksumCompelted` and `filesCt` are `0`. So this would trigger `checkFileExists` N times where N is the number of files
     - `checkFileExists` iterates over all rows and all files, it's intended to be called once ALL checksums are complete

### DIFF
--- a/src/components/modals/upload-progress-modal.tsx
+++ b/src/components/modals/upload-progress-modal.tsx
@@ -48,8 +48,8 @@ const UploadProgressModal = ({ rows, show, onSuccess, onCancel }: UploadProgress
   const [isCreateUploadJob, setIsCreateUploadJob] = useState<boolean>(false);
   const [isFileExists, setIsFileExists] = useState<boolean>(false);
 
-  const [erred, setErred] = useState<boolean>(false);
-  const [aborted, setAborted] = useState<boolean>(false);
+  const erred = useRef<boolean>(false);
+  const aborted = useRef<boolean>(false);
 
   // total number of files being uploaded
   const [filesCt, setFilesCt] = useState<number>(0);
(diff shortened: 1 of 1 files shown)


RFSH (approved):
The changes look good to me. I should've paid more attention when we were finalizing the recordedit migration. The `erred` and `aborted` are not used during the render logic or hook change detection, so it makes sense to have them as `useRef` and not `useState`.

I locally ran the upload test cases that we have (`multi-edit.conf.js`, `add.conf.js`) and everything is working properly.

P.S. I'm not sure entirely if this is related or not, but the `multi-edit` spec is noticeably faster now (when I was finalizing the test cases for React migration I noticed a delay in the upload step of this spec which was causing some timeouts).

PR #2382, opened 2023-12-19
Clone spinner

jrchudy:
Add a spinner over recordedit page when more forms are added (1 at a time or many). More details about the specifics of the implementation are in the comments in the code.

### DIFF
--- a/src/components/recordedit/form-row.tsx
+++ b/src/components/recordedit/form-row.tsx
@@ -139,6 +139,10 @@ const FormRow = ({
       if (!formsRef || !formsRef.current || !showCloneSpinner) return;
 
       if (formsRef.current.children.length === forms.length) setShowCloneSpinner(false);
+    // trigger this when number of forms changes
+    // `callAddForm` in /components/recordedit.tsx calls addForm(#_forms_to_add) which will then set the `forms` state variable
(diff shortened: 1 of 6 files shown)


src/components/recordedit/recordedit.tsx
```diff
@@ -297,7 +298,19 @@ const RecordeditInner = ({
     }
   }, [formProviderInitialized]);
 
+  // This useEffect triggers when addFormsEffect is set to true when "clone" is clicked
```
RFSH, line 301: Thank you for explaining it. I was confused why this is needed. Is this something worth documenting in our dev guides? As a bullet in React/TypeScript section where we're talking about hooks.

Also for readability I think it's better if we use jsdoc style of comments for these lengthy multi-line comments.
```
/**
 *
 */
```

src/components/recordedit/form-row.tsx
```diff
@@ -125,6 +128,24 @@ const FormRow = ({
     };
   }, []);
 
+  if (columnModelIndex === 0) {
```
RFSH, line 131: I think it's better if you move this check to be inside the `useEffect` line 139. React itself says that we should avoid doing something like this.

RFSH (approved):

Issue #2386, opened 2024-01-09
Update Font-Awesome to get rid of Firefox warning


RFSH:
A few months ago Firefox started showing a warning related to the Font-Awesome font:

```
downloadable font: Glyph bbox was incorrect (glyhp ids ...)
```

According to [Font-Awesome](https://github.com/FortAwesome/Font-Awesome/issues/19925) this wasn't a breaking issue but it's something that they have fixed now.

We should make sure we're also taking care of the following after changing the included version:
- Make sure the icons looks the same after upgrade and nothing breaks.
- Publish a new version so deriva-webapps start using the new version.
- Update the version that we're using in openseadragon-viewer repo.
- Update the version that we're using for AngularJS implementation (used by navbar lib).

PR #2390, opened 2024-01-16
Use latest version of fontawesome

RFSH:
As I described in [this issue](https://github.com/informatics-isi-edu/chaise/issues/2386), this is mainly done to get rid of the warning on Firefox.

The only thing that has changed is the `pencil` (edit) icon:

| old | new |
| --- | --- |
| Old   | New   |

Other changes:
- This also includes a fix to the alignment of `.chaise-btn-group`. In Firefox, the `"align-item": baseline` was causing issues as it couldn't correctly align items with different heights. So, while the content was properly aligned, the bounding box of each button would not align with the rest. Now, we only add this rule for the cases where we're not showing any borders (currently only used in one place, but I decided to use a general name in case we want to do something similar later).

P.S. We should publish a new version of Chaise so deriva-webapps can use these changes as well. I haven't done it yet because I wasn't sure if there are any other pending changes that we would like deriva-webapps to use (and so we should do the version update after merging them all)


jrchudy (approved):

RFSH:
I'll go ahead and merge this PR. There are some other changes pending, so I'll bump the version with those changes.

PR #2392, opened 2024-01-17
Improve show more to run after images load

jrchudy:
Add onload to image tags in recordset table rows that updates a state variable to trigger overflow logic when all images have loaded. When `rowValues` changes, the sensor will be attached and the DOM will be searched for any `` tags that might be present with `-chaise-post-load` class. Each `` will have an `onload` function attached to it that will update another state variable to compare against the total.

NOTE: It's okay if the overflow logic triggers when 1 out of 4 images loads and again when 4 out of 4 images load. The important thing is that this logic triggers when ALL images have loaded to ensure the show more displays properly.


src/components/recordset/table-row.tsx
```diff
@@ -153,30 +156,59 @@ const TableRow = ({
...
-        // images take ~200ms to load so wait at least that long before running
-      }, 200)
+      () => initializeOverflows()
     )
 
     setSensor(tempSensor);
+    
+    // fetch all <img> tags with -chaise-post-load class and keep count of the total
+    // attach an onload function that updates how many have loaded
+    const imgTags = rowContainer.current.querySelectorAll('img.-chaise-post-load');
```
RFSH, line 179: It seems like we're not consistent in how we're attaching this class in ermrestjs:

- If you use the `` syntax, we're attaching the class to the `img` tag like the selector that you have.
- If you use the `:::image` syntax, we're attaching the class to `figure` element that is wrapping the image.

So I think you should change this selector to `img.-chaise-post-load, .-chaise-post-load img`.

And let's not hardcode `-chaise-post-load` here and instead add it to `CLASS_NAMES` that we have in the `constants.ts`.

src/components/recordset/table-row.tsx
```diff
@@ -153,30 +156,59 @@ const TableRow = ({
...
+    setNumImages(imgTags.length);
+    
+    imgTags.forEach((image: HTMLImageElement) => {
+      image.onload = () => {
```
RFSH, line 183: Let's use `addEventListener` and `removeEventListener` instead of `onload`.

RFSH: I don't think you need to have `numImages` and `numImagesLoaded` state variables. They can just be `useRef`, and then in the `load` event listener you would check whether all the images have been loaded and run the `initializeOverflows` function.

RFSH (changes requested):
I don't think you need to involve the component state in this logic and this could be done with `useRef`. That way we're not causing unnecessary re-renders.

RFSH (changes requested):
I was looking around to see if using `load` event handler is enough, and it seems that's not enough.

- The image might already be loaded when we try to add the `load` event listener. To get around this, we can use the `element.complete` to see if the image has been loaded or not.
- Images might error out.

So to solve this,

1. Change line 179 to be something like this
```tsx
const imgTags = Array.from<HTMLImageElement>(rowContainer.current.querySelectorAll(
      `img.${CLASS_NAMES.CONTENT_LOADED}, .${CLASS_NAMES.CONTENT_LOADED} img`
    )).filter(img => !img.complete);
```

2. Add `error` event listeners
```tsx
imgTags.forEach((image: HTMLImageElement) => {
      image.addEventListener('load', onImageLoad);
      image.addEventListener('error', onImageLoad);
    });

...


return () => {
      tempSensor.detach();
      imgTags.forEach((image: HTMLImageElement) => {
        image.removeEventListener('load', onImageLoad);
        image.removeEventListener('error', onImageLoad);
      });
    }
```

RFSH (approved):

Issue #2393, opened 2024-01-25
Support inline comments for columns on the recordedit page


RFSH:
As the title suggests, we should support inline comments for columns on the recordedit page. This way, data-modelers can provide instructions on how users should populate the values.

UI-wise, it should behave similarly to the `multi-form input row`. The inline comment should be an extra row expanding the row of inputs and should remain in place while the user is scrolling horizontally.

Issue #333, opened 2016-05-04
Set max # of copies


howdyjessie:
Set a limit on how many rows the user can duplicate and append to the form to avoid the browser choking up in the case when a user selects a large number. This may require both a deployment-independent hard limit and a deployment-specific soft limit(?)

karlcz:
I think we should just drop the notion of having a "number of duplicates" control and have a 1-click per duplicate control labeled "Add record".

robes:
I think, per Karl's comment above, that this issue is no longer valid.

PR #2396, opened 2024-02-05
Support inline comments for columns in recordedit app

RFSH:
This PR will add support for inline comments for columns in the recorded it app. As described in the #2393 issue, the inline comment will look similar to the "select all/some" row.

So far, we've only added inline comments for the record title and related table section. So, we've intentionally increased their font size. I'm using the same class name for inline tooltips here so that it will use the same `16px` font. We should discuss whether we want to apply the same font here or not (I don't think we should use a bigger font for inline comments of columns). 

Screenshot 2024-02-05 at 10 05 38 AM

P.S. As part of [the recent changes to comment APIs in ermrestjs](https://github.com/informatics-isi-edu/ermrestjs/pull/1000) I've already implemented the support for `column_comment_display` (and `column_display` for columns), so no changes would be needed for this on ermrestjs.


RFSH:
@jrchudy This PR is ready to be reviewed. I applied the UI changes that we talked about in the meeting:

- The inline comment scrollable logic should be the same as the inputs (not the multi-form input)
- The inline comment is now greyer (#666 instead of #333), and also smaller (13px)
- There's a hr between the inline comment and inputs (instead of relying on border). This was done to make the inline comment look like part of the inputs (the line doesn't go all the way do the edges and it's just a separator)

I also added a new section to `custom-css.md` to explain how these styles can be customized.

Issue #2397, opened 2024-02-05
Proper support for domain_filter in the select all/some input (multi form input)


RFSH:
## Summary

The `domain_filter` allows data modelers to limit the options in a foreign key input. This can be done by writing a static filter or using the currently selected values in the form. But since the "select all/some" input doesn't belong to any of the forms, we cannot support `domain_filter` in all the cases.

The simplest solution is disabling the "select all/some" button when the foreign key column has `domain_filter`. But that would be too limiting, and in this issue, we would like to explore ways to support both simultaneously. This would be useful for PDB, as we discussed [here](https://github.com/informatics-isi-edu/protein-database/issues/184#issuecomment-1924655802).

## Details

I will focus on foreign key columns with `domain_filter` in the following. Foreign key columns without this annotation should continue to offer the "select all/some" feature without any changes. 

As long as the computed `domain_filter` (or more generally the `Reference.uri`) in all the forms is the same, we should be able to support `domain_filter` in "select some/all" input. So, we could compute the `domain_filter` for all the forms and disable the "select some/all" toggle button if some forms had a different value. But that would mean we have to rerender the `KeyColumn` component on each user input, which would slow the page down. 

After discussing this, we decided to show an error when users attempt to open the foreign key picker instead. We will only run this expensive operation when users click on the foreign key. We are also going to leave the error and only reevaluate when users click again (after each click, we're showing a spinner since calculating it might be slow).

### Error message

We should add `pattern_sources` property to `domain_filter` so data-modelers can list the columns that are used in the pattern. And then use them to show an error like this:

```
This feature is constrained by "x", "y", and "z". Make sure all the records you want to set "Asym Id" for, have the same values for those fields. Try again after upadting those fields.
```

If this property is missing, the error should be:

```
This feature is constrained by other fields (hint: `domain_filter_string`). Make sure all the records you want to set "Asym Id" for, have the same values for those fields. Try again after upadting those fields.
```

P.S. While discussing this, we realized that even if users don't use the "select all/some", the final selected values might not match the `domain_filter`s. So, as part of a separate task, we should also do similar checks before attempting to save (#2401).

PR #2400, opened 2024-02-17
Proper support for domain_filter in the select all/some input (multi form input)

RFSH:
This PR makes sure `domain_filter` is properly supported in multi-form input.

Issue with more info: #2397


src/components/recordedit/multi-form-input-row.tsx
```diff
@@ -257,10 +263,79 @@ const MultiFormInputRow = ({
...
+      if (formIndex === 0) {
+        domainFilterFormNumber = formNumber;
+        computedRefURI = ref.uri;
+      } else {
+        return computedRefURI === ref.uri;
+      }
+      return true;
+    });
+
+    // some domain-filters are different, so we cannot proceed
```
jrchudy, line 297: This comment should be updated for the if statement it is coming before or move it after this if statement.

src/components/recordedit/multi-form-input-row.tsx
```diff
@@ -257,10 +263,79 @@ const MultiFormInputRow = ({
...
+    let computedRefURI = '';
+    const areAllSame = activeForms.every((formNumber: number, formIndex: number) => {
+      const ref = createForeignKeyReference(cm.column, reference, formNumber, foreignKeyData, getValues);
+      if (formIndex === 0) {
+        domainFilterFormNumber = formNumber;
+        computedRefURI = ref.uri;
+      } else {
+        return computedRefURI === ref.uri;
+      }
+      return true;
```
jrchudy, line 294: I think this would read cleaner as:
```
if (formIndex === 0) {
  domainFilterFormNumber = formNumber;
  computedRefURI = ref.uri;
  return true
}

return computedRefURI === ref.uri;
```

jrchudy (approved):
Changes look good. It works how I expect after messing around with the test case you added.

I left a comment about one of your comments which seemed out of place and suggested changing an if/else to be more straightforward.

Don't forget to change the command in e2e.yml

PR #2402, opened 2024-02-21
Change some buttons to links

### LINKED ISSUE #2051: Buttons vs Links
In different places in chaise we have links, links that look like buttons,  and buttons that call functions for navigation. 

We need to evaluate the desired behavior for each of these cases in regards to opening in the same tab or a new tab. Some feedback from others included:
 - [Rob] we went through this some time ago that we wanted to make sure that all “internal” links are opened in same window. so its inconsistent for Go To Rid to open a new window/tab, actually.
 - [Karl] Please also make internal link buttons all actually be links (so browser context menu works on them to allow open-in-tab, copy-link-location etc.).  I realize there are some buttons which are not links (submit, facet controls/expand/collapse).

jrchudy:
In this PR are changes to change 3 buttons in our app to links. Those buttons are:
 - `create` on recordset
 - `bulk edit` on recordset
 - `add records` on record in the related table actions

This PR issue resolves #2051.

### DIFF
--- a/src/components/modals/iframe-field-modal.tsx
+++ b/src/components/modals/iframe-field-modal.tsx
@@ -18,6 +18,7 @@ import { ChaiseAlertType } from '@isrd-isi-edu/chaise/src/providers/alerts';
 
 // utils
 import { isStringAndNotEmpty } from '@isrd-isi-edu/chaise/src/utils/type-utils';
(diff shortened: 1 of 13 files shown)


jrchudy:
We are not planning on making these change at this time since it would create inconsistency if we don't change all 4 proposed buttons to links. More details in the attached issue

Issue #1059, opened 2017-04-14
More problems with float values


karlcz:
[Example record](https://synapse-dev.isrd.isi.edu/chaise/recordedit/#1/Zebrafish:Behavior/RID=17DJ)
There is a small value `1e-12 ` stored in the table.

1. When I load the editor it shows the value `0.000000000001` correctly
2. If I edit another field like the `Trial Counts` and submit, I get a validation error "please enter a decimal value"
3. If I edit the field to remove and re-add the trailing `1` digit it validates
4. The submit works (the update was made to the server)
5. But the app sometimes dies with a terminal error instead of displaying the success (this seems to be intermittent)

It seems like both error and success flags flash at the top of the screen on success.  So I suspect a race condition...

karlcz:
Sadly I accidentally closed the tab and can't yet reproduce the error to capture the terminal error details...

jrchudy:
This is a weird issue. If you look at this page in Chrome, you notice that the value is not automatically changed from `1e-12` to `0.000000000001`. This causes a validation error 100% of the time until the user changes the value to a float format.

If you open this in Firefox, you can see that the browser changes the value for the input field. We are not doing that in our code so what is being represented in the UI and what is being validated is out of sync.

Maybe we need to format values that appear in scientific notation as their true float value before passing that to the UI. Maybe a change in what is returned from `Tuple.values` in `ermrestJS` would be more appropriate to fix this so the UI only deals with numerical floats.

Moving this to `Blocked` until @hongsudt and @karlcz come back.

karlcz:
I confirmed that this bug is still present.  I created a new test record illustrating it:

https://synapse-dev.isrd.isi.edu/chaise/recordedit/#1/Zebrafish:Behavior/RID=17DJ

The original example record no longer exists..

jrchudy:
I can create records for this table but I cannot update any. Could you update the permissions for `isrd-testers`?

jrchudy:
Moving this from blocked to backlog. @hongsudt maybe this is something we can look into adding better support for?

RFSH:
We had a brief discussion about precision issues in recordedit:

- The precision issue is not just about the ermrestjs/chaise logic truncating the values. For `float`, ermrest might also truncate the value stored in the database. So ideally, we should have a "dirty" flag for our inputs to check if an input has been touched or not. And if it hasn't, we should avoid sending that value.

[...]

PR #2405, opened 2024-02-27
modify float regular expression to allow for scientific notation 

jrchudy:
This PR is ready for review. We found that the previous validation wasn't handling scientific notation properly even though scientific notation is a valid number. Changes are made to use functional validation for float and number input types instead of a regular expression. Using a regular expression relies on us writing a proper regular expression and maintaining it as javascript and the browser change. The functional solution uses the javascript functions `isNaN`, `parseFloat`, and `isFinite` to verify the supplied value is a valid Number type value.

More details in issue #1059.


src/components/input-switch/numeric-field.tsx
```diff
@@ -8,16 +8,17 @@ import { VALIDATE_VALUE_BY_TYPE } from '@isrd-isi-edu/chaise/src/utils/input-uti
 
 const NumericField = (props: InputFieldProps): JSX.Element => {
   let rules;
-  if (props.type === 'integer2' || props.type === 'integer4' || props.type === 'integer8') {
+  if (props.type === 'int') {
+    // only 'int' uses a regular expression for validation
     rules = {
-      validate: VALIDATE_VALUE_BY_TYPE[props.type]
+      pattern: VALIDATE_VALUE_BY_TYPE[props.type]
```
RFSH, line 14: I feel like at this point it's easier if we also change the int validator to be a function. And then just use the regular expression inside it. That's exactly what we're doing for other `int2`, `int4`, etc anyways. 

That way all of our validators are just functions and we don't need to worry about pattern vs. validate here.

src/utils/input-utils.ts
```diff
@@ -278,9 +278,9 @@ const integerFieldValidation = {
   message: ERROR_MESSAGES.INVALID_INTEGER
 };
 
-const numericFieldValidation = {
-  value: dataFormats.regexp.float,
-  message: ERROR_MESSAGES.INVALID_NUMERIC
+const numericFieldValidation = (value: any) => {
+  if (!value) return;
+  return !isNaN(parseFloat(value)) && isFinite(value) || ERROR_MESSAGES.INVALID_NUMERIC;
```
RFSH, line 283: This doesn't properly handle leading/trailing whitespace. So for example `123123.23  ` and `   123` are treated like valid numbers. 

ermrest is actually properly ignoring the whitespace when we send these values, but we most probably should just invalidate such values, right?

If we want to disallow any white space we can just do `/\s/g.test(s)` or if we want to only disallow leading/trailing we could `.trim()` and see if they match or not .

RFSH (commented):
Apart from the comments, I think you should update one of our recordedit test cases to use scientific notation.

RFSH (approved):

PR #2407, opened 2024-02-29
2363 navbar submenu stays open

ladukaraniket:
### It was an issue with the way dropdown behaves in react-bootstrap. This change should fix it

TSvGhNUMCv

### DIFF
--- a/src/components/navbar/navbar.tsx
+++ b/src/components/navbar/navbar.tsx
@@ -56,7 +56,7 @@ const ChaiseNavbar = (): JSX.Element => {
   /**
    * Keeps track of most recently opened dropdown
    */
-  const [openDropDown, setOpenDropDown] = useState<number |undefined>(undefined);
+  const [openDropDown, setOpenDropDown] = useState<number | undefined>(undefined);
 
   const dropdownWrapper = useRef<any>(null);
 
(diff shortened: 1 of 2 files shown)


src/components/navbar/navbar.tsx
```diff
@@ -53,6 +53,10 @@ const ChaiseNavbar = (): JSX.Element => {
   const [showRidSpinner, setShowRidSpinner] = useState(false);
   const [topBanners, setTopBanners] = useState<NavbarBanner[]>([]);
   const [bottomBanners, setBottomBanners] = useState<NavbarBanner[]>([]);
+  /**
+   * Keeps track of most recently opened dropdown
+   */
+  const [openDropDown, setOpenDropDown] = useState<number | undefined>(undefined);
```
RFSH, line 59: Let's rename this to `openedDropdownIndex`. You can also just use `number` as type and not pass anything for the default value. In the comment you should also mention why you've added it (mention the bug)

```tsx
  const [openedDropdownIndex, setOpenedDropdownIndex] = useState<number>();
```

src/components/navbar/navbar.tsx
```diff
@@ -185,7 +189,16 @@ const ChaiseNavbar = (): JSX.Element => {
   };
 
   // TODO: onToggle event type
-  const handleNavbarDropdownToggle = (isOpen: boolean, event: any, item: MenuOption) => {
+  const handleNavbarDropdownToggle = (isOpen: boolean, event: any, item: MenuOption, index: number) => {
+    /**
+     * Update the state to reflect most recently opened dropdown
+     */
+    if (isOpen) {
```
RFSH, line 196: You can write this in one line and I think it's easier to read that way
```tsx
setOpenedDropdownIndex(isOpen ? index : undefined);
```

RFSH (changes requested):
Apart from the minor code style changes that I mentioned in my comments, you should also apply similar changes to `navbar-dropdown.tsx`. 

In the `navbar-dropdown.tsx` case the bug is that the previous item is still highlighted and has the `show` class. While this is a less annoying bug, applying the same change will fix it.

RFSH (approved):

PR #2436, opened 2024-04-13
change footer test from protractor to playwright

jrchudy:
This PR updates the footer spec from using protractor to using playwright.


test/e2e/specs/all-features-confirmation/footer/footer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+            await page.goto(`${baseURL}${PAGE_URL}`);
+        });
+
+        test('Checking footer in ' + name + ' page:', async ({ page }) => {
+            const footerMain = PageLocators.getFooterContainer(page);
+            const footerLink = PageLocators.getFooterLink(page);
+            
+            await test.step('footer should be visible on load', async () => {    
+                await footerMain.waitFor({ state: 'visible' });
```
RFSH, line 24: While this works and I think I actually done something like this in navbar tests, doing the following is better

```tsx
expect.soft(footerMain).toBeVisible();
```

I'm going to rewrite the `waitFor`s that I have in navbar specs to use expect instead.

test/e2e/specs/all-features-confirmation/footer/footer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            })
+
+            await test.step('Page footer link should match with "privacy-policy"', async () => { 
+                // plink should be the `https://${server}/privacy-policy/`
+                let plink = await footerLink.getAttribute('href') || '';
+
+                // slice off the last '/'
+                plink = plink.slice(0, -1);
+                // compare everything after the last '/' after slicing to verify the path
+                expect.soft(plink.substring(plink.lastIndexOf('/') + 1)).toEqual('privacy-policy');
```
RFSH, line 38: In general, if you can use [the built-in assertions](https://playwright.dev/docs/test-assertions), we should use that instead of resolving promises and calculating things ourselves. For this purpose you could use `toHaveAttribute` which will allow you test by exact match or regular expression. You could use the regular expression for this case

```tsx
expect.soft(plink).toHaveAttribute('href', /privacy-policy/);
```

If you look for the test cases that I've done, you can see both `toHaveAttribute` and `getAttribute` similar to what you have. I tried to only use `getAttribute` when I couldn't easily come up with a regular expression, or when we wanted to test `toContain`.

RFSH (approved):

PR #2446, opened 2024-04-25
Saved query playwright

jrchudy:
This PR is for rewriting `saved-query.spec.js` in playwright with typescript. Many more selectors added and changes use as many existing functions as possible.

I'm unsure of using `beforeAll` to update the catalog annotation but it should be fine since it doesn't affect the other test case in this document. Moving the `beforeAll` into the following `test` block throws an error saying it can't be defined there.

Opening this to run tests in github and see if they pass in the context of all tests.


test/e2e/locators/recordset.ts
```diff
@@ -161,6 +165,68 @@ export default class RecordsetLocators {
...
+    return facet.locator('.chaise-checkbox label')
+  }
+
+  static getFacetOption(facet: Locator, optionIdx: number) {
+    return facet.locator(`.checkbox-${optionIdx}`);
+}
 
+  /* range facet selectors */
+  // there's integer/float/date/timestamp inputs
+  static getRangeInput(facet: Locator, className: string): Locator {
```
RFSH, line 191: I think we should improve these functions instead just rewriting them in playwright. Passing the actual class from the test completely defeats the purpose of these functions. Our goal is to keep all the implementation details hidden in these functions. So if we want to change a class/implementation we would just change these functions and not the actual specs.

You could do something similar to what we did for [here](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/locators/recordedit.ts#L143)). So,

```ts
static getFacetRangeInputs(facet:Locator, isTimestamp?: boolean) : Locator {
  return {
    minInput: facet.locator(...),
    maxInput: facet.locator(...),
    minClear: ..,
    maxClear: ...,
    ...
  }
}
```

test/e2e/locators/recordset.ts
```diff
@@ -161,6 +165,68 @@ export default class RecordsetLocators {
...
+    return facet.locator(`.${className}`);
+  }
+
+  static getInputClear(facet: Locator, className: string): Locator {
+    // NOTE: same as getRangeInput selector
+    return facet.locator(`.${className}`);
+  }
+
+  static getRangeSubmit(facet: Locator): Locator {
```
RFSH, line 200: You're adding functions to the "Recordset" class. So try to be more specific in the names. Just adding Facet should be enough (`getFacetRangeSubmit`, `getFacetHistogram`).

test/e2e/locators/recordset.ts
```diff
@@ -161,6 +165,68 @@ export default class RecordsetLocators {
...
+
+  // all dropdown menu items
+  static getSavedQueryOptions(container: Page | Locator): Locator {
+    return container.locator('.saved-query-menu-item');
+  }
+
+  // recordedit option to save a query
+  static getSaveQueryOption(container: Page | Locator): Locator {
+    // substring matching
+    return container.locator('text=Save current search criteria');
```
RFSH, line 224: I didn't know you could do something like this. Playwright documentation actually encourages us to use [`getByText()`](https://playwright.dev/docs/locators#locate-by-text) which is similar to what you have done here. That way instead of relying the actual implementation details (class, id, etc), we're going by what users sees.

But given that this required a lot of rethinking in our locators, I kept using what we had. As long as we make sure we're keeping all these implementation details in the locators, we could eventually change all to use [these recommended locators](https://playwright.dev/docs/locators) instead.

test/e2e/locators/recordset.ts
RFSH: By the way the `.range-input-submit-btn` could technically be returned by `getFacetRangeInputs ` that I suggested above. It's up to you if you would like to have it as a separate function or not.

test/e2e/specs/all-features/recordset/saved-query.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+  firstSavedQueryName: 'main with int_col ( 11 to 22)',
+  secondSavedQueryName: 'main with int_col ( 11 to 22); one',
+  key: {
+    name: 'id',
+    operator: '=',
+    value: '1'
+  }
+};
+
+test.describe('View recordset page and form a query,', () => {
```
RFSH, line 31: This is something that I always assumed that we need, but we don't have to have one outer `test.describe` on our tests. We can have multiple or none at all.

In this case given that the share cite test is opening a completely different page, I think it's better if you move it outside the `describe` like I think you originally planned to do. 

```ts
test.describe('saved query feature', () => {
  ...
})

test('citation on the share and cite modal', ...)
```

test/e2e/specs/all-features/recordset/saved-query.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      // close the dropdown
+      await sqDropdown.click();
+    });
+
+    await test.step('apply filter in int_col facet and open save query form', async () => {
+      const facetIdx = 1;
+      const facet = RecordsetLocators.getFacetById(page, facetIdx);
+
+      // inputs
+      const minInput = RecordsetLocators.getRangeInput(facet, testParams.minInputClass);
```
RFSH, line 94: I think if you do the change that I mentioned in the locators, it also solves the redundancy that we have here. That way you're just calling one function and then you can use its props.

test/e2e/specs/all-features/recordset/saved-query.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+
+      const createSavedQueryModal = ModalLocators.getCreateSavedQueryModal(page);
+      await expect.soft(ModalLocators.saveQuerySubmit(createSavedQueryModal)).toBeVisible();
+    });
+
+    await test.step('have correct modal title and default values then save the query', async () => {
+      const createSavedQueryModal = ModalLocators.getCreateSavedQueryModal(page);
+      await expect.soft(ModalLocators.getModalTitle(createSavedQueryModal)).toHaveText('Save current search criteria for table main');
+
+      await expect.soft(RecordeditLocators.getInputForAColumn(page, 'name', 1)).toHaveAttribute('value', testParams.firstSavedQueryName);
```
RFSH, line 135: Use `toHaveValue` instead of `toHaveAttribute('value')`.

test/e2e/specs/all-features/recordset/saved-query.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+
+      const textAreaVal = 'main with:\n  - int_col (1 choice): int_col ( 11 to 22);';
+      await expect.soft(RecordeditLocators.getTextAreaForAColumn(page, 'description', 1)).toHaveText(textAreaVal);
```
RFSH, line 138: Use `toHaveValue` instead of `toHaveText` as it's more appropriate.

test/e2e/specs/all-features/recordset/saved-query.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      await expect.soft(RecordsetLocators.getFacetOptions(facet)).toHaveCount(6);
+      await RecordsetLocators.getFacetOption(facet, optionIdx).click();
+
+      await RecordsetLocators.getSavedQueryDropdown(page).click();
+      await RecordsetLocators.getSaveQueryOption(page).click();
+
+      const createSavedQueryModal = ModalLocators.getCreateSavedQueryModal(page);
+      await expect.soft(ModalLocators.saveQuerySubmit(createSavedQueryModal)).toBeVisible();
+
+      await expect.soft(RecordeditLocators.getInputForAColumn(page, 'name', 1)).toHaveAttribute('value', testParams.secondSavedQueryName);
```
RFSH, line 184: Use `toHaveValue`.

RFSH (changes requested):
I asked for some changes to the locators. I think implementing the improvement I mentioned would make both the locator and the spec nicer and easier to read.

It seems like I forgot to add `toHaveValue` to `e2e-test-writing.md#assertions`. Can you please add that? Please also document `toContainText `.

test/e2e/locators/modal.ts
```diff
@@ -101,4 +113,18 @@ export default class ModalLocators {
     return modal.locator('.bibtex-download-btn');
   }
 
+  // --------- recordset select related functions ------------ //
+
+  static getModalRows(modal: Locator): Locator {
```
RFSH, line 118: I just realized that you added this. Let's not do this and instead use the `RecordsetLocators.getRows` with proper container passed to it like [here](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/utils/record-utils.ts#L581)

PR #2453, opened 2024-05-03
Migrate input-iframe spec to playwright + fix small chaise issues

RFSH:
This PR will migrate the last remaining `allfeatures` protractor test to playwright.

As part of this I also fixed the following small chaise issues:

- Bulk delete is now going to be disabled if there aren't any rows that can be deleted.
- We were supposed to use the `#333333` color for our text, but bootstrap's default color (`#212529`) was getting used instead. In here I made sure we're using `#333333` by adding the color rule to `.chaise-body` as well as changing default boostrap CSS variables to use `#333333`.

P.S. I don’t know why they haven’t properly documented this, but it turns out there’s a `locator.clear()` function that you can use to clear the inputs. Based on my local testing, it seems to be consistent. So, I removed the `clearInput` function I initially added and replaced all the instances with this as part of this PR.


test/e2e/specs/default-config/multi-form-input/multi-form-input.config.ts
```diff
@@ -3,9 +3,9 @@ import getConfig from '@isrd-isi-edu/chaise/test/e2e/setup/playwright.configurat
 export default getConfig({
   testName: 'default-config/multi-form-input',
   testMatch: [
-    'multi-form-input-clone.spec.ts',
+    // 'multi-form-input-clone.spec.ts',
     'multi-form-input-create.spec.ts',
-    'multi-form-input-edit.spec.ts'
+    // 'multi-form-input-edit.spec.ts'
```
jrchudy, line 8: These 2 specs should be uncommented. Probably from when I noticed the permission issue in hatrac?

jrchudy (approved):

test/e2e/specs/default-config/multi-form-input/multi-form-input.config.ts
RFSH: Thank you for catching it. No I was debugging the `.clear()` changes that I did and forgot to revert it back before pushing.

PR #2454, opened 2024-05-03
Make sure defaultTable chaise-config is properly supported

RFSH:
As the title suggests, this PR will make sure we're properly using the `defaulTable` when users navigate to pages like `/chaise/recordset/#1`. The actual change in this PR is fixing the implementation of `getCatalogId` so that it does not rely on the `chaiseURItoErmrestURI` and finds the ID by itself.

Assume we don't have any `defaultCatalog` defined on `chaise-config.js` and that the following catalog annotation is used on catalog with id `1`:

```
"tag:isrd.isi.edu,2019:chaise-config": {
  "defaultTable": {
    "schema": "isa",
    "table": "dataset"
  }
}
```

If users navigate to the `/chaise/recordset/#1` location, we expect chaise to use the default table. Instead, they would see an error saying that the default table is not defined.

This is because in `ConfigService.configure` we call `getCatalogId` to find the catalog we have to fetch and use its annotation. But the old implementation of `getCatalogId` was trying to parse the whole URL and since the table name is missing and `defaultTable` is not part of the `chaise-config.js`, it would throw an error.

The new implementation is not going to parse the whole URL, and it will only try to find the catalog ID.

### DIFF
--- a/src/utils/head-injector.ts
+++ b/src/utils/head-injector.ts
@@ -139,7 +139,7 @@ function addCanonicalTag() {
     canonicalTag.setAttribute('rel', 'canonical');
 
     // the hash returned from this function handles the case when '#' is switched with '?'
-    const hash = getURLHashFragment(windowRef.location);
+    const { hash } = getURLHashFragment(windowRef.location);
     const canonicalURL = windowRef.location.origin + windowRef.location.pathname + stripSortAndQueryParams(hash);
     canonicalTag.setAttribute('href', canonicalURL);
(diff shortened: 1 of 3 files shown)


jrchudy (approved):

RFSH:
@jrchudy The changes I originally pushed for this PR broke the "help" app. This app doesn't have any hash fragments and only query parameters. I added a comment on the line that I changed with more information.

P.S. This happened when I last modified the `chaiseURItoErmrestURI` implementation too. Good thing we have a test case for the help app 😄

PR #2455, opened 2024-05-04
migrate foreign-key-dropdown spec

jrchudy:
The changes in this pull request are primarily for migrating the foreign key dropdown spec from protractor to playwright.

Changes for the tests include:
 - changing the name of saved-query config to match the format we've been using in playwright 
 - fk dropdown spec specifics
   - changed the visible columns in recordedit and record to both be 3 columns to reuse `testSubmission` function
     - the recordedit form only tests 3 inputs with the id showing but disabled, record app was showing those columns and some of the system columns that weren't tested before
   - bug with how foreign key dropdowns were opening which was causing them to open up and be "behind" the top container and making the option to be selected unclickable
     - I think this passed in the old test because we were using javascript click instead of moving to the element and doing a browser click
     - In recordedit app, we are adding a class to the form container so that the dropdown can show fully below the form when scrolling
     - set `flip={false}` on the component to disable this popper feature
     - I can spend more time on monday trying to understand why this case with only 3 inputs would have 1 dropdown open down properly and the other 2 would open up without enough space

The other changes in this PR are to add more special classes (I'll document these classes in ermrestJS when this PR is merged):
 - aligning content (deriva-webapps labels for user controls) 
 - other classes for use in markdown containers to align some custom content above iframes (atlas-d2k use case hongsuda asked about)


src/components/input-switch/foreignkey-dropdown-field.tsx
```diff
@@ -509,7 +509,7 @@ const ForeignkeyDropdownField = (props: ForeignkeyDropdownFieldProps): JSX.Eleme
                 </button>
               </div>}
             </Dropdown.Toggle>
-            {!props.disableInput && dropdownReference && <Dropdown.Menu className='responsive-dropdown-menu' ref={dropdownMenuRef}>
+            {!props.disableInput && dropdownReference && <Dropdown.Menu flip={false} className='responsive-dropdown-menu' ref={dropdownMenuRef}>
```
RFSH, line 512: Can you please add an inline comment to explain why you've added the flap? you could just explain briefly and then add a link to this PR.

src/assets/scss/app.scss
```diff
@@ -1154,6 +1154,28 @@ html {
     --bs-table-active-color: var(--bs-body-color);
     --bs-table-hover-color: var(--bs-body-color);
   }
+
+  /****** classes referenced by webapps ******/
+  .chaise-content-left {
```
RFSH, line 1159: This is something that we should discuss in the meeting, but do we want these classes to be only available under markdown content? are we planning to use these in the internal implementation? or maybe we want to encourage static sites to use these?

src/assets/scss/_markdown-container.scss
```diff
@@ -101,6 +101,18 @@
     margin: 2px;
   }
 
+  // in atlas-d2k, some iframe content has a <h3> then <p> tag before the iframe
```
RFSH, line 104: I think the names that we're choosing all should have the `chaise-` namespace. Also let's not talk about specific deployments in our code.

We should also list these classes and then find appropriate names for all of them.

RFSH: This relates to the other comment I made about discussing where we want these classes to be defined.

While defining these in two places gives us flexibility, it might confuse data modelers. It would be harder for them to determine if a class is available throughout chaise or just markdown.

test/e2e/specs/default-config/recordedit/foreign-key-dropdown.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  schema_name: 'fk-display-mode',
+  table_displayname: 'fkey-dropdown',
+  fk1_name: 'Y8Fm0o1t3FcHt0S8UjXs6A',
+  fk2_name: 'jZz6GY0Bq-0EXpzflh6zZg',
+  fk3_name: '-wCXD7GyYPDYVhTgm9in3A',
+  column_names: ['fk1 dropdown', 'fk2 dropdown', 'fk3 dropdown'],
+  column_values: [['two', 'five', 'six']]
+};
+
+test.describe(`Create a record, For table ${testParams.table_name}`, () => {
```
RFSH, line 21: This `test.describe` is not needed and it's just making the report lengthier. You should just remove it. So then this file will just have the `test` that you have.

You could even rewrite the `test` description to something like `foreignkey dropdown support` or something like that.

src/assets/scss/_markdown-container.scss
```diff
@@ -101,6 +101,18 @@
...
+  // the following 2 classes can be used to reduce the space between the content
+  // <h3> has margin-bottom: 16px; by default
+  .chaise-reduce-margin-after-header {
```
jrchudy, line 107: change to `.chaise-reduce-header-margin`

PR #2459, opened 2024-05-08
2326 customize order of facets

ladukaraniket:
# Added support to rearrange facets

JnDUcum1D1

# Know issue 
- `react-beautiful-dnd` does not support nested scroll containers yet. This does not affect the functionality for us but throws a warning in dev mode. We are currently suppressing the warning using a window property like so :<br>
 `(window as any)['__react-beautiful-dnd-disable-dev-warnings'] = true;`

Code_1S4yDMZGWp


RFSH:
While the reordering works on UI, it's not changing the order that our flow-control uses. Fixing this would require more changes, so I'm going to close this PR for now.

Issue #2066, opened 2021-03-17
Page appears to jump after paging 


jrchudy:
From chaise channel in slack:
> I'm not sure if this is a bug or a browser dependent issue but when I click "Next" to go to the next 25 records on a nested table, after the paging, the page focus jumps to a different location. 

> For example, if I go to this dataset https://dev.facebase.org/chaise/record/#1/isa:dataset/RID=1-43F6 and scroll down to the Experiments section, after clicking "Next" to page to the next 25 records, the focus of the page jumps to the bottom of the page or to the next nested table. I've noticed different responses in different browsers

I looked into this and it seems to be behaving as expected. I tested against the Experiment related table and found that as I was paging forward in the set, I could see that the content in the table would sometimes grow and sometimes shrink. I believe this is because the content in the “Biosample(s)” column can vary since it is a set. This causes the display of “25 rows” to vary from page to page.

From discussion, the preferred solution would be to scroll to the top of that specific related table when a paging action occurs.

jrchudy:
This can be seen on this page in atlas-d2k:
https://dev.atlas-d2k.org/chaise/record/#2/Common:Collection/RID=16-X2BE (should also be on staging)

PR #2462, opened 2024-05-13
Scroll bug fix for table pagination button click

KenyShah24:
Changes in this PR are made to add scrolling behavior when a user clicks next/previous for a related table in record app. Summary of changes made:
 - pass a reference to the main container from record app to each related table for scrolling to the top of the related table after the content has finished loading
 - change determine scroll element to only find the element and not the related table model
 - add a `useEffect` hook to `RelatedTable` component that watches for when a related table has finished loading it's contents to decide if scrolling should occur
 - modified `recordset provider` to have a property called `pagingSuccess` that is used to signal when a paging operation has completed to help decide when to scroll to the related table 
 - calculate if the related table header is above the main container to see if it needs to be scrolled into view

Other changes unrelated to the attached issue:
 - for table of contents, show a `+` after the number is there is a `previous` page of results

#### Other thoughts:
 - Should a similar change be applied to `recordset` component in other contexts? 
   - As in, should the value I added called `pagingSuccess` be used in these other contexts to trigger the scroll, or should those continue to rely on `isLoading`?
     - `isLoading` seems to be fine in those contexts because there is only 1 recordset table rendered, so when it finishes loading, it is safe to scroll
     - for related tables, if we only rely on `isLoading`, we might get scroll events for other actions on the page:
       - actions for "this" related table include next/previous, page size change, sorting
       - a different related table had an add/delete, link/unlink, or edit occur
         - would trigger ALL related tables to scroll without some change to the `recordset` provider that is directly associated with "this" related table
   - Currently recordset component on recordset page and in popup modals will scroll to top after "any" request is made that updates the rows being displayed (paging, sort, page size, search, and faceting). 

Resolves: #2066


src/providers/recordset.tsx
```diff
@@ -300,6 +302,7 @@ export default function RecordsetProvider({
   const [isLoading, setIsLoading, isLoadingRef] = useStateRef<boolean>(!initialPage);
   const [pageLimit, setPageLimit, pageLimitRef] = useStateRef(typeof initialPageLimit === 'number' ? initialPageLimit : RECORDSET_DEFAULT_PAGE_SIZE);
   const [page, setPage, pageRef] = useStateRef<any>(initialPage ? initialPage : null);
+  const [pagingSuccess, setPagingSuccess] = useState<boolean>(false);
```
RFSH, line 305: Is there a reason why this is not implemented using custom events? Similar to what we've done for "row edit", "row delete", etc? You can see how `src/components/record.tsx` is already acting upon the existing custom events. We should just add another custom event for clicking on next/prev.

I think using events is much better than adding this state variable. The variable is set to true when users click on next/previous and then is set to false when we're doing the scroll. So this is more or less is implemented as an event handler already just without the actual event.

src/utils/record-utils.ts
```diff
@@ -360,3 +361,24 @@ export function getPrefillCookieObject(ref: any, mainTuple: any): {
     keys: keys
   };
 }
+
+export function determineScrollElement(displayname: string): Element | false {
```
RFSH, line 365: Please add a comment for this function and summarize its purpose. Its name is very general but it seems like it's specifically implemented for finding the section/related column element for scrolling.

jrchudy:
@RFSH I made the changes based on your suggestion. Firing the event after `success = true` is set by update would cause scrolling to occur before the related table content was loaded sometimes (a race condition). If this content loaded after the scrolling started to occur, the scrolling would abruptly stop and the user would be "lost" on the page still.

I decided to have a local state variable in `recordset-table.tsx` that signals a paging action occurred, then when `isLoading` is false for that instance of recordset, we check if it was a paging action and fire the custom event if it was.

Comments in the `recordset-table.tsx` point out this need.

src/components/recordset/recordset-table.tsx
```diff
@@ -46,13 +48,14 @@ const RecordsetTable = ({
...
 
-  const tableContainer = useRef<any>(null);
+  const tableContainer = useRef<HTMLDivElement>(null);
 
   const [currSortColumn, setCurrSortColumn] = useState<SortColumn | null>(
     Array.isArray(initialSortObject) ? initialSortObject[0] : null
   );
 
   const [showAllRows, setShowAllRows] = useState(!(config.maxDisplayedRows && config.maxDisplayedRows > 0));
+  const [pagingSuccess, setPagingSuccess] = useState<boolean>(false);
```
RFSH, line 58: Please add a comment explaining why this has been added

RFSH (approved):

PR #2464, opened 2024-05-16
Migrate errors spec to playwright

RFSH:
This PR will migrate the errors spec to playwright.


RFSH:
@jrchudy As you can see the initial commit failed because it couldn't find the error modal on load. Since I'm not doing anything special I don't think this was failing due to flaky tests. So I assumed it's because of the default timeout (the page took longer than expected to load). That's why I ended up increasing the timeout. I reran the CI tests multiple times and the Playwright tests passed in all of them.

P.S. The timeout for each expect/wait is just 5 seconds by default which I increased to 15 seconds (similar to what we did for Protractor). Also the default timeout for each individual `test` is 30 seconds which I increased to 60 seconds. We can adjust these later if needed.

jrchudy (approved):

PR #2466, opened 2024-05-21
Migrate ind-facet spec to playwright

jrchudy:
This PR is ready for review. These changes went through a few iterations that included originally writing all of the "tests" in one `*.spec.ts` file and running them ALL in sequence. This was split into 3 different files:
 - `facet-presentation.spec.ts` - tests the page loading and initializing how we expect it to
 - `ind-facet.spec.ts` - tests each individual facet for every type of facet that we allow
   - page starts with some selections made already that are cleared and facets are closed
 - `four-facet-selections.spec.ts` - 2 tests that do 4 facet selections that were at the end of the original `ind-facet-spec.js`
   - page starts with some selections made already that are cleared and facets are closed
   
After this, I went through and refactored functions out of each spec and put them in `recordset-utils.ts`. These are meant to do the same thing for every test case so code isn't repeated (or different for some tests when it shouldn't be).

#### Notes
 - using `check()` instead of `click()` when making a selection on a checkbox
 - playwright runs through the tests very fast, so waiting for the right elements to be visible (or not visible) is more necessary now
 - for inputs, we should be calling `fill()` followed by checking the right value was entered
   - this goes for clearing inputs as well
   - added `fillRangeInput` and `clearRangeInput` to manage this

#### Failing tests
3 different test cases were failing randomly after my initial implementation, 2 locally and 1 in github actions. The tests that were failing locally were the following. It seems that I was able to fix them and can't get either to fail after 10 runs of `make testdeleteprohibited` locally:
 - `'select a value to filter on and update the search criteria'` in `ind-facet.spec.ts`
   - selecting the facet option would sometimes fail to `check()` the element 
   - this was fixed by clearing the filters before closing the facets when the page loads
   - closing the facets first seemed to cause the facet state to not update until it was opened again, changing this order allowed the state to update before it was closed
 - `'should open facets, click an option in each, and verify the data after'` in `four-facet-selections.spec.ts`
   - selecting a facet option would sometimes fail for the 2nd, 3rd, or 4th selection
   - making sure facet spinner is no longer visible seems to have fixed this
   - this ensures the javascript has finished executing for open facets before an action is taken

The one failing in github actions:
 - `'should have 2 filters selected'` in `facet-presentation.spec.ts`
   - the page would load and every check would succeed until this check where only 1 filter would be shown
     - added "screenshot on failure" to debug this
     - noticed the id filter would show but the `int_col` facet would not have a selection or a filter shown
   - can't understand why this is happening in github actions
   - this can't be reproduced locally even when running `make testdeleteprohibited`
      c07a26565871c7d2c2fdfefd531dc61e2466e80d


test/e2e/specs/delete-prohibited/recordset/facet-presentation.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+    expect.soft(testCellDimensions?.height).toBeGreaterThan(cellHeight);
+  });
+
+  await test.step('should have ' + testParams.totalNumFacets + ' facets', async () => {
+    await expect.soft(RecordsetLocators.getAllFacets(page)).toHaveCount(testParams.totalNumFacets);
+
+    const titles = RecordsetLocators.getFacetTitles(page);
+    const count = await titles.count();
+    for (let i = 0; i < count; i++) {
+      await expect.soft(titles.nth(i)).toHaveText(testParams.facetNames[i])
```
RFSH, line 84: `toHaveText` handles array of elements. So just the following works too:

```tsx
await expect.soft(titles).toHaveText(testParams.facetNames);
```

In other places that I had to do something similar (like [here](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/all-features/record/related-table.spec.ts#L75,L76)) I usually test the count separately just so it's easier to debug.

RFSH, line 84: toHaveText handles array of elements. So just the following works too:
```tsx
await expect.soft(titles).toHaveText(testParams.facetNames);
```

In other places that I had to do something similar (like [here](https://github.com/informatics-isi-edu/chaise/blob/master/test/e2e/specs/all-features/record/related-table.spec.ts#L75,L76)) I usually test the count separately just so it's easier to debug.

test/e2e/specs/delete-prohibited/recordset/facet-presentation.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+    }
+  });
+
+  await test.step('should have 3 facets open', async () => {
+    await expect.soft(RecordsetLocators.getOpenFacets(page)).toHaveCount(testParams.defaults.openFacetNames.length);
+
+    const titles = RecordsetLocators.getOpenFacetTitles(page);
+    const count = await titles.count();
+    for (let i = 0; i < count; i++) {
+      await expect.soft(titles.nth(i)).toHaveText(testParams.defaults.openFacetNames[i])
```
RFSH, line 94: Same comment about `toHaveText`

test/e2e/specs/delete-prohibited/recordset/facet-presentation.spec.ts
RFSH: While I think what you have is fine, we should try to avoid using locator functions like `count()` if we can. That's what the official document says as well

https://playwright.dev/docs/api/class-locator#locator-count
>If you need to assert the number of elements on the page, prefer [expect(locator).toHaveCount()](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-count) to avoid flakiness. See [assertions guide](https://playwright.dev/docs/test-assertions) for more details.


P.S. I'm not saying that we shouldn't use `.count()` at all. I cases like the `testRangeInputSubmitThenClear` function I think using `count()` makes sense, as it will allow us to avoid hardcoding the number. But in cases where we already know the count, or if we just want to eventually use `toHaveText` we should try to avoid these flaky functions.

test/e2e/specs/delete-prohibited/recordset/facet-presentation.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+    await searchSubmitBtn.click();
+
+    const modalOptions = RecordsetLocators.getCheckboxInputs(facetPopup);
+    // make sure search result is displayed
+    await expect.soft(modalOptions).toHaveCount(13);
+
+    // make sure the first row is selected
+    await expect(modalOptions.nth(0)).toBeChecked();
+
+    const optionsCount = await modalOptions.count();
```
RFSH, line 142: You've already hardcoded the number of options in line 137, so let's also use the hardcoded value instead of calling `.count()`

test/e2e/specs/delete-prohibited/recordset/four-facet-selections.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    `${baseURL}/recordset/#${getCatalogID(testInfo.project.name)}/${testParams.schema_name}:${testParams.table_name}${testParams.sort}`,
+    testParams.totalNumFacets,
+    [0, 1, 11],
+    testParams.defaults.pageSize
+  );
+
+  await test.step('should open facets, click an option in each, and verify the data after', async () => {
+    const numFacets = testParams.multipleFacets.length;
+
+    const facet1 = RecordsetLocators.getFacetById(page, testParams.multipleFacets[0].facetIdx);
```
RFSH, line 71: Is there a reason why you're not using `for` loop here? I don't think you need to rewrite this and I'm just curious.

test/e2e/specs/delete-prohibited/recordset/facet-presentation.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  await test.step('"int_col" facet should not show the histogram with 1 facet options checked', async () => {
+    // use 1 index
+    const facet = RecordsetLocators.getFacetById(page, 1);
+    await expect.soft(RecordsetLocators.getFacetSpinner(facet)).not.toBeVisible();
+    await expect.soft(RecordsetLocators.getFacetHistogram(facet)).not.toBeVisible();
+    await expect.soft(RecordsetLocators.getCheckedFacetOptions(facet)).toHaveCount(1);
+  });
+
+  await test.step('should have 2 filters selected', async () => {
```
RFSH, line 114: Let's add a condition to only run this step locally. I'm not sure why this is failing on CI either.

test/e2e/utils/recordset-utils.ts
```diff
@@ -35,3 +38,200 @@ export async function testRecordsetTableRowValues(container: Page | Locator, exp
...
+ * @param filterOptions array of filter option labels
+ */
+export async function openFacetAndTestFilterOptions(page: Page, facet: Locator, facetIdx: number, filterOptions: string[], numOpenFacets: number) {
+  await openFacet(page, facet, facetIdx, filterOptions.length, numOpenFacets);
+
+  const facetOptions = RecordsetLocators.getFacetOptions(facet);
+  const numFacets = await facetOptions.count();
+
+  for (let i = 0; i < numFacets; i++) {
+    await expect.soft(facetOptions.nth(i)).toHaveText(filterOptions[i]);
```
RFSH, line 70: Same comment about `toHaveText`

RFSH (approved):
Overall I like the separations that you did. It makes the facet spec easier to follow.

I made some comments about how we might want to avoid using `.count()` if we can, and how `toHaveText()` already handles array of elements for us.

And related to the flaky test on CI, let's add a condition to skip it on CI. Please also add a TODO so we know that we keep track of this (in case something similar comes up).

test/e2e/specs/delete-prohibited/recordset/four-facet-selections.spec.ts
jrchudy: I think it was this way in protractor because they were opened in the reverse order so the facet headers didn't "move" as each was opened. Then we tested them in the same order as the previous test.

Since it is reverse order for opening and then normal order for testing, I guess I didn't want to have 2 for loops? But that shouldn't matter anymore and I can change it.

PR #2472, opened 2024-05-25
Migrate composite-key and domain-filter specs to playwright

RFSH:
This PR migrates composite-key and domain-filter specs to playwright.

Both of these specs were very trivial. The only noticeable change that I had to do was introducing a new `testRecordMainSectionPartialValues` function. In some recordedit test cases we just want to make sure the value that we set on the recordedit page is properly saved, and we don't care about other automatically generated columns. This function allows us to do that. We had a similar function in Protractor for this.


jrchudy (approved):
I was thinking `openRecordediPage` might make sense to be factored out since it will be reused many times for other specs. I don't know if that makes sense though since it is only a few lines.

PR #2484, opened 2024-06-27
Misc facet playwright

jrchudy:
In this PR, the `misc-facet` spec is migrated to playwright. With this, all of the delete prohibited specs are also migrated so that protractor configuration is removed as well. I also removed the command in the makefile and updated `e2e.yml`.

Other changes in this PR:
 - add chance declaration for generating random content in an input field
 - moved a check in histogram tests to validate button is disabled before checking the input values
   - button is no longer disabled when we know the input values have updated after zooming in
 - classname of a spinner in recordset changed


test/e2e/specs/delete-prohibited/recordset/misc-facet.spec.ts
```diff
@@ -0,0 +1,1126 @@
...
+    });
+  });
+
+  test('navigating to recordset with custom facet', async ({ page, baseURL }, testInfo) => {
+    const params = testParams.customFacet;
+
+    await test.step('should load recordset page', async () => {
+      const PAGE_URL = `/recordset/#${getCatalogID(testInfo.project.name)}/${testParams.schema_name}:${testParams.table_name}`;
+
+      await page.goto(`${baseURL}${PAGE_URL}/*::facets::${params.cfacetBlob}`);
```
RFSH, line 1059: You should use `*::cfacets::` not `*::facets::`.

test/e2e/utils/recordset-utils.ts
```diff
@@ -87,12 +90,40 @@ export async function testSelectFacetOption(page: Page, facet: Locator, optionId
...
   const clearAll = RecordsetLocators.getClearAllFilters(page);
   await clearAll.click();
   await expect.soft(clearAll).not.toBeVisible();
-  await expect.soft(RecordsetLocators.getFacetOption(facet, optionIdx)).not.toBeChecked();
   await expect.soft(RecordsetLocators.getRows(page)).toHaveCount(pageSize);
+
+  if (facet && optionIdx) await expect.soft(RecordsetLocators.getFacetOption(facet, optionIdx)).not.toBeChecked();
+}
+
+export async function testShowMoreClick(page: Page, facet: Locator, modal: Locator, numRows: number, numCheckedRows: number) {
```
RFSH, line 102: Can you please add jsdoc comment to these functions that you've added?

For example for this function you should mention that it will open the show more modal and leave it open.

RFSH (approved):
I like the changes that you've done and love how fast the playwright facet spec is!

I just have some minor suggestions before pushing the changes:

- While I didn't see any issues related to this, but to make sure it doesn't cause issues at some point in the future, you should make sure the modal is closed after clicking on the close button. So basically when you're doing `getCloseBtn().click`, the next line should be: `await expect.soft(modal).not.toBeAttached();`. We could actually have a helper function for this to. The function would do the click and wait for the modal to disappear. It's up to you if a function is needed or not.

- Like I mentioned in my inline comment, I think your new functions need jsdoc so it's easier to know what they're doing.

- I think we should rename `ind-facet.config.ts` to something like `main-facet.config.ts` or just `facet.config.ts`. It's a bit confusing that it's using the same name as one of the specs, but it includes more than just that one spec.

Issue #2416, opened 2024-03-26
Improve multi-create workflow for almost pure and binary relationships


jrchudy:
This issue focuses on tables that are almost pure and binary associations. These tables have 2 foreign keys and some static columns. A [good example](https://www.atlas-d2k.org/chaise/recordset/#2/Protocol:Protocol_Author@sort(Protocol,Rank,RID)) of this is the `Protocol_Author` table in atlas-d2k. We want to be able to create rows using the modal interface but also need a way to allow for setting values for these associated rows after selections have been made in the modal.

To improve this workflow, we are changing what the `Add records` button does on record page for inbound foreign keys when certain criteria are met. If the table is "almost a pure and binary association" and has other columns that are not foreign keys, we will trigger the following alternate workflow. Currently for adding protocol author associations, add records opens `recordedit` app with only 1 row and the "author" value is empty.

### Workflow change:
We discussed changing the workflow to the following for adding rows to an association like table when the pair of foreign keys are a unique relation (1 table entry for X main entity and Y leaf entity pair) and non-unique (N table entries for X main entity and Y leaf entity pair):
 1. Start on record app like above and click "Add Records"
     <img width="1393" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/1842fba3-37e5-44b4-aee8-334b6e45f71e">
 2. Recordedit is opened in a new tab with an recordset modal picker shown on top of the recordedit inputs on page load
     <img width="889" alt="image" src="https://github.com/user-attachments/assets/f8907d05-39fd-44db-bce7-e769124ddbc9">
 3. After selections are made, click "Continue" to use those selections to fill in recordedit forms
     <img width="1728" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/93c8ed26-291b-48d0-a1fd-647fa79aeb54">
 4. More forms can be added using "Add more" which will show the recordset modal picker again
   a. in the case that the pair of foreign keys are a part of a unique key, this modal will show selected rows from step 2 as "greyed" out since they can't be chosen again
      <img width="1706" alt="image" src="https://github.com/user-attachments/assets/d52a5cf8-ec26-49b6-9675-e73428e65ea3">
 5. Click "Save" to add the new selections to the form
      <img width="1728" alt="image" src="https://github.com/informatics-isi-edu/chaise/assets/2932600/047b2f1e-8e4d-4851-b868-98d9131c91fe">
[...]

PR #2490, opened 2024-07-16
Recordedit with association picker when adding rows using prefill

jrchudy:
This pull request will change how add records on record app works for adding rows to certain related tables (outlined in issue #2416 ). This is ready to review so we can discuss which functions to add to ermrestJS to make support for this simpler in chaise.

### The changes to support this new workflow are as follows:
 - recordedit component changed to check if there is a prefill and if we have the leaf column for the association visible in create mode of recordedit before showing the association modal on load
   - on submission of association modal, the first form is updated with the row to associate and more forms are added if there were multiple selections
   - "add more" button added to reopen the association modal to selected more rows to associate and add more forms in recordedit on submission
 - recordedit provider keeps track of the prefill object from cookie storage and the selected rows when the association is unique
 - attach `foreignKeyCallbacks` when generating the foreign key fields if the column for the foreign key field is the one used for the leaf column in the association
 - `foreignkey-field` changed to update selected rows used for disabling when each input is changed or cleared
 - `removeForm` updated to remove the selected row(s) associated with the removed form

#### Other changes:
 - `foreignkey-dropdown-field` also works with disabling rows and updating which rows are selected
   - if the selected rows in the form are changed in any other dropdown or more rows are added using "add more", the foreign key dropdowns will have their disabled rows updated
 - hide multi set button for the leaf column foreign key field
 - new `RecordsetDisplayMode` added for customizing bulk foreign key picker title and "continue" button
 - disabled tuples promise moved to `recordedit-utils` so the generated promise function can be reused
 - `chaise-input-group-text-sm` rule was not using `btn-padding-sm-*` variables
 - `addTooManyFormsAlert` added for when trying to pick more than 200 foreign key values to add as recordedit forms
   
### Testing:
Since this applies to both foreign key picker popups and dropdowns, this should be tested for both cases. The ermrestJS branch `association-display` should be checked and installed alongside this branch.

~~To test the modal picker, navigate to [this Protocol](https://dev.atlas-d2k.org/~jchudy/chaise/record/#2/Protocol:Protocol/RID=17-DSQ0) and click "Add Records" for the "Protocol Author" related table~~ dev.atlas-d2k is unavailable for testing, see comment below for testing modal picker.
 
To test the foreignkey dropdown, [this catalog](https://dev.derivacloud.org/chaise/record/#84426/edit_w_association:main/RID=1Y0) is using the foreign key dropdown for the recordedit inputs. If you can't test with this one, this was created using `createAssociationEdit` script in the `testScripts` folder of `ErmrestDataUtils`. This schema could also be used to test the modal popup too if the `table-display` annotation on the table named "leaf" is modified.


src/providers/recordedit.tsx
```diff
@@ -401,6 +429,29 @@ export default function RecordeditProvider({
           updateHeadTitle('Create new ' + reference.displayname.value);
         }
 
+        if (prefillObj) {
+          setPrefillObject(prefillObj);
+
+          // TODO: if reference.table is an association
+          // if (prefillObject.isAssociation) {
```
jrchudy, line 436: Need to add API in ermrestJS.

src/providers/recordedit.tsx
```diff
@@ -276,10 +296,18 @@ export default function RecordeditProvider({
     if (!reference || setupStarted.current) return;
     setupStarted.current = true;
 
+    const prefillObj = getPrefillObject(queryParams);
+
+    // add properties to the prefillObject that is stored here in the provider for use in other components
+    if (prefillObj) {
+      // TODO: isAssociation and associationIsUnique
+      // prefillObj.isAssociation = reference.table.isAssociation || false;
+      prefillObj.hasUniqueAssociation = reference.table.isAssociation || false;
```
jrchudy, line 305: Need to add APIs in ermrestJS.

src/assets/scss/_recordset-table.scss
```diff
@@ -21,37 +21,37 @@
     display: table-row-group;
     vertical-align: middle;
 
-    > tr > td.disabled-cell {
-      background-color: map-get(variables.$color-map, 'table-disabled-background');
-    }
+    // > tr > td.disabled-cell {
```
jrchudy, line 24: Should these rules be removed?

src/components/recordedit/recordedit.tsx
```diff
@@ -292,20 +308,112 @@ const RecordeditInner = ({
...
+    if (prefillObject) {
+      // trigger the association modal when there is an assoication and
+      // we know the leaf column for the association is visible in create mode
+      if (prefillAssociationFkLeafColumn?.reference) {
+        // check if leaf column is defined and set the reference to use if it is
+        const domainRef: any = prefillAssociationFkLeafColumn?.reference;
+
+        const andFilters: any[] = [];
+        // loop through all columns that make up the key information for the association with the leaf table and create non-null filters
+        prefillAssociationFkLeafColumn.foreignKey.key.colset.columns.forEach((col: any) => {
```
jrchudy, line 325: Maybe this should be a function of `ForeignKeyPseudocolumn` in ermrestJS

src/utils/recordedit-utils.ts
```diff
@@ -891,3 +884,106 @@ export function validateForeignkeyValue(
...
+    logStack: any,
+    logStackPath: string,
+    requestCauses?: any,
+    reloadStartTime?: any
+  ): Promise<{ page: any, disabledRows?: any }> => {
+    return new Promise((resolve, reject) => {
+      const disabledRows: any = [];
+
+      const disabledRowsFilters: any[] = [];
+      Object.keys(prefillObject.keys).forEach((key: string) => {
```
RFSH, line 926: This should be probably be moved to ermrestjs

src/utils/recordedit-utils.ts
```diff
@@ -891,3 +884,106 @@ export function validateForeignkeyValue(
...
+        .setSamePaging(page)
+        .read(pageLimit, logObj, false, true)
+        .then((newPage: any) => {
+          newPage.tuples.forEach((newTuple: any) => {
+            const index = page.tuples.findIndex((tuple: any) => {
+              return tuple.uniqueId === newTuple.uniqueId;
+            });
+
+            if (index > -1) {
+              page.tuples[index].disabledType = 'associated';
```
RFSH, line 962: We should treat the objects that we get from ermrestjs as immutable. Eventually ermrestjs will be converted to TypeScript, and we won't be able to do stuff like this.

I think you should do what we did for `selectedRows` prop for the `disabledRows`. So add a new `DisabledRow` type that has `tuple` and `disabledType` prop.

src/components/input-switch/foreignkey-field.tsx
```diff
@@ -83,9 +81,10 @@ const ForeignkeyField = (props: ForeignkeyFieldProps): JSX.Element => {
   const usedFormNumber = typeof props.formNumber === 'number' ? props.formNumber : 1;
 
   const { setValue, getValues } = useFormContext();
+  const { prefillAssociationSelectedRows } = useRecordedit();
```
RFSH, line 84: This should be prop that you pass to this component. We should try to avoid adding recordedit provider as a dependency of `InputSwitch` and its children. You can add it to the `foreignKeyCallbacks` prop.

src/components/input-switch/foreignkey-field.tsx
```diff
@@ -83,9 +81,10 @@ const ForeignkeyField = (props: ForeignkeyFieldProps): JSX.Element => {
...
 
   const [recordsetModalProps, setRecordsetModalProps] = useState<RecordsetProps | null>(null);
-
+  const [inputSelectedRow, setInputSelectedRow] = useState<SelectedRow | null>(null);
```
RFSH, line 87: Why is this state variable needed? We can figure this out by looking at `prefillAssociationSelectedRows`. right? I feel like looking up an array that won't go beyond 200 elements is better than the unnecessary renders that this is going to add.

src/components/input-switch/foreignkey-dropdown-field.tsx
```diff
@@ -90,6 +91,7 @@ const ForeignkeyDropdownField = (props: ForeignkeyDropdownFieldProps): JSX.Eleme
 
   const { setValue, getValues } = useFormContext();
   const { dispatchError } = useError();
+  const { prefillAssociationSelectedRows } = useRecordedit();
```
RFSH, line 94: Same comment about not using RecordeditProvider here.

src/providers/recordedit.tsx
```diff
@@ -104,6 +106,17 @@ export const RecordeditContext = createContext<{
   lastContiguousChunkRef: any,
   /* max rows allowed to add constant */
   MAX_ROWS_TO_ADD: number,
+  /* the prefill object from cookie storage based on prefill query param */
+  prefillObject: PrefillObject | null,
```
RFSH, line 110: When you move the prefill/association logic to ermrestjs, you most probably don't need any of these props. You most probably can either get them from the `reference` itself. Or similar to how `citation` works in Record provider, we could just have one prop here which would have all the APIs and props that you need in it.

Also I feel like `setPrefillAssociationSelectedRows` and `updateAssociationSelectedRows` are very similar to each other. If you pass `prefillAssociationSelectedRows` to the foreignkey components, you don't need `updateAssociationSelectedRows` function anymore.

RFSH (changes requested):
Functionality-wise, everything looks good and works as expected. 

The only odd thing I noticed is how the `clone` button works. I feel like we shouldn't have both `clone` and `Add more`. Also, do we want to change the title of recordedit in this particular mode? To reinforce the idea that this UI is special and some features will work differently than other recordedit pages.

The comments I left inline are mainly about code, not functionality.

src/assets/scss/_input-switch.scss
```diff
@@ -264,6 +264,19 @@
       .dropdown-item {
         cursor: pointer;
 
+        &.disabled, &:disabled {
+          color: #adb5bd;
```
jrchudy, line 268: Use `map.get` to get this color from the `color-map`

src/components/input-switch/foreignkey-field.tsx
jrchudy: This was left over from when I was updating association selected rows differently. Since it's index based now, you are right this isn't needed and can be a simple lookup each time instead

src/providers/recordedit.tsx
```diff
@@ -401,6 +427,29 @@ export default function RecordeditProvider({
           updateHeadTitle('Create new ' + reference.displayname.value);
         }
 
+        if (prefillObj) {
+          setPrefillObject(prefillObj);
+
+          // TODO: if reference.table is an association
+          // if (prefillObject.isAssociation) {
+          reference.columns.forEach((column: any) => {
```
jrchudy, line 435: Move this to ermrestJS too as part of `reference.prefillAssociation`

jrchudy:
APIs to move to ErmrestJS:
 - Reference.prefillAssociation
   - similar to other APIs that will "compute/construct" the object the first time this is called
   - FK leaf column
   - FK main column
   - isAssociation
     - what does this check mean?
     - at least 2 outbound foreign keys on the association table?
   - isUnique
     - there only 2 keys and they are both part of the key information for the association table
   - `disabledRowsFilter` that uses fk leaf column and fk main column.
 - function to create `andFilters` based on the fk leaf column
   - function of pseudo column instead of `PrefillAssociation` object?

src/providers/recordedit.tsx
jrchudy: I think the calculated `prefillObject` could be moved entirely to something like `reference.prefill.prefillObject` in ermrestJS. I'm going to wait until we discuss the changes and review this chaise PR before making this decision since it will cause more refactoring in the code.

Regarding `updateAssociationSelectedRows`, I didn't want to have to have the contents of this function four times (twice in each of foreign key field and foreign key dropdown field) in case the functionality needed to change.

src/providers/recordset.tsx
```diff
@@ -352,6 +358,19 @@ export default function RecordsetProvider({
           } else {
             removeURLLimitAlert();
           }
+        } else if (config.displayMode === RecordsetDisplayMode.FK_BULK) {
+          // TODO: how to get numberForms?
+          // const numberFormsAllowed = RECORDEDIT_MAX_ROWS - numberForms
```
jrchudy, line 363: I need to think about this a little bit more. 

How should `numberForms` be made available in recordset provider when it is used as a modal in recordedit app?

jrchudy:
@RFSH This is ready for review now with the changes to ermrestJS that are ready for review as well. There are few things that we should discuss together about this PR:
 - How should numberForms be made available in recordset provider when it is used as a modal in recordedit app?
   - I'll update this comment later after I think about this a bit more
   - most likely needs to be a prop for recordset modal, most likely another foreign key callback?
 - should prefill object be kept as a variable in recordedit provider or moved to the `BulkCreateForeignKeyObject` calculated in ermrestJS?
 - I noticed 2 things in the test code that we should think about:
   - our recordedit selectors inconsistently use 0 or 1 index
     - getForeignKeyInputButton - uses 1 indexing 
     - getDeleteRowButton - uses 0 indexing
     - we should do one or the other for all selectors in recordedit app
   - RecordsetRowValue is a type that is an array of “column” values
     - should the “type” be called RecordsetRowValues (plural) since it’s intended to be an array of individual column values?
     - this is often used as a type for arrays so we can have an array of arrays

src/components/recordedit/recordedit.tsx
```diff
@@ -350,56 +447,258 @@ const RecordeditInner = ({
...
+    // NOTE: forms.length "should" be 1 at this point before subtracting 1
+    if (selectionsFillFirstForm) numForms--;
+
+    return (rows.length + numForms < RECORDEDIT_MAX_ROWS);
+  }
+
+  // user closes the modal without making any selections
+  const closeBulkForeignKeyCB = () => {
+    // if the page was loaded with a modal showing and it is dismissed, update app state variable and do nothing else
+    if (selectionsFillFirstForm) setSelectionsFillFirstForm(false);
```
jrchudy, line 593: @RFSH pointed out that there was a bug when:
 1. the recordedit app was loaded
 2. the user closed the first modal
 3. then used "Add more" to add 3 values, resulting in 4 forms displayed
 4. remove form 2
 5. open "Add more" again and the wrong row was still disabled

jrchudy: The issue was that when the first modal is shown and closed, the first form is initialized and I didn't initialize `bulkForeignKeySelectedRows` to have a `null` value for the first "selection". The following change to this condition will initialize that "array" with a `null` value as the first element representing a null value in the first form.

```js
if (selectionsFillFirstForm) {
  setSelectionsFillFirstForm(false);
  setBulkForeignKeySelectedRows([null]);
}
```

test/e2e/specs/all-features/record/related-table.spec.ts
```diff
@@ -831,6 +836,36 @@ test.describe('Related tables', () => {
     });
   });
 
+  /**
+   * The following 2 tests are for testing the prefill functionality when the inbound foreign key is part of a table that is "almost" pure and binary
+   *
+   * This means there are 2 foreign keys that are part of the same key (making the pair unique) and there are other columns that are not foreign keys
+   */
+  test.describe('for a table that is almost pure and binary and the foreign keys are a unique key', async () => {
```
jrchudy, line 844: From the comment above, there was a bug regarding selecting leaf values and removing forms that was fixed.

I need to update the test cases to ALSO check for the values that are disabled in the modals/dropdowns and not just the “# of disabled rows”. 

I will also add a test case for the example where the initial modal is “closed” and make sure the first form remains “unfilled” for the “leaf value” since no selections were made.

src/components/recordedit/recordedit.tsx
```diff
@@ -350,56 +447,258 @@ const RecordeditInner = ({
...
+
+        tempFormValues[`c_${formNumber}-${col.RID}`] = val;
+      });
+    });
+
+    return tempFormValues;
+  }
+
+  // show the prefill bulk foreign key modal if we have a prefill object and bulk foreign key recordset props
+  const openBulkForeignKeyModal = () => {
```
RFSH, line 550: I think you should modify this function to do everything that you're doing in the initial useEffect. And then just call that as well. You're doing the exact same thing as that useEffect here but some stuff are missing (which I don't think is intentional. For example the `domainRef` change that you've done above is missing from here).

src/components/recordedit/recordedit.tsx
```diff
@@ -350,56 +447,258 @@ const RecordeditInner = ({
...
+
+    if (selectionsFillFirstForm) {
+      if (modalSelectedRows.length > 1) {
+        initialValues = createNewForms(tempFormValues, modalSelectedRows.length - 1).tempFormValues;
+      }
+
+      startFormNumber = 1;
+
+      setSelectionsFillFirstForm(false);
+      // } else if (annotation.doClone) {
```
RFSH, line 647: I don't think we're going to allow this anymore, right? So you should just remove the commented code.

RFSH (changes requested):

src/models/recordset.ts
```diff
@@ -66,7 +66,8 @@ export enum RecordsetDisplayMode {
   FK_POPUP_EDIT = 'popup/foreignkey/edit',
   PURE_BINARY_POPUP_ADD = 'popup/purebinary/add',
   PURE_BINARY_POPUP_UNLINK = 'popup/purebinary/unlink',
-  SAVED_QUERY_POPUP = 'popup/savedquery'
+  SAVED_QUERY_POPUP = 'popup/savedquery',
+  FK_BULK = 'popup/foreignkey/bulk'
```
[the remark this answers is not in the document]
jrchudy: I went with `FK_POPUP_BULK_CREATE`

src/models/recordset.ts
```diff
@@ -160,6 +161,16 @@ export type SelectedRow = {
   // cannotBeRemoved?: boolean;
 }
 
+export enum DisabledRowType {
+  ASSOCIATED= 'associated',
```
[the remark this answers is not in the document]
jrchudy: added a comment or this line and the next one

src/utils/message-map.ts
```diff
@@ -69,7 +69,10 @@ export const MESSAGE_MAP = {
     showDetails: 'Click to show more details about the filters',
     saveQuery: 'Click to save the current search criteria',
     export: 'Click to choose an export format.',
-    liveData: 'You are viewing snapshotted data. Click here to return to the live data catalog.'
+    liveData: 'You are viewing snapshotted data. Click here to return to the live data catalog.',
+    // tooltips for disabled rows in recordset single select modal for foreignkey-field and foreignkey-dropdown-field
+    otherInput: 'This row is selected in another input in the form',
```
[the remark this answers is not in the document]
jrchudy: I went with your suggestions for the names since these are only used for "disabled rows"

src/components/recordedit/recordedit.tsx
```diff
@@ -100,16 +107,24 @@ const RecordeditInner = ({
...
   } = useRecordedit()
 
   const [formProviderInitialized, setFormProviderInitialized] = useState<boolean>(false);
   const [addFormsEffect, setAddFormsEffect] = useState<boolean>(false);
 
+  // the next 3 state variables are used when there is a prefill object for starting recordedit with more than one form to associate on creation
+  const [showBulkForeignKeyModal, setShowBulkForeignKeyModal] = useState<boolean>(false);
+  const [bulkForeignKeyPickerProps, setBulkForeignKeyPickerProps] = useState<RecordsetProps | null>(null);
+  // when initializing the page, the selections in the modal that appears first should fill the first form
+  const [selectionsFillFirstForm, setSelectionsFillFirstForm] = useState<boolean>(true);
```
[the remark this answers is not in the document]
jrchudy: Changed to `bulkForeignKeySelectionsFillFirstForm`

src/components/recordedit/recordedit.tsx
```diff
@@ -292,20 +308,101 @@ const RecordeditInner = ({
...
+        deletable: false,
+        sortable: true,
+        selectMode: RecordsetSelectMode.MULTI_SELECT,
+        showFaceting: true,
+        disableFaceting: false,
+        displayMode: RecordsetDisplayMode.FK_BULK,
+      };
+
+      const stackElement = LogService.getStackNode(
+        LogStackTypes.RELATED,
```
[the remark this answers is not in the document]
jrchudy: changed to `LogStackTypes.FOREIGN_KEY`

src/components/recordedit/recordedit.tsx
```diff
@@ -292,20 +308,101 @@ const RecordeditInner = ({
...
+        domainRef.table,
+        { source: domainRef.compressedDataSource, entity: true, picker: 1 }
+      );
+
+      const logInfo = {
+        logObject: null,
+        logStack: [stackElement],
+        // logStack: getRecordLogStack(stackElement),
+        logStackPath: LogService.getStackPath(null, LogStackPaths.ADD_PB_POPUP),
```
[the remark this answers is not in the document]
jrchudy: Changes to `LogStackPaths.ADD_BULK_FOREIGN_KEY`

src/components/recordedit/recordedit.tsx
```diff
@@ -292,20 +308,101 @@ const RecordeditInner = ({
...
+        domainRef.pseudoColumn.filterProps.leafFilterString
+      ) {
+        // NOTE should we display the filters or not?
+        customFacets = {
+          ermrest_path: domainRef.pseudoColumn.filterProps.leafFilterString,
+          removable: false,
+        };
+      }
+
+      const modalReference = domainRef.addFacets(andFilters, customFacets)
```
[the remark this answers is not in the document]
jrchudy: New context added to ermrestJS called `compactSelectBulkForeignKey`. See ermrestJS PR for comments about those changes.

src/components/input-switch/foreignkey-field.tsx
```diff
@@ -147,7 +149,18 @@ const ForeignkeyField = (props: ForeignkeyFieldProps): JSX.Element => {
       getValues
     );
 
+    let currentSelectedRow;
```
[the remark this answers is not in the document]
jrchudy: One issue with doing this is in the case of edit when a value is set for the input without making a selection in the modal. 

We need the `uniqueId` of the "selected row" to ensure it is the row to be selected in the modal. I need to think more about this. In this case, the `uniqueId` needs to be stored somewhere alongside the display value that is set in react hook form, maybe another value in RHF `values` specifically for the `uniqueId`.

If the `uniqueId` is tracked differently, then the implementation should change where bulk foreign key selections should set the display and uniqueId so the row is "set up" already instead of being fetched from `bulkForeignKeySelectedRows`.

src/providers/recordedit.tsx
```diff
@@ -769,11 +823,37 @@ export default function RecordeditProvider({
...
+      tempSelectedRows.forEach((row, index) => {
+        console.log(`row at index ${index}: `, row?.displayname.value);
+      });
+
+      indexes.forEach((index: number) => {
+        // use splice to remove the element from the array and shift all array values after this element forward
+        tempSelectedRows.splice(index, 1);
+      });
+
+      console.log('after: ');
```
RFSH, line 850: Leftover console.log

RFSH (approved):
Overall everything looks and works as expected. My comments are mainly about code style or leftover code.

I left a comment about the "LogStackPath" that you introduced for this new popup and how we should change it to be more accurate. Please also update [logging.md](https://github.com/informatics-isi-edu/chaise/blob/master/docs/user-docs/logging.md#action-definition) to include the new `stack-path`.

PR #2497, opened 2024-08-01
Inject defaultCatalog property to chaise-config and remove catalog alias creation step

RFSH:
# Summary

A while ago, I modified the catalog creation step of our test framework to add aliases to the created catalog. This allowed us to test the `defaultCatalog` property since we could hardcode the alias inside the `chaise-config.js` files.

This worked properly while I was testing it on my own. However, since then, we have found that the alias creation has limitations when multiple users use the same alias or if multiple users want to run test cases simultaneously.

So this PR will remove the alias creation step and instead will inject a line into the `chaise-config.js` like the following:

```
chaiseConfig.defaultCatalog = 'id_of_the_newly_created_catalog';
```

Since we don't need this for all the test cases, I added a new `addDefaultCatalogToChaiseConfig` to our test options.

# Details

The following are the issues that we found with the alias method:

1. Since an alias can be associated with just one catalog at a time, running multiple instances of our test framework simultaneously could result in unpredicted behavior. For example, assume user1 runs the test cases and assigns `my-alias` to a catalog. At the same time, user2 runs the test case and assigns `my-alias` to their own newly created catalog. In this case, using the `my-alias` will refer to one of these catalogs, so one of the users is using the wrong catalog.
2. If we don't provide the `owner` property while creating the alias, whoever creates it first will be its sole owner. So, if someone else attempts to use the same alias for a different catalog, they will get an error. To solve this, we would have to provide the `isrd-testers` as the `owner` of the alias. But that means that we have to hardcode this as part of chaise.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -56,6 +56,7 @@ module.exports = {
     '@typescript-eslint/no-unsafe-function-type': 'warn',
     '@typescript-eslint/ban-ts-comment': 'warn',
     '@typescript-eslint/no-inferrable-types': 'warn',
+    '@typescript-eslint/no-require-imports': 'warn',
 
     // ------------------ testing ------------------
     '@typescript-eslint/no-floating-promises': 'warn'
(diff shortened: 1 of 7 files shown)


jrchudy (approved):
I understand what you mean as "hacky" because you have to read the file into memory, add to it, and then write it back since this is "before" the javascript test environment starts running.

PR #2519, opened 2024-08-14
Record presentation playwright

jrchudy:
Migrate record presentation spec and remove old protractor spec files.

Changes in this PR besides the test migration:
 - Change `resolverImplicitCatalog` to `100` to avoid colliding with a catalog number that GA will use each time the tests run
   - we have 4 parallel configurations that will use catalogs 1-4, this spec was run in the 1st catalog in protractor, so using `resolverImplicitCatalog = 2` was fine before
 - modify `testRecordMainSectionValues` to test inline related table values when the custom mode shows a single link
   - also test custom mode with multiple values
   - required a change to `RecordsetColValue` type in `recordset-utils.ts`
 - modify `testRelatedTablePresentation` to check if the table is in tabular mode, if it isn't click the button so the following tests won't fail
 - delete downloaded files locally and in GA
   - doesn't necessarily matter in GA since the server is a fresh install and cleaned up each time
 - Test CSV item in export dropdown and use `test.skip` for `BDBag` tests in CI


RFSH (changes requested):
I was able to run the test cases and everything is working as expected. I really like that you moved the one-time-use function in record-helpers to be directly inside the spec. I always disliked how that function has a generic name but its implemented for just this one page.

My comments are mostly about reusing the functions that we've already added.

jrchudy:
@RFSH I updating this PR with the changes you asked for and what I mentioned in slack about testing BDBag in GA. I also updated the top comment to include details about each change that was out of the ordinary.

The major change since you last reviewed that I want to draw attention to is about testing the main record section values:
 - modify `testRecordMainSectionValues` to test inline related table values when the custom mode shows a single link
   - also test custom mode with multiple values
   - required a change to `RecordsetColValue` type in `recordset-utils.ts`
   
  I did notice that the "export" tests I wrote are _almost_ the same as the ones in the other spec but decided not to add a helper function for this since `clickAndVerifyDownload` is the bulk of it and the setup to get the elements is the "difference" in both places (even though the dropdown options are the same)

test/e2e/specs/all-features-confirmation/record/presentation.spec.ts
```diff
@@ -662,13 +618,15 @@ test.describe('View existing record', () => {
       if (!process.env.CI) {
         await test.step('delete files', async () => {
           // delete files that may have been downloaded during tests
-          await deleteDownloadedFiles(testParams.file_names);
+          await deleteDownloadedFiles(testParams.file_names.map((name: string) => {
+            return `${process.env.PWD}/test/e2e/.download/${name}`
```
RFSH, line 622: Please use the `DOWNLOAD_FOLDER` constant (`import { DOWNLOAD_FOLDER } from '@isrd-isi-edu/chaise/test/e2e/utils/constants'`)

jrchudy: Sorry, I didn't even think about that as a constant when you clearly wrote it as a "variable" in the last review

RFSH (approved):

PR #2522, opened 2024-08-20
Allow users to reorder facets

RFSH:
This PR is the updated version of #2459. 

New changes that are included in this PR:

- Changed how we're storing the order. Originally it was an array of arrays, where each element would be in the `["original_index", "new_index"]` format. We don't need to keep the original order at  all. `reference.facetColumns` will always give us the original order. So now the stored order is just an array of numbers similar to how `forms` state variable works in recordedit.

- Modified the initialization step to use the given order. Without this change, the facets wouldn't work as expected, and the flow control would ignore the user's preferred order. I went through different implementations as you can see in the git history. In the end, I decided to add a `getInitialFacetOrder` function, which consults both annotation and the stored order. I also added a `getInitialFacetOpenStatus` to make finding open facets easier (it returns a key-value pair which indicates the open status of a given `facetIndex`).

- Added a new `reorder-facet` spec to test the changes:
  - As part of this, I wanted to ensure users can interact with facets after reordering them. Instead of copying/pasting the `individual-facet` spec code, I refactored it into a function so it can be used in other places.


src/components/faceting/faceting.tsx
```diff
@@ -634,27 +701,58 @@ const Faceting = ({
...
-              <Accordion.Body>
-                {renderFacet(fc, index)}
-              </Accordion.Body>
-            </Accordion.Item>
-          ))}
-        </Accordion>
+        <DragDropContext onDragEnd={handleOnDragEnd}>
+          <ChaiseDroppable droppableId={'facet-droppable'}>
+            {
+              (provided: DroppableProvided) => (
```
jrchudy, line 707: Is `DroppableProvided` documented somewhere in our code/docs? Probably isn't needed though since we can inspect it in `react-beautiful-dnd`

jrchudy: I see there is a "similar" code block defined inside of this one for another `DraggableProvided`. Are both needed? If so, can you add a comment somewhere that explains why

src/utils/faceting-utils.ts
```diff
@@ -45,3 +47,115 @@ export function getInitialFacetPanelOpen(config: RecordsetConfig, initialReferen
   }
   return res;
 }
+
+export const getFacetOrderStorageKey = (reference: any): string => {
+  return `facet-order-${reference.table.schema.catalog.id}_${reference.table.schema.name}_${reference.table.name}`;
```
jrchudy, line 52: Do we want to have the same configuration for this table when it's shown in different contexts? I see that this is the specified "key" to use in the issue, but wasn't sure if that was discussed.

test/e2e/specs/delete-prohibited/recordset/individual-facet.spec.ts
```diff
@@ -1,18 +1,22 @@
...
 import {
-  clearRangeInput, fillRangeInput, openFacet, openFacetAndTestFilterOptions,
-  openRecordsetAndResetFacetState, testClearAllFilters, testDefaultRangePickerInitialValues,
-  testRangeInputSubmitThenClear, testSelectFacetOption,
-  testSelectFacetOptionThenClear, testTimestampRangePickerInitialValues
+  openRecordsetAndResetFacetState, testIndividualFacet, TestIndividualFacetParams,
 } from '@isrd-isi-edu/chaise/test/e2e/utils/recordset-utils';
 
-const testParams: any = {
+const testParams: {
```
jrchudy, line 9: I've avoided typing the `testParams` in specs that I've been writing for different reasons. We should make sure each spec is typed before we finalize the test migration

test/e2e/utils/recordset-utils.ts
```diff
@@ -390,3 +401,477 @@ export async function testFacetOptionsAndModalRows(page: Page, facetIdx: number,
...
+
+  justMax: {
+    max: string | { date: string, time: string },
+    filter: string,
+    numRows: number
+  },
+
+}
+
+export type TestIndividualFacetParams = Either<
```
jrchudy, line 510: Heh, another use of Either except now with 3 types to consider. So this looks like:
```
Either<A, Either<B, C>>
```

test/e2e/utils/recordset-utils.ts
```diff
@@ -390,3 +401,477 @@ export async function testFacetOptionsAndModalRows(page: Page, facetIdx: number,
...
+  TestIndividualFacetParamsChoice,
+  Either<TestIndividualFacetParamsCheckPresence, TestIndividualFacetParamsRange>
+>;
+
+/**
+ * This function can be used for testing an individual facet. This opens the facet, test its features, and then will close it.
+ */
+export async function testIndividualFacet(page: Page, pageSize: number, totalNumFacets: number, facetParams: TestIndividualFacetParams) {
```
jrchudy, line 518: I see what you mean by "moving" this function. I didn't think we would reuse it outside of the `individual-facet` spec but this is a good use for it. We want to "test" it's all working _without_ rewriting how to test a facet is working. 

This may be "more verbose" than it needs to for your test case after the facet is moved but that's not an issue anymore. We don't need to "ignore" or "skip" testing something anymore since playwright is so fast

jrchudy (approved):
Changes look good, functionality works as expected.

I added a few comments but the one I'm most unsure about is `provided: DroppableProvided` and `provided: DraggableProvided`

Also I didn't see the issue linked directly to this PR. You can find it by navigating to the "closed" PR that you mentioned and finding the issue from there.

test/e2e/utils/recordset-utils.ts
RFSH: Yeah exactly. This actually caused a bit of issue for me as this branch was old and I had a bunch of uncommitted changes. I had to ignore my local uncommitted changes and merge with master first.

src/utils/faceting-utils.ts
RFSH: Yeah, we want this to be used in different contexts. I'll comment this

src/components/faceting/faceting.tsx
RFSH: I changed the arguments to properly capture what they are (`draggableArgs` and `droppableArgs`). I also moved the fast list render to a different function so it's easier to follow the code.

PR #2523, opened 2024-08-20
Change link/unlink behaviour to directly apply changes without showing modal or message

### LINKED ISSUE #2273: Improvements related to link and unlink behavior
While discussing the changes for #2259, we realized that the link and unlink behavior could be more consistent. The following are changes that we should make:

- [x] Use "Close" for the unlink popup dismiss button (#2257).

- [x] Close the modal (and don't show the summary popup) after a successful unlink operation. When we implemented the unlink UI flow, given that the unlink operation could be broken into multiple requests to the server (due to the URL length limit), we decided to show a popup summary. But this summary is not needed when all the requests have been successful.

- [x] After a successful "link" do NOT show the alert at the top of record app. This should be removed entirely 

The following are other improvements that came up, but we decided to leave for the future:

- We might want to disable the submit button in the unlink case if the request exceeds the URL length limit. This way, we don't need to worry about partially successful cases.

- Currently, we're not showing any information about the total number of available records when a related entity is in custom mode. This can be confusing as we're only showing the first page of data and users might think we're showing all the available data.

KenyShah24:
Link/Unlink behavior updated to not show the confirmation/summary modals or message when the link/unlink is successful.

Resolves #2273


src/components/record/related-table-actions.tsx
```diff
@@ -553,39 +520,17 @@ const RelatedTableActions = ({
           dispatchError({ error: err, isDismissible: true });
         };
 
-        if (!CONFIRM_DELETE) {
```
jrchudy, line 556: For unlink, we are not trying to remove the delete confirmation modal. We only want to remove the modal with title "Batch Unlink Summary" that shows after clicking "Unlink". The confirm delete modal is a conditional popup that is configured in chaise-config.

The other change is the alert after success which you did remove properly

src/components/record/related-table-actions.tsx
```diff
@@ -519,30 +510,6 @@ const RelatedTableActions = ({
...
       validateSessionBeforeMutation(() => {
-        const deleteResponse = (response: any) => {
-          setShowPureBinarySpinner(false);
-
-          // Show modal popup summarizing total # of deletions succeeded and failed
-          response.clickOkToDismiss = true;
-
-          // TODO: - improve partial success and use TRS to check delete rights before giving a checkbox
-          //       - some errors could have been because of row level security
-          dispatchError({
```
jrchudy, line 530: If you pause the code at this line, you will see the `response` parameter in `deleteResponse` will be the "Batch Unlink Summary" which we want to avoid showing.

src/components/record/related-table-actions.tsx
```diff
@@ -519,29 +510,15 @@ const RelatedTableActions = ({
...
-          });
+        const deleteResponse = () => {
+          leafReference
+            .deleteBatchAssociationTuples(relatedModel.recordsetProps.parentTuple, selectedRows)
+            .then(()=>{
+              setUnlinkPureBinaryModalProps(null);
+              updateRecordPage(true, LogReloadCauses.RELATED_BATCH_UNLINK);
+            })
+            .catch(deleteError)
+            .finally(()=>setShowPureBinarySpinner(false));
```
jrchudy, line 521: `setShowPureBinarySpinner(false)` should be called before `setUnlinkBinaryModalProps(null)` a few lines above this and `finally` can be removed.

src/components/record/related-table-actions.tsx
```diff
@@ -519,29 +510,15 @@ const RelatedTableActions = ({
...
-                // NOTE: This feels very against React but the complexity of our flow control provider seems to warrant doing this
-                fireCustomEvent(CUSTOM_EVENTS.FORCE_UPDATE_RECORDSET, containerRef.current, {
-                  cause: LogReloadCauses.ENTITY_BATCH_UNLINK,
-                  pageStates: { updateResult: true, updateCount: true, updateFacets: true },
-                  response: response,
-                });
-              }
-            },
-          });
+        const deleteResponse = () => {
```
jrchudy, line 513: This is no longer the response to the `deleteBatchAssociationTuples` request. This should be renamed or undo the change to "refactor" this function.

src/components/record/related-table-actions.tsx
```diff
@@ -519,29 +510,15 @@ const RelatedTableActions = ({
...
+          leafReference
+            .deleteBatchAssociationTuples(relatedModel.recordsetProps.parentTuple, selectedRows)
+            .then(()=>{
+              setUnlinkPureBinaryModalProps(null);
+              updateRecordPage(true, LogReloadCauses.RELATED_BATCH_UNLINK);
+            })
+            .catch(deleteError)
```
jrchudy, line 520: Since this was refactored, `deleteError` is only being called in this one place. Either undo the change to `deleteResponse` so that both `deleteResponse` and `deleteError` are used twice or put the contents of `deleteError` in the catch here.

jrchudy (changes requested):
The functionality is good but the refactored code looks like you did some refactoring but didn't finish the changes.

As I mentioned in my inline comments, we should do 1 of 2 things. Either:
 1. keep `deleteResponse` and `deleteError` the way they were and undo the change to call `deleteBatchAssociationTuples` in the function
    - `deleteResponse` would be changed to:
        ```
        setShowPureBinarySpinner(false);
        setUnlinkPureBinaryModalProps(null);
        updateRecordPage(true, LogReloadCauses.RELATED_BATCH_UNLINK);
        ```
 2. rename `deleteResponse` to `deleteTuples` since it's now calling the delete function and handling both the "delete response" and "delete error"
    - remove `deleteError` as a named function and put its contents directly in the `catch` block here since the function is only used in 1 place

test/e2e/specs/all-features/record/related-table.spec.ts
```diff
@@ -403,6 +403,7 @@ test.describe('Related tables', () => {
 
         // click on submit
         await ModalLocators.getSubmitButton(rsModal).click();
+        await page.pause();
```
jrchudy, line 406: This should be removed

test/e2e/specs/all-features/record/related-table.spec.ts
```diff
@@ -544,7 +528,7 @@ test.describe('Related tables', () => {
         rowViewPaths: [
           [{ column: 'id', value: '1' }], [{ column: 'id', value: '3' }]
         ],
-        count: 2, // by load time it's one but when we add another related for the other table this should be updated too.
+        count: 1, // by load time it's one but when we add another related for the other table this should be updated too.
```
jrchudy, line 530: I don't think this should be updated, I'll look at the tests more closely when I review this later and understand why this might have changed.

jrchudy (changes requested):
Functionality is good, changes to test cases are good too.

Only need to cleanup the changes based on my inline comments.

jrchudy (approved):

PR #2525, opened 2024-08-22
Safari class + navbar bug fix

jrchudy:
Add chaise-safari class to the html body element. Add styles to fix a navbar menu bug with scrollbars showing.


jrchudy:
This PR fixes the bug in issue #2090 for safari browsers

PR #2527, opened 2024-08-29
Recordset presentation spec playwright

jrchudy:
This PR is for rewriting the recordset presentation spec from protractor to playwright.

Changes include the following:
 - Added `getAggregatesSpinners` for use in cases when there are many aggregate spinners in recordset table
   - when there are more than 1, `waitForAggregates` returns with an error asking which spinner did we mean
   - use `getAggregateSpinners` with `toHaveCount(0)` to accomplish the same thing
   - should this be added to the documentation?
 - modify `testExportDropdown` functionI added in the previous test I migrated to include app name since the CSV option has different text in each app
 - added `testTotalCount` function for testing the `Displaying` text that includes the page limit dropdown
   - as described in the comment for the function, if the dropdown was opened before testing this, the content needs to be tested differently
 - added `getWindowName` and `getPageId` page utility functions
 - added `testRecordsetDisplayWSortAfterPaging` for testing data on the current page after a sort occurs
   - this is specific to the tests in recordset presentation spec since we are sorting then testing page 1, paging to page 2 and testing it, then back to page 1
   - couldn't reuse `testColumnSort` since that was doing the sort and only testing 1 column of values after sorting for the displayed page
     - different type of test
 - added `testMainSearch` for testing  a search string and the # of rows displayed specific to this spec too
 
Other Notes:
 - we had a test marked `pend` in `recordset/presentation.spec.js` (old protractor spec) that was testing no catalog id OR schema:table. Now that we are injecting `defaultCatalog` as part of the copy chaise config step, I think we can test this now
   - since we know the default catalog to use (based on the property that was recently added), changing the default schema and table in the chaise config allows for this to be tested now both locally and in CI
 - added comment to facet-presentation spec describing why a test was skipped
   - comment was in old github PR


test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
```diff
@@ -0,0 +1,792 @@
...
...i and the possibility to bring your pets with you. NH Hotels has six resorts in the city of Munich. Very close to Munich Main Train Station -- the train being one of the most interesting choices of transport for travelling around Germany -- is the four-star NH München Deutscher Kaiser Hotel. In addition to the excellent quality of accommodation that it offers, the hotel is located close to Marienplatz, the monumental central square in the city, the Frauenkirche church, Stachus (Karlsplatz) and the Viktualienmarkt. Other places of interest to explore in Munich are the English garden, the spectacular Nymphenburg Palace and the German Museum, a museum of science and technology very much in keeping with the industrial spirit of the city. Do not forget to visit Munich at the end of September and beginning of October, the time for its most famous international festival: Oktoberfest! Beer, sausages, baked knuckles and other gastronomic specialities await you in a festive atmosphere on the grasslands of Theresienwiese. Not to be missed! And with NH Hotels you can choose the hotels in Munich which best suit your travel plans, with free WiFi and the possibility to bring your pets with you.',
```
jrchudy, line 64: This summary text is the same content twice (e.g. "abc123" + "abc123"). Before this was tested using `contains` but I wanted to reuse the `testRecordsetTableRowValues`. Otherwise, the "key/value" pairs in this object were reordered to be in the same order as the columns are displayed in the recordset table.

Note: This could be improved more since the tests never use the "keys". The only value that is accessed by key is `id`.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
```diff
@@ -0,0 +1,804 @@
...
+
+    const PAGE_URL = `/recordset/#${getCatalogID(testInfo.project.name)}/${params.schema_name}:${params.table_name}`;
+
+    await test.step('should load recordset page', async () => {
+      await page.goto(`${baseURL}${PAGE_URL}@sort(${params.sortby})`);
+      await RecordsetLocators.waitForRecordsetPageReady(page);
+
+      // "wait for" table column aggregate spinners to all be hidden by waiting until the count retruned in 0
+      // NOTE: using `waitForAggregates` fails because there are "multiple spinners on the page"
+      await expect.soft(RecordsetLocators.getAggregateSpinners(page)).toHaveCount(0);
```
RFSH, line 302: Oh good catch, I didn't realize `waitForAggregates` has this issue.

I don't think we should have two different ways of doing this, we should have just one `waitForAggregates` (or any other names that you see fit) that is used in all places and it should handle count. If we want to go by the way we've done this, it should most probably go to `recordset-locators`. And then maybe change the function to be `waitForRecordsetAggregates`

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
```diff
@@ -0,0 +1,804 @@
...
+      await page.goto(`${baseURL}${PAGE_URL}/${params.key}@sort(${params.sortby})`);
+      await RecordsetLocators.waitForRecordsetPageReady(page);
+
+      // NOTE: same comment from above about multiple spinners on the page
+      await expect.soft(RecordsetLocators.getAggregateSpinners(page)).toHaveCount(0);
+    });
+
+    await test.step('presentation of the recordset page', async () => {
+      if (!process.env.CI) {
+        await test.step('delete files', async () => {
```
RFSH, line 335: Please update the description to be the comment that you have

```
await test.step('delete files that may have been downloaded before', async () => {
```

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
```diff
@@ -0,0 +1,804 @@
...
+          }));
+        });
+      }
+
+      await test.step(`should have ${params.title} as title`, async () => {
+        await expect.soft(RecordsetLocators.getPageTitleElement(page)).toHaveText(params.title);
+      });
+
+      await test.step('should have the correct tooltip.', async () => {
+        await testTooltip(RecordsetLocators.getPageTitleElement(page).locator('.chaise-icon-for-tooltip'), params.comment, APP_NAMES.RECORDSET);
```
RFSH, line 348: We should avoid calling `.locator()` here. You should create a function for this instead. Similar to `getPageTitleTooltip` in `chaise.page.js`.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
```diff
@@ -0,0 +1,804 @@
...
+      });
+
+      await test.step('check the link of the view details after searching', async () => {
+        const keyValues = [{ column: 'id', value: params.data[3].id }];
+        const dataRow = getEntityRow(testInfo, params.schema_name, params.table_name, keyValues);
+
+        const filter = params.shortest_key_filter + dataRow.RID;
+        const viewUrl = `/record/#${catalogID}/${params.schema_name}:${params.table_name}/${filter}`;
+
+        const firstRowActionButton = RecordsetLocators.getRows(page).nth(0).locator('.view-action-button');
```
RFSH, line 474: Do `RecordsetLocators.getRowViewButton(page, 0)` instead.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
```diff
@@ -0,0 +1,804 @@
...
+        // search for a row that is not the first one after sorting
+        await searchBox.fill('9876.3543');
+        await searchSubmitButton.click();
+        await RecordsetLocators.waitForRecordsetPageReady(page);
+
+        await expect.soft(RecordsetLocators.getRows(page)).toHaveCount(1);
+        // clear search in next it case
+      });
+
+      await test.step('check the link of the view details after searching', async () => {
```
RFSH, line 467: I feel like it might be better if this step and the previous one are merged into one. Although this is completely subjective.

RFSH (commented):
I like the new utility functions that you've added. I understand that they are currently only used in recordset, but the way you have them is general-purpose. 

So I think having them in utils folder is ok. We can potentially use these functions to test sorting and search for the popups now that we're not limited by the test run time.

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
jrchudy: I did that everywhere else in this spec for action buttons but this one place. Good catch

test/e2e/specs/all-features-confirmation/recordset/presentation.spec.ts
```diff
@@ -0,0 +1,804 @@
...
+    detailedColumns: ['RID', 'id', 'text', 'int', 'RCB', 'RMB', 'RCT', 'RMT'],
+    compactColumnsPersonTable: ['id', 'text', 'RCB', 'RMT'], // no int column because it's the foreign key link (would be redundent)
+    entryColumns: ['id', 'text', 'int', 'RCB', 'RMB', 'RMT']
+  }
+};
+
+test.describe('View recordset', () => {
+  test.describe.configure({ mode: 'parallel' });
+
+  // NOTE: should this NOT run in CI still?
```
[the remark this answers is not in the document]
jrchudy: Passed twice in CI. removing this comment

PR #2539, opened 2024-09-06
Migrate Recordedit add, edit, and delete specs

RFSH:
This PR migrates the `add.spec.js`, `edit.spec.js`, `delete.spec.js`, `multi-edit.spec.js`, and `multi-col-types.spec.js`.  The following is the list of new specs and the changes that I made:

- `add.spec.ts`: I removed two specs from this file. One is the "prefill" test which we already are doing as part of related spec. And the other one is the "Markdown edit help button", which I moved to be done as part of validating the markdown input itself.
- `edit.spec.ts`: I removed the "clear all foreign key" test from this spec and added it to the `multi-form-input-edit.spec.ts` spec. The `edit.spec.js` was only focused on single edit, but I don't think we needed two different specs. So I ended up removing `multi-edit.spec.js` completely and so `edit.spec.ts` is now testing both single and multi edit.
- `delete.spec.ts`: I didn't make any significant changes to this. It's basically the rewrite of `delete.spec.js`.
- `null-values.spec.ts`: The new version of `multi-col-types.spec.js`. The old spec was testing disabled inputs and null values. But we already have a dedicated `immutable-inputs.spec.js` for disabled inputs. So I renamed it to be more accurate.

While doing this I also made these changes to the chaise code to solve the bugs/issues that I found:

- We're supposed to use `00:00:00` for timestamp's time input when it's empty. But chaise was actually showing an error when users left the time empty. So I had to make some changes to the `date-time-field.tsx` component.
- There was an eslint error/warning in `navbar.tsx` related to how doing `{ ...cc.navbarMenu } || {};` is not going to do what you would expect as the first part of this statement is always truthy.

The following is what's left:
- [x] Figuring out why array of timestamps are not showing the warning that reminds user to click on "add".
- [x] Figure out why the title tooltip test is not working.
- [x] Figure out why the asset tooltip test is not working.
- [x] Adding validation test for arrays.


RFSH:
@jrchudy This PR is ready to be reviewed. In my last commit, I fixed the timestamp issue that I mentioned. I also realized that the `remove-edit-form` spec relies on the schema that I deleted. So I added a test for this one to `edit.spec.ts` as well.

As I mentioned in the description, the asset and title tooltip tests are not working, and I will work on those now. But I won't make a lot of changes for that, so you can review the PR in the meantime.

jrchudy (approved):
This looks good. All the tests in `make testrecordadd` and `make testrecordedit` passed for me.

RFSH:
@jrchudy I just pushed a small fix for tooltip test cases:

- When I originally implemented the `testTooltip` function, I modified it so it would hover over a dummy element after it's done with testing the tooltip. This way the function is doesn't interfere with other tests. In recordedit I chose "required info" because it's always there. But it turns out the title tooltip test cases were failing because the tooltip would show up on top of "required info". So Playwright couldn't hover over it and it would just fail. So I added an optional argument to this function so we can pass a different "hover element".

- For the asset tooltip, when we select a file, the input stays on focus. So doing a "locator.hover()" wasn't doing anything. That's why I had to add a dummy hover over "required info" before testing the asset tooltips.

PR #2540, opened 2024-09-07
migrated recordset add and edit specs

jrchudy:
Migrated the last 2 specs for recordset, the add and edit specs. Since these are the last 2 recordset tests to be migrated, the makefile has been cleaned up some more to remove the protractor recordset commands.

Note:
 - The changes for `add.spec` include testing the show more/less in a cell with a longtext value that overflows
 - testing the dimensions like I am looks wrong but I couldn't think of another way to do it while using `boundingBox()` which returns a promise


test/e2e/specs/default-config/recordset/add.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    // default config: maxRecordsetRowHeight = 160
+    // 160 for max height, 10 for padding
+    const cellHeight = 170;
+
+    const firstRow = RecordsetLocators.getRows(page).nth(0);
+    const testCell = RecordsetLocators.getRowCells(firstRow).nth(4);
+    await expect.soft(testCell).toHaveText(/... more/);
+
+    let dimensions = await testCell.boundingBox();
+    if (!dimensions) return;
```
RFSH, line 47: Why would `dimensions` be falsy? Should this block other test cases? I think it should just be a condition for the calculation that you're doing 

```
if (dimensions) {
  // the calculations might be one pixel off
   expect.soft(Math.abs(dimensions.height - cellHeight)).toBeLessThanOrEqual(1);
}
```
So it can proceed with testing the `.click()` regardless of this.

RFSH (approved):
Changes look good and tests pass locally. I just left a minor comment about how the dimension test is done. I think the way you have it as a promise is fine, but I don't think it should block us from testing the other steps.

test/e2e/specs/default-config/recordset/add.spec.ts
jrchudy: This was how I had it initially but thought we wanted to break from this `test.step` if it was not defined.

Issue #2544, opened 2024-09-16
Display facet panels as part of the entity facet popups (facet within facet)


RFSH:
## Summary
When we initially implemented facet popups, we decided to hide the facet side panel for them. To get around this limitation, data-modelers customize the facet list and define all the properties we care about on the initial recordset page (even if they are directly connected to other facets).

To avoid the need for this customization by default, we are going to allow the display of the facet panel on the facet popups.

## Details

We currently have a `showFaceting` chaise-config property to control whether we should display facets. As part of the changes for this issue, we will deprecate this property. Instead, we'll add a new `maxFacetDepth` property to the `facetPanelDisplay`:

```
"facetPanelDisplay": {
  "open": [],
  "close": [],
  "maxFacetDepth": 2
}
```
The following are acceptable values for this property:

- 0: disable the faceting feature.
- 1: The current behavior (facet panel is displayed on the initial recordset page or popups. But it's disabled for facet popups).
- 2: Facet panels are displayed on the facet popups as well.
- 2+: Not allowed for now, as we think allowing more facets would confuse the users. When we encounter this value, we will use the maximum we support (2).

We will also allow data-modelers to define this property as part of the `display` annotation (which can be defined on a table, schema, or catalog):
```
"tag:misd.isi.edu,2015:display": {
  "max_facet_depth": {
    "compact": 2
  }
}
```

We will use the display setting defined for the "main" table the facet panel belongs to. We're also not going to recursively look for a different setting once users opened a facet popup. So,
- In a recordset app, the main displayed table will dictate whether we should show the facet panels on the facet popups.
- We will derive this setting for the record page link and unlink popups from the table users see on the popup (not the main record page).
- For foreign key popups, the table that the foreign key refers to will be used (as we're showing that table in the popup).

P.S. We discussed the possibility of allowing each facet to change the allowed depth, but since it was confusing, we decided not to allow it for now. If anything, we might add a `"show_facet_in_facet"` boolean to facets instead of depth, so it's directly obvious what the setting does.

PR #2545, opened 2024-09-20
Facet within facet

RFSH:
This PR implements the changes described in #2544. Summary of changes:

- Deprecated chaise-config's `showFaceting` in favor of the new `maxFacetDepth` property of `facetPanelDisplay`.
- While discussing the UI, we realized that we should add a "context" above the title of facet popups to distinguish between different layers of facet popup. To implement this I added the `uiContextTitles` prop to recordset component. This property captures a list of titles:
  - For the facet popups in recordset page, the first element is the title of the recordset page. The next item would be the name of the facet that is opened.
  - For the foreign key popups in recordedit page, the first element is the name of the foreign key column.
  - For the popups in record page, the first element is the name of the visible column.

The bulk of changes in this P.R. are related to testing. Supporting multi-level facets is something that we had in ermrestjs for a long time, and we just never allowed it in Chaise. While the unit tests in ermrestjs ensure parser can understand such urls, we're not testing all the cases in ermrestjs. Since testing this in Chaise is much more straightforward, I decided to do the main testing here. The new `facet-within-facet` spec goes through all different types of entity facets that we allow, and ensures the second level facet shows up and the parser doesn't throw any errors.

P.S. I originally added the spec to `all-features`. But I didn't realize `all-features` is disabling parallelism since multiple test cases change the catalog annotation. So, I moved it to `default-config` in my last commit.


RFSH:
@jrchudy This PR is ready to be reviewed. I updated the PR description to reflect the latest changes.

test/e2e/specs/default-config/recordset/facet-within-facet.spec.ts
```diff
@@ -0,0 +1,740 @@
...
+      /**
+       * useful if you want to select something without affecting other facets.
+       */
+      clearAll?: boolean
+    }[]
+  }
+
+}
+
+const popuFacetNames = [
```
jrchudy, line 110: Not a big deal, misspelled variable name

test/e2e/specs/default-config/recordset/facet-within-facet.spec.ts
```diff
@@ -0,0 +1,740 @@
...
+          ]
+        }
+      }
+    ]
+  }
+}
+
+/******************** helpers ************************/
+
+const getChaiseURL = (appName: string, tableName: string, baseURL: string | undefined, testInfo: TestInfo) => {
```
jrchudy, line 570: This should probably be a common function that all of our tests can use. You did something similar in errors spec

jrchudy (approved):
The changes look good. I added comments to better my understanding of the facet within facet relations.

Other than the inline comments, I had this question/observation:
 - `showFaceting` is now a property of reference instead of something we hard code for each recordset "props"?
   - since it's based on `maxFacetDepth` now, I think that's why you had to do this
   - so it will check for maxFacetDepth in:
     1. display annotation on table, schema, then catalog
     2. catalog chaise-config
     3. server chaise-config
     4. use the default of 1

test/e2e/specs/default-config/recordset/facet-within-facet.spec.ts
```diff
@@ -0,0 +1,740 @@
...
+ *     <-- main_inbound2
+ *                      <-- main_inbound2_inbound1 : "Entity path with shared prefix"
+ *                      --> main_inbound2_outbound1: "Entity path with shared prefix and filter"
+ *     <-- main_inbound3
+ *                      <-- main_inbound3_inbound1: "Entity Path with fast filter"
+ *                      <-- main_inbound3_inbound1_fast_filter: the fast_fitler_source of prev facet
+ *
+ * To simplify the second layer facets, I added a facet_2nd_layer_hub which has fk to all the tables above,
+ * and other new tables have inbound to them to be displayed as facets. so regardless of which facet we opened
+ * for the first layer, the second layer facets all look the same in UI but they have different paths to the main table.
```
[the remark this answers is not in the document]
RFSH: These are the reasons for why I designed the tables this way:

1. I wanted to make sure that I can have two levels of facets, and on each level I wanted to have all the 6 different 
facets. 

2. All paths are using two hops because I wanted to force the `null` facet for all these paths to use the right outer join (we try to optimize the `null` facet and in some cases we might eliminate one hop so we can use inner join). All the bugs that I found were related to how the right outer join works, so I wanted to make sure that I test this for all the cases.

3. I wanted to reuse as many tables as possible so it's a bit easier to follow and test.


So the facets that are displayed for the first facet all are based on a 2-hop path to main. To better understand why I added `facet_2nd_layer_hub` let's focus on only two of the facets to show how this helped me reduce the size of schema. For example these two facets (using shorter name so it fits the page):

```
main <-- main_i1 <-- main_i1_i1

main <-- main_i2 <-- main_i2_i1
```

When you open the first facet, we're going to show the rows in `main_inbound1_inbound1`, and when you open the other one we will show the `main_inbound2_outbound1` table. If we want to design a facet for these tables similar to the first layer, we should create 2-hop paths from these tables to the facet that I want. For example I could do this:

```
main_i1_i1  <-- main_i1_i1_i1 <-- main_i1_i1_i1_i1

main_i1_i1  <-- main_i1_i1_2 <-- main_i1_i1_i2_i1

main_i2_i1 <-- main_i2_i1_i1 <-- main_i2_i1_i1_i1

main_i2_i1 <-- main_i2_i1_i2 <-- main_i2_i1_i2_o1
```

And this is just for 2 facets and showing 2 facets on the second layer. As you can imagine doing this for all 5 paths and having to create 5 other unique path for each one would be hard to manage. Since all I care about is having a 2-hop path, I added a `facet_2nd_layer_hub` in between, so the above could be simplified as just:

```
main_i1_i1 <-- facet_2nd_layer_hub
main_i2_i1 <-- facet_2nd_layer_hub

facet_2nd_layer_hub <-- facet_2nd_layer_hub_i1
facet_2nd_layer_hub <-- facet_2nd_layer_hub_i2
```

This way, the facets for the `main_i1_i1` are:

```
main_i1_i1 <-- facet_2nd_layer_hub <-- facet_2nd_layer_hub_i1
main_i1_i1 <-- facet_2nd_layer_hub <-- facet_2nd_layer_hub_i2
```

And the facets for `main_i2_i1`:

```
main_i2_i1 <-- facet_2nd_layer_hub <-- facet_2nd_layer_hub_i1
main_i2_i1 <-- facet_2nd_layer_hub <-- facet_2nd_layer_hub_i2
```

test/e2e/specs/default-config/recordset/facet-within-facet.spec.ts
RFSH: Yeah, I actually have done something similar in almost all the specs that I wrote at this point 😄 I'll refactor it.

src/components/modals/recordset-modal.tsx
```diff
@@ -391,6 +412,7 @@ const RecordsetModal = ({
         {showRecordset &&
           <Recordset
             {...recordsetProps}
+            uiContextTitles={uiContextTitles}
```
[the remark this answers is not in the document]
RFSH: Yes, that's correct. I know the name is not the best but I called the element `modal-header-context` and so I named this one similar to that.

test/e2e/specs/default-config/recordset/facet-within-facet.spec.ts
RFSH: I'm going to introduce a new function for this. But I'll do the refactoring of existing functions as part of another PR that I'm planning to do for test improvements.

PR #2547, opened 2024-09-26
Remove old/unused files and folders

RFSH:
This PR removes all the folder and files that are not needed anymore. The following are the other changes that I had to do as a result of this:

- We had an `images` folder which was used by both e2e tests and our chaise code (markdown help). So I moved the one that chaise uses to `src/assets` and the test images to `test/e2e` folder. I also had to make some changes to `playwright.setup.ts` to properly copy the this new folder into location.

- Cleaned up the makefile.


jrchudy (approved):

RFSH:
The older deployments that are still using AngularJS navbar app don't have automatic updates and making this change won't affect them. That's why we decided to merge this PR.

PR #2564, opened 2024-10-29
Added sticky scrollbar and intersection API to recordset tables

### LINKED ISSUE #2448: Make recordset horizontal scrollbar sticky
We would like to make the horizontal scrollbar that we're displaying above the table to be sticky. 

If we only want this on the recordset page, we can follow what we did for the recordedit page. But if we want this to work on all the recordset table instances, we probably need to develop a new solution.

P.S. On the AngularJS version of recordset page we tried making the whole table header sticky, but given the issues we faced, we didn't merge the changes. The approaches we tried and could try with React are documented in #2101.

KenyShah24:
There are two scenarios for handling the sticky top horizontal scrollbar issue:

For pages with a single table (recordset page), make the top scrollbar sticky.
For pages with multiple tables (record page), keep the top scrollbar sticky until the bottom scrollbar becomes visible. 

Resolves #2448


src/components/recordset/recordset-table.tsx
```diff
@@ -491,6 +526,7 @@ const RecordsetTable = ({
           </tbody>
         </table>
       </div>
+      <div ref={tableEndRef} style={{ height: '1px' }}></div>
```
jrchudy, line 529: Avoid using inline styles

src/components/recordset/recordset-table.tsx
```diff
@@ -472,10 +507,10 @@ const RecordsetTable = ({
     const tableSchemaNames = `s_${makeSafeIdAttr(reference.table.schema.name)} t_${makeSafeIdAttr(reference.table.name)}`;
     return classNameString + ' ' + tableSchemaNames;
   }
-
   return (
     <div className='recordset-table-container' ref={tableContainer}>
-      <div className='chaise-table-top-scroll-wrapper'>
+      <div ref={stickyScrollbarRef}
+        className={`chaise-table-top-scroll-wrapper ${isBottomVisible ? 'no-scroll-bar' : ''}`}>
```
jrchudy, line 513: Better to write this like:
```
`chaise-table-top-scroll-wrapper${isBottomVisible ? ' no-scroll-bar' : ''}`
```
so that the _whitespace_ is only added if the extra class is also added.

jrchudy:
Overall the changes look good and the feature behaves how I expect. I'll let Aref review and merge this PR when he gets a chance.

src/components/recordset/recordset-table.tsx
```diff
@@ -120,6 +123,38 @@ const RecordsetTable = ({
 
     rowDetails = tempRowDetails;
   }
+  const [isBottomVisible, setIsBottomVisible] = useState(false);
+  const stickyScrollbarRef = useRef(null);
+  const tableEndRef = useRef(null);
+
+  useEffect(() => {
+    //Only implement intersection observer for top scrollbar when intersectScroll is true otherwise top scrollbar will be sticky all the time
+    if(intersectScroll){
```
RFSH, line 132: Instead of having an `if` statement that wraps the whole function, it's cleaner if you check for the opposite and just return:

```tsx
if (!intersectScroll) return;

// the rest of the function goes here 

```

This way you're eliminating an extra indentation (which you didn't even add anyways :D )

src/components/recordset/recordset-table.tsx
```diff
@@ -29,12 +29,14 @@ import { addTopHorizontalScroll, fireCustomEvent } from '@isrd-isi-edu/chaise/sr
 type RecordsetTableProps = {
   config: RecordsetConfig,
   initialSortObject: any,
+  intersectScroll?: boolean,
```
RFSH, line 32: I'm assuming you used `intersectScroll` because you're using the `Intersection Observer`, but you should name it based on what it does not how it's implemented. Maybe something like `showOnlyOneHorizontalScrollbar` (which is inverse of what you have)? Or if you could come up with a better and shorter name?


Also please add a jsdoc style of comments to all the `type`s that you added the new prop:

```tsx
type Whatever = {
  /**
   * Explaing what's its purpose 
   * and if there's a default behavior also explain that here
   **/
  someProp?: Type
}
```
(I know that we should've added jsdoc to other props in these `type`s. But don't worry about existing props. Just add jsdoc for your own new prop).

RFSH (changes requested):
Functionality-wise, the scrollbar works very well and I couldn't find any issues. I just made some comments about code changes that you should apply.

Apart from the comments that I made, I noticed that you've moved the import statements around. I saw the same the thing in your previous PRs. Is that something that you're doing manually? or is your formatter doing it? Regardless, please stop doing this. It inflated the PR for no reason and makes tracing the history and looking at your PR necessarily harder.

So please revert all the import changes that you did. And if it's done by your formatter, make sure to run your formatter only on the lines that you modified.

RFSH (approved):

PR #2608, opened 2025-02-11
Sticky header using intersection observer for recordset page

KenyShah24:
This PR includes code changes to ensure the table header remains sticky while scrolling content on the recordset page using the Intersection Observer.

Issues Identified During Testing:
1. E2E Testing with Docker:
- The Docker image build process encounters compatibility issues with Node.js and npm versions required for the Chaise project.
- Additionally, openjdk-8-sdk is missing when running the script.
The included changes address these dependency errors
2. E2E Testing Locally:
- On Windows, the mkdir -p command is not functioning as expected.
- A solution for this issue is still being explored.


RFSH (changes requested):
The implemented solution is not responsive and therefore we cannot merge it as is.

The vertical position of the sticky header is calculated on the initial load and remains the same. So if the user squeezes the page or remove a banner/alert (which changes the height of the top section), the sticky header shows up in the wrong spot.

Also there's no reason to keep the height in the React state. It's only used for manipulating the position of sticky header and not used in the render logic, so we shouldn't add it to state.

Other changes that should be applied:

- Please remove `.last-run.json`.
- Revert the docker file changes that you did.
- Revert the changes in `locators/recordset.ts`. Instead of using filter, we just need to change the selector to use the `:not(.sticky-header)` stynax. You can just revert you changes. Once we think this PR is ready, I'll take care of test cases myself.

RFSH (changes requested):
We had a call about this and I'm going to summarize what we talked about here:

The `attachContainerHeightSensors` function already does the calculations that we need and so we shouldn't add another `resizeSensor` for doing similar calculations. 

Instead we can add a callback as an argument to `attachContainerHeightSensors` which would tell us about the height changes. And then we can use that callback for setting the `top` of the sticky header.

src/components/recordset/recordset-table.tsx
```diff
@@ -534,7 +612,14 @@ const RecordsetTable = ({
       </div>
       {/*  This div will be used as the target (end of table) for the intersection observer to hide the 
       top scrollbar when the bottom one is visible */}
-      <div className='dummy-table-end-div' ref={tableEndRef}/>
+      <div className='dummy-table-end-div' ref={tableEndRef} />
+      {config.displayMode.indexOf(RecordsetDisplayMode.RELATED) !== 0 && <div className='sticky-header' id='sticky-header' ref={stickyHeaderRef}>
```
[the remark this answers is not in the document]
RFSH: Remove the `id` attribute.

src/utils/ui-utils.ts
```diff
@@ -94,6 +98,12 @@ export function attachContainerHeightSensors(parentContainer?: any, parentContai
           resetHeight();
         }
       }
+      if (onContainerHeightChange) {
+        //To notify height change
+        onContainerHeightChange({
+          top: containerSticky.offsetHeight + parentContainerSticky.offsetHeight,
```
RFSH, line 104: Instead of recalculating this, is it possible to change the code above to first calculate the top. And then sets the container height using this?

What you have is not working properly for popups. You can test this on `isa:dataset` table (https://dev.derivacloud.org/chaise/recordset/#1/isa:dataset@sort(accession::desc::). For example the `Title` facet is a good example.

src/components/recordset/recordset-table.tsx
```diff
@@ -157,10 +164,75 @@ const RecordsetTable = ({
...
+      },
+      { root: null, threshold: 0 }
+    );
+    observer.observe(headRef.current);
+
+    // Sync widths of the columns
+    const syncWidths = () => {
+      if (stickyHeaderRef.current && tableRef.current) {
+
+        const originalThs = tableRef.current.querySelectorAll('tbody > tr > td');
```
RFSH, line 199: Why is this selecting `td`s? and not `th`?

src/components/recordset/recordset.tsx
```diff
@@ -367,7 +369,8 @@ const RecordsetInner = ({
     }
 
     // handle the scrollable container
-    const resizeSensors = attachContainerHeightSensors(parentContainer, parentStickyArea);
+    const resizeSensors = attachContainerHeightSensors(parentContainer, parentStickyArea, undefined, setTopPanelDimensions);
```
RFSH, line 372: The more I think about it, the more I don't like this solution. Since the callback is called on each height or width change, React is rerendering recordset component and all its children on each height/width change. So this solution would never work on record page, and it can slow the page down on long recordset pages.

As I mentioned before, we're just manipulating the styles attached to the elements, so there's no reason to involve React state in this. Instead of calling `setTopPanelDimensions` and passing the state variable to `RecordsetTable`, is there anyway that we can pass the `top` value from here? Maybe attach it as a `data-*` attribute to an element and then grab it from that page?

RFSH (changes requested):
With the latest changes, the sticky header is now properly adjusting as the available space above the table changes. The width of the headers are also adjusting properly.

But there are still two functionality issues that I mentioned in my comments and I'll summarize here:

1. The `top` calculation is not working properly for the popups. It's not honoring the vertical space outside of the modal, so the sticky header shows up on top of the page size dropdown.

2. This implementation is forcing React to rerender the recordset component and all its children after each height/width change. This can cause performance issues on busier pages.

By the way, please update this branch with the latest changes of the main branch.

PR #2621, opened 2025-03-24
Allow access to outbound foreignkey values in url_pattern attribute of assets

RFSH:
Chaise side of changes for https://github.com/informatics-isi-edu/ermrestjs/pull/1020


RFSH:
While working on test cases I realized that the `"comment": false` support was a bit buggy. So I fixed it in this PR and also added test cases for it.

